PR #86, opened 2026-07-10
ci: run the pipeline on a self-hosted runner

nghiadaulau:
Runs the CI jobs (lint, backend, frontend) on a self-hosted runner instead of `ubuntu-latest`.

## Before merging
A self-hosted runner must be registered for the repo (Settings > Actions > Runners), or every workflow run will queue forever waiting for a runner, including this PR's own checks.

## Security note (public repo)
synapse-ce is public and CI runs on `pull_request`, so an approved fork PR will execute on your self-hosted runner. To keep that safe:
- Use an ephemeral runner (fresh VM/container per job) and isolate it from your internal network and any long-lived credentials.
- Keep "Require approval for all outside collaborators" on in Settings > Actions (first-time contributor runs already require approval).
- Consider restricting the runner to a dedicated, disposable environment.

A safer alternative, if you want it later, is a hybrid: self-hosted for `push`/`workflow_dispatch` and GitHub-hosted for `pull_request` from forks. Happy to switch to that.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -6,13 +6,16 @@ on:
   pull_request:
     branches: [main]
 
+# Runs on a self-hosted runner. Note: this is a PUBLIC repo, so a fork pull request executes on your
+# infrastructure once its workflow run is approved. Keep the runner ephemeral and network-isolated, and
+# keep "require approval for all outside collaborators" on in Settings > Actions.
 permissions:
   contents: read
 
 jobs:
   lint:
     name: Lint (Go)
-    runs-on: ubuntu-latest
+    runs-on: self-hosted
     steps:
       - uses: actions/checkout@v7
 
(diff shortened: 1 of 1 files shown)


nghiadaulau:
Closing: the self-hosted runner is a Coder workspace (a nested container). Verified on the box (systemd PID1, Delegate=yes, pids/memory controllers present, sudo) that it still cannot enforce memory.max or a read-only host-root mount inside the sandbox, even with full cgroup delegation, and netns egress does not hold either. That is a nested-virtualization limit, not a config gap. The four privileged sandbox integration tests only pass on a real (non-nested) Linux host. Staying on github-hosted ubuntu-latest, where those tests skip cleanly (no bubblewrap). If we want them validated in CI later, a dedicated Linux VM runner plus hybrid triggers (self-hosted only for push) is the safe path.

PR #117, opened 2026-07-11
feat(sca): add Conan 1.x dependency graph edges

VietSory:
## Summary

Add deterministic dependency graph extraction for Conan 1.x `graph_lock` files.

The existing Conan parser already emits resolved components from graph nodes. This change also resolves each node’s `requires` and `build_requires` node IDs into normalized PURL relationships, allowing downstream dependency-path analysis to understand how Conan packages depend on one another.

## Changes

* Extend Conan 1.x graph nodes with `requires` and `build_requires`.
* Build a two-pass node-ID-to-PURL index before resolving graph edges.
* Emit normalized `sbom.Dependency` relationships.
* Merge Conan graph nodes that normalize to the same PURL.
* Union outgoing targets for duplicate source identities.
* Deduplicate repeated dependency targets.
* Remove direct and PURL-collapsed self-edges.
* Ignore dangling node IDs and malformed package references without aborting the entire scan.
* Keep `python_requires` out of node-ID resolution because Conan 1.x serializes those values as package references rather than graph node IDs.
* Sort components, dependency records, and `DependsOn` targets deterministically.
* Preserve existing Conan 2.x and `conanfile.txt` behavior.
* Add parser, registry, endpoint-integrity, path-to-root, context-cancellation, edge-case, and determinism tests.
* Bump the OwnSBOM generator version to `ownsbom/0.4.0`.
* Update the changelog.

## Behavior

Given a Conan 1.x graph such as:

```text
app/1.0
├── lib-a/2.0
│   └── openssl/3.0.0
├── lib-b/3.0
└── cmake/3.29.0
```

OwnSBOM now emits relationships equivalent to:

```text
pkg:conan/app@1.0
  → pkg:conan/cmake@3.29.0
  → pkg:conan/lib-a@2.0
  → pkg:conan/lib-b@3.0

pkg:conan/lib-a@2.0
  → pkg:conan/openssl@3.0.0
```

## Edge handling

The parser intentionally:

* skips source nodes without a valid package reference;
* skips missing or malformed target nodes;
* removes duplicate targets;
* removes self-edges;
* merges nodes that collapse to the same normalized PURL;
* emits only dependency endpoints that also exist in the generated component list;
* omits empty dependency records for leaf packages.

## Non-goals

* No dependency graph inference for Conan 2.x flat lockfile lists.
* No graph inference for `conanfile.txt`.
* No parsing of `conanfile.py`.
* No synthetic project-root component.
* No Conan CLI or network execution.
* No component-scope changes.
* No host/build-context PURL qualifiers.
* No changes to the SBOM domain model or parser interfaces.

## Known limitation

Conan host and build nodes with the same package reference normalize to the same PURL identity. Their outgoing targets are therefore unioned, and any resulting self-edge is removed.

## Testing

The following targeted checks were run locally:

```bash
go test ./internal/infrastructure/tools/ownsbom \
  -run 'TestConan' \
  -count=1 \
  -v

CGO_ENABLED=0 go build ./cmd/...

cd web
pnpm run typecheck
```

The Conan test suite covers:

* Conan 1.x and Conan 2.x regression behavior;
* basic dependency graph extraction;
* duplicate and missing targets;
* malformed references;
* duplicate source identity aggregation;
* collapsed self-edges;
* reference-less root nodes;
* exclusion of `python_requires` from node-ID resolution;
* endpoint integrity;
* downstream `PathToRoot` behavior;
* registry propagation;
* context cancellation using `errors.Is`;
* deterministic components and dependency output across repeated parses;
* existing `conanfile.txt` parsing and lockfile precedence.

## Checklist

* [x] Conan 1.x `requires` edges supported
* [x] Conan 1.x `build_requires` edges supported
* [x] `python_requires` not misinterpreted as node IDs
* [x] Duplicate source identities aggregated
* [x] Duplicate targets removed
* [x] Self-edges removed
* [x] Dangling and malformed targets handled fail-soft
* [x] Dependency endpoints validated against components
* [x] Output ordering deterministic
* [x] Conan 2.x behavior preserved
* [x] `conanfile.txt` behavior preserved
* [x] Registry integration tested
* [x] Generator version bumped
* [x] Changelog updated
* [x] No new dependencies introduced


nghiadaulau:
Hi @VietSory , please resolve conflict. Thank you

nghiadaulau:
Thanks — solid, well-tested work. Verified independently before merge:

- Merges cleanly on `main`; `go build`, `go vet`, and the `ownsbom` tests pass locally.
- Ran an independent fixture (separate from the PR tests): app → {cmake, lib-a} sorted, lib-a unions {cmake, openssl} across two nodes that collapse to the same PURL, self-edges dropped, dangling node IDs skipped, PURL-collapse dedup correct, output deterministic. All correct.
- Architecture review is clean: dependency rule holds, EcosystemParser contract intact, the two-pass node-ID→PURL index + edge aggregation is correct.

Approving and merging. Two small polish items I'll clean up in a quick follow-up on our side (non-blocking): the type doc comment ended up with the old block's first lines left in front of the new one (a duplicated/truncated comment), and the `if len(targets) > 0` guard in the materialize loop is always true. Nice addition.

nghiadaulau (approved):

Issue #123, opened 2026-07-11
SAST: `generic-sql-dynamic-execute` over-matches non-SQL `.execute()` (Executor.execute → false CWE-89)


nghiadaulau:
## Summary

The `generic-sql-dynamic-execute` rule (`internal/infrastructure/tools/sast/patterns.go`) massively over-matches: it flags ordinary, non-SQL `.execute(...)` calls — most notably Java's `Executor`/`ExecutorService.execute(Runnable)` — as CWE-89 SQL injection (High).

On a real Java project this rule produced multiple High false positives such as:

```java
taskExecutor.execute(decorator.decorate(command));   // java.util.concurrent Executor — NOT SQL
worker.execute(new SomeContext(lookup(id)));          // domain worker dispatch — NOT SQL
```

## Root cause

```
(?i)(cursor\.execute|execute(Query|Update)?|mysqli_query|pg_query|sequelize\.query|ActiveRecord::Base\.connection\.execute)\s*\([^;\n]*(\+|%|\.|\$\{|#\{|params\[|request\.|req\.|\$_(GET|POST|REQUEST))
```

Two defects:

1. **`execute(Query|Update)?` matches a bare `execute(`.** Any `.execute(` sink matches — including `Executor.execute(Runnable)`, `CompletableFuture`, custom worker APIs. There is no SQL context.
2. **A literal `.` is in the "tainted data" alternation** (`(\+|%|\.|...)`). A single dot is not evidence of dynamic SQL — so `foo.execute(bar.baz())` matches purely on the dot inside the arguments.

Together, *any* `x.execute(y.z(...))` is reported as SQL injection.

## Fix

- Remove bare `execute(` — require a genuine SQL sink (`cursor.execute`, `executeQuery/executeUpdate` on a Statement, `mysqli_query`, `pg_query`, `sequelize.query`, `Connection.createStatement().execute`, …), not `execute(` alone.
- Drop the standalone `\.` from the dynamic-string evidence group; keep `+`, `%`, `${`, `#{`, string-format, and request-derived markers.
- Scope with `exts` where the sink is language-specific (see companion issue on `exts`).

## Acceptance

- `Executor.execute(Runnable)` / `future.execute(...)` / worker `.execute(ctx)` produce **no** finding.
- A true positive (`stmt.executeQuery("SELECT ... " + userInput)`) still fires.
- Table tests cover both the negative (executor) and positive (concatenated SQL) cases.

[...]

PR #128, opened 2026-07-11
fix(sast): scope language-specific rules + stop generic-sql-execute matching non-SQL .execute() (#122, #123)

### LINKED ISSUE #122: SAST: language-specific rules missing `exts` scope fire on foreign languages (e.g. Python rule on .java)
## Summary

Several language-specific SAST rules in `internal/infrastructure/tools/sast/patterns.go` do not set the `exts` field, so they run against **every** source file regardless of language. A Python- or JS-idiom rule then fires on unrelated languages (e.g. Java, Go), producing false positives with a misleading title.

Observed on a real Java-only project: `sqlalchemy-raw-sql-dynamic` (titled *"Python SQLAlchemy/raw SQL uses dynamic string construction"*) fired on `.java` files that contain no SQL at all — the title alone ("Python …" on a `.java` file) makes the finding obviously wrong to a reviewer.

## Root cause

The infrastructure for language scoping already exists and is unused by these rules:

```go
// rule.exts, when non-nil, restricts the rule to those file extensions.
func (r *rule) appliesTo(ext string) bool { return r.exts == nil || r.exts[ext] }
```

`nil` exts == "runs on all files". Rules that encode a single-language idiom left `exts` nil:

- `sqlalchemy-raw-sql-dynamic` — Python only (`.py`)
- `prisma-raw-sql-unsafe` — JS/TS only
- `child-process-exec-template` — JS/TS only
- `unsafe-deserialization-node-serialize` — JS/TS only

(Audit the full list; any rule whose title/desc names a language or a language-specific API should be gated.)

## Fix

Set `exts` on every language-specific rule (mirroring the existing `cSourceExts` pattern already used for the C rules). Language-agnostic rules (generic credential, private-key block) keep `exts: nil`.

## Repro

Run `synapse-cli scan <java-project>`; observe a `sqlalchemy-raw-sql-dynamic` (Python) finding on a `.java` file.

## Acceptance

- Python/JS/TS-idiom rules only match their own file extensions.
- A table test asserts each language-scoped rule does **not** fire on a foreign extension.

nghiadaulau:
Fixes #122 and #123 — two SAST false-positive classes, verified on a real Java monorepo.

## #122 — language-specific rules fired on foreign files

The `rule` struct already had an `exts` language gate (`appliesTo`), but seven single-ecosystem rules left it nil and so ran on every file. The clearest symptom: the `sqlalchemy-raw-sql-dynamic` rule, titled *"Python SQLAlchemy…"*, fired on `.java` files.

Gated:
- `sqlalchemy-raw-sql-dynamic` → Python (`.py`, `.pyi`, `.pyw`)
- `prisma-raw-sql-unsafe`, `possible-idor-prisma-id-only`, `child-process-exec-template`, `unsafe-deserialization-node-serialize`, `react-dangerous-html`, `dom-xss-inner-html` → JS/TS family (`.js .jsx .mjs .cjs .ts .tsx`)

Language-agnostic rules (generic credential, private key, `generic-sql-dynamic-execute`, …) stay `exts: nil`.

## #123 — `generic-sql-dynamic-execute` matched non-SQL `.execute()`

The taint-evidence group contained a bare literal `.`, so any `x.execute(y.z(...))` matched — most notably `java.util.concurrent` `Executor.execute(Runnable)` and worker dispatch — and was reported as CWE-89 (High). The evidence now requires a request/interpolation marker (`req.`/`request.`/`params[`/`$_GET`/`${`/`#{`) **or** a quoted string literal followed by concatenation (`"..." +`). The bare `execute(` sink stays (JDBC), so `stmt.executeQuery("SELECT … " + q)` still fires.

## Verification (real Java monorepo, not in this repo)

Before: 8 SAST findings including 5 false positives (2× SQLAlchemy-on-Java, 3× `Executor.execute`).
After: those exact 5 disappear; the mass-assignment + hardcoded-credential findings and both secret findings remain. Zero new findings.

Tests: extended the rule corpus (executor negatives + a JDBC positive), set `file:` on the now-gated corpus entries, and added `TestLanguageGatedRulesRespectExtensions` (each gated rule fires in its own language, silent in a foreign one). `go build ./...`, `go vet`, and the full suite pass.

### DIFF
--- a/internal/infrastructure/tools/sast/analyzer.go
+++ b/internal/infrastructure/tools/sast/analyzer.go
@@ -264,10 +264,14 @@ func sourceFromContextBlock(lower string, line int) (source, evidence string) {
 }
 
 func (a *Analyzer) findingFromRule(rel string, line int, ruleID string, lines []string, project projectContext) (ports.SASTRawFinding, bool) {
+	ext := strings.ToLower(filepath.Ext(rel))
 	for ri := range a.rules {
 		r := &a.rules[ri]
 		if r.id != ruleID {
 			continue
+		}
+		if !r.appliesTo(ext) {
+			return ports.SASTRawFinding{}, false // honor the language gate on the contextual path too
 		}
 		h := ports.SASTRawFinding{
 			File: rel, Line: line, RuleID: r.id, CWE: r.cwe,
--- a/internal/infrastructure/tools/sast/analyzer_test.go
+++ b/internal/infrastructure/tools/sast/analyzer_test.go
@@ -187,7 +187,10 @@ func TestRuleCorpus(t *testing.T) {
 			tp: []string{
 				`cursor.execute("SELECT * FROM users WHERE id=" + request.args["id"])`,
 				`mysqli_query($db, "SELECT * FROM users WHERE id=".$_GET["id"])`,
-				`stmt.executeQuery("SELECT * FROM users WHERE id=" + id)`, // JDBC bare-execute TP survives the fix
+				`stmt.executeQuery("SELECT * FROM users WHERE id=" + id)`,          // JDBC bare-execute TP survives the fix
+				`mysqli_query($db, "SELECT * FROM users WHERE n='" . $name . "'")`, // PHP indirect concat via `.$`
+				`cursor.execute("SELECT * FROM users WHERE n='{}'".format(name))`,  // Python str.format
+				`cursor.execute(f"SELECT * FROM users WHERE id={uid}")`,            // Python f-string
 			},
 			fp: []string{
 				`cursor.execute("SELECT * FROM users WHERE id=?", [id])`,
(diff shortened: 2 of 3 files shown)


nghiadaulau:
Addressed review feedback (go-arch + security):
- **go-arch:** `possible-idor-prisma-id-only` was also emitted via the contextual (multi-line block) path, which bypassed the `exts` gate. `findingFromRule` now honors `appliesTo`; covered by a contextual-path idor case.
- **security (recall):** added `.mts`/`.cts` + SFC `.vue`/`.svelte`/`.astro` to `jsExts` (were silently skipped after scoping); re-added PHP `.`-concat and Python `.format`/f-string to `generic-sql-dynamic-execute` as quote-anchored markers so those true positives survive without reintroducing the `Executor.execute` FP. Bare `%` stays out (constant `LIKE '%x%'` FP), locked by a new negative case.

Full suite green; the real Java monorepo still yields exactly the 3 SAST + 2 secret true positives, with the 5 SQL false positives gone.

PR #131, opened 2026-07-11
feat(misconfig): scan docker-compose and GitHub Actions workflows (#126)

### LINKED ISSUE #126: Misconfig: add docker-compose and GitHub Actions workflow scanners
## Summary

The misconfiguration scanner (`internal/infrastructure/tools/misconfig/`) covers Dockerfile, Kubernetes, Helm, Terraform and CloudFormation, but not two very common sources in real repos:

- **`docker-compose.yml` / `compose.yaml`** — privileged containers, host network mode, `:latest` images, bind-mounting the Docker socket, host path mounts, `cap_add`, secrets in `environment:`.
- **GitHub Actions workflows (`.github/workflows/*.yml`)** — unpinned third-party actions (`uses: org/action@main` instead of a commit SHA), `pull_request_target` with checkout of untrusted refs, broad `permissions`, secrets echoed to logs, script injection via `${{ github.event.* }}` in `run:`.

A monorepo can easily have a dozen workflow files and several compose files, all currently unscanned for misconfiguration.

## Proposal

Add two scanners in the existing `misconfig` package, dispatched by path/filename in `scanner.go`:

- `compose.go` — parse compose services, flag the checks above.
- `githubactions.go` — parse workflow YAML, flag unpinned actions / risky triggers / permissions / injection sinks.

Emit `ports.MisconfigRawFinding` with file:line, matching the existing rule shape (e.g. `dockerfile-run-as-root`).

## Acceptance

- `docker-compose.yml` with `privileged: true` / socket mount / `:latest` produces findings.
- A workflow using `uses: some/action@main` or `pull_request_target` + untrusted checkout produces findings.
- Rules carry stable IDs, CWE mapping, and file:line; covered by table tests with sample fixtures.

nghiadaulau:
Fixes #126. Adds two line-based scanners to the misconfig package (which previously covered only Dockerfile/K8s/Helm/Terraform/CloudFormation):

**compose.go** – docker-compose/compose files (conventional name or top-level `services:` sniff): privileged, host network/pid/ipc, mounted Docker socket, dangerous `cap_add`, unpinned image (`:latest`/no tag), literal secrets in `environment` (interpolated `${VARS}` are not flagged).

**githubactions.go** – workflows under `.github/workflows/*.yml|yaml` (path-gated): `pull_request_target`, `permissions: write-all`, third-party actions not pinned to a 40-char commit SHA, and `${{ github.event.* }}`/`${{ github.head_ref }}` interpolated into a `run:` script (indentation-tracked).

Findings carry exact file:line. Table tests cover TP + FP (hardened compose, SHA-pinned + env-var-passed workflow) and the path gate. Full misconfig suite + vet green.

### DIFF
--- a/internal/infrastructure/tools/misconfig/compose.go
+++ b/internal/infrastructure/tools/misconfig/compose.go
@@ -1,0 +1,215 @@
+package misconfig
+
+import (
+	"regexp"
+	"strings"
+
+	"github.com/KKloudTarus/synapse-ce/internal/domain/shared"
+	"github.com/KKloudTarus/synapse-ce/internal/usecase/ports"
+)
+
+// isComposeName recognises the conventional Docker Compose file names.
+func isComposeName(name string) bool {
+	n := strings.ToLower(name)
+	switch n {
+	case "docker-compose.yml", "docker-compose.yaml", "compose.yml", "compose.yaml":
+		return true
+	}
+	// docker-compose.<env>.yml / compose.override.yaml, etc.
+	return (strings.HasPrefix(n, "docker-compose.") || strings.HasPrefix(n, "compose.")) &&
+		(strings.HasSuffix(n, ".yml") || strings.HasSuffix(n, ".yaml"))
+}
+
+var reComposeServices = regexp.MustCompile(`(?m)^services:\s*$`)
+
+// looksCompose sniffs a non-conventionally-named YAML file that is a Compose file: a top-level
+// `services:` mapping plus at least one service-shaped key. It intentionally does not fire on a
+// Kubernetes manifest (which the caller has already ruled out) since those never have top-level services.
+func looksCompose(data []byte) bool {
+	s := string(data)
+	if !reComposeServices.MatchString(s) {
+		return false
+	}
+	return strings.Contains(s, "image:") || strings.Contains(s, "build:") || strings.Contains(s, "container_name:")
+}
+
+var (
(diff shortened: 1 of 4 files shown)


nghiadaulau:
Addressed all review findings (security + go-arch):

**Security:**
- **A (HIGH FP):** the `run:`/`script:` block is now tracked by the key's own column (not the list-dash indent), so a sibling `env:` *after* `run:` (the recommended injection mitigation) is no longer misread as script content and false-flagged.
- **B (FN):** inline `cap_add: ["SYS_ADMIN"]` now strips quotes (matched the list form only).
- Added `security_opt` seccomp/apparmor=unconfined + `label:disable`, `userns_mode: host`, `actions/github-script` `script:` sinks, and broadened the injectable-field suffix.

**go-arch:**
- Gated the trigger/permissions/uses checks with `!inRun` (YAML echoed inside a run script no longer false-positives); strip trailing `#` comments (quote-aware) before trigger matching.
- Compose evaluates `environment:` entries before the directive regexes, so an env key like `PRIVILEGED: true` isn't misclassified; service header matched via regex (anchor-tolerant); exact `environment:` header.
- gofmt-clean.

Tests extended for every fixed case. Full suite (118 pkgs) + vet green. On the real Java monorepo the scanners still surface the true misconfigs (13 unpinned `@main` actions, compose secrets/unpinned images, a docker-socket mount) and the env-routing/comment/env-key cases correctly do not fire.

PR #134, opened 2026-07-11
feat(sca): static build.gradle direct-dep fallback + JDK auto-detection (#124)

### LINKED ISSUE #124: SCA: static direct-dependency fallback + JDK auto-detection for unresolved Gradle projects
## Summary

When a Gradle project cannot be resolved (invalid/missing JDK, offline, private/unreachable artifact repo, or an incompatible Gradle version), the SCA scan resolves **0 components** and therefore reports **0 vulnerabilities** — even though the declared dependencies are sitting right there in `build.gradle` and the version catalog.

The scan already prints an honest completeness warning in this case, which is good:

> Unresolved build system(s) present: gradle. Their dependencies are NOT fully captured … result is INCOMPLETE …

But we can do materially better than zero by falling back to the statically declared coordinates.

## Proposal

1. **Static direct-dependency fallback.** When resolution fails, parse declared coordinates from:
   - `build.gradle` / `build.gradle.kts` `dependencies { }` blocks, and
   - `gradle/libs.versions.toml` version catalogs (`[versions]` + `[libraries]`).
   Emit these as **direct** SBOM components (clearly marked "declared, transitive tree not resolved") so advisory matching can run on the direct set. This still can't see transitive deps, but a pinned vulnerable direct dependency becomes detectable instead of invisible.

2. **JDK auto-detection.** The resolver failed here purely because `JAVA_HOME` pointed at an absent JDK. Prefer the Gradle wrapper's toolchain, then a discovered JDK (`/usr/libexec/java_home`, common install dirs), before giving up. Surface a `JAVA_HOME`/toolchain error as a first-class, actionable warning (it is currently buried in `source_warnings`).

## Acceptance

- A Gradle project with no lockfile and no working JDK still yields direct-dependency SBOM components + any advisory matches on them.
- Components from the static fallback are flagged as declared-only (completeness stays honest: still `confident: false`).
- The `JAVA_HOME`/toolchain failure is reported prominently, with the remediation (`gradle dependencies --write-locks`) that already appears in the warning.

nghiadaulau:
Fixes #124. When a Gradle project can't be fully resolved (no JDK, offline, private/unreachable repo), SCA saw ~0 usable components and reported 0 vulns despite concrete-version deps declared in the build script.

**Static fallback** — new owned parser `ownsbom.BuildGradle` (markers `build.gradle`/`build.gradle.kts`) extracts inline `dependencies { }` Maven coordinates with a **literal** version (version-less BOM-managed + `${interpolated}` are skipped — not advisory-matchable). Registered in the owned registry + wired into the Syft-enrichment path (mirrors the existing `libs.versions.toml` parser). Registry dedups by PURL, so no double-count with the catalog/resolver.

**JDK auto-detection** — on the trusted-local direct-exec path the gradle resolver injects a discovered `JAVA_HOME` (macOS `java_home`, then well-known JDK roots) when the inherited one is missing/invalid, so resolution succeeds instead of failing with "JAVA_HOME is set to an invalid directory"; if none is found the env is untouched and Gradle's own error is still surfaced.

**Verified on a real Java monorepo** (no lockfile, unreachable private repo): SBOM grows **12 → 47** components and SCA reports **4 real CVEs** (commons-fileupload 1.5 HIGH, commons-lang3 3.12.0, bcpkix-jdk15on 1.70 ×2) where it previously found none. Completeness warning stays (confident:false — transitive tree not captured). Full suite (118 pkgs) + vet + gofmt green.

### DIFF
--- a/internal/infrastructure/tools/gradleresolve/javahome_test.go
+++ b/internal/infrastructure/tools/gradleresolve/javahome_test.go
@@ -1,0 +1,72 @@
+package gradleresolve
+
+import (
+	"os"
+	"path/filepath"
+	"strings"
+	"testing"
+)
+
+func makeJDK(t *testing.T) string {
+	t.Helper()
+	jdk := t.TempDir()
+	if err := os.MkdirAll(filepath.Join(jdk, "bin"), 0o755); err != nil {
+		t.Fatal(err)
+	}
+	if err := os.WriteFile(filepath.Join(jdk, "bin", "java"), []byte("#!/bin/sh\n"), 0o755); err != nil {
+		t.Fatal(err)
+	}
+	return jdk
+}
+
+func TestJavaHomeValid(t *testing.T) {
+	if !javaHomeValid(makeJDK(t)) {
+		t.Errorf("a dir with bin/java must be a valid JAVA_HOME")
+	}
+	if javaHomeValid(t.TempDir()) {
+		t.Errorf("an empty dir must not be a valid JAVA_HOME")
+	}
+	if javaHomeValid("") {
+		t.Errorf("empty string must not be valid")
+	}
+}
+
+func TestEnsureJavaHome(t *testing.T) {
+	jdk := makeJDK(t)
+
+	// A valid JAVA_HOME is kept and detect() is never consulted.
+	env := ensureJavaHome([]string{"PATH=/usr/bin", "JAVA_HOME=" + jdk}, func() string {
+		t.Fatal("detect must not run when JAVA_HOME is already valid")
+		return ""
+	})
+	if envValue(env, "JAVA_HOME") != jdk {
+		t.Errorf("valid JAVA_HOME should be kept, got %q", envValue(env, "JAVA_HOME"))
+	}
+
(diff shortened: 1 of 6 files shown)


nghiadaulau:
Addressed review (security + go-arch); golden rules were confirmed clean, these were detection-correctness fixes:
- **buildscript classpath leak (go-arch MAJOR):** rewrote `ParseBuildGradleDeps` as a comment/string-aware scanner with a block-name stack. `buildscript { dependencies { classpath ... } }` coords are no longer emitted as production components (they were phantom production CVEs).
- **dynamic version (security MEDIUM):** `concreteVersionRE` drops `+`, so `1.+`/`2.+` are rejected instead of collapsing to a bare major and phantom-matching.
- **trailing/inline comments (both):** `//` and `/* */` inside a dependency line no longer contribute coords or miscount braces.
- `@ext` (1.0@aar) stripped + 4th classifier part ignored (recovers those TPs).
- `javaHomeValid` now requires `bin/javac` (JDK, not JRE); `detectJDK` is ctx-aware; added method doc comments.

Tests added for every case (buildscript/comment/dynamic exclusion, splitGradleCoord, JRE rejection). Victim rescan unchanged (47 components, 4 real CVEs); full suite + vet + gofmt green.

PR #136, opened 2026-07-11
feat(sca): resolve a lockfile-less package.json (zero-setup JS dependency SCA)

nghiadaulau:
Makes Synapse zero-setup for JS SCA. A raw JS source usually ships `package.json` but no committed lockfile, so its deps are declared only as semver ranges — the SBOM had no resolvable version and SCA reported **nothing** (the same "you must prepare the source first" friction as Trivy, and the same class as the earlier Gradle limit fixed in #124).

**New `npmresolve` adapter** (`ports.NPMResolver`) runs the package manager in `--package-lock-only --ignore-scripts` mode over a **throwaway copy** of `package.json`, resolving the full transitive tree into a lockfile **without** downloading `node_modules` and **without** running any lifecycle script, then reuses the owned lockfile parser to emit pinned `pkg:npm` components.

**Safety**
- `--ignore-scripts` → no `preinstall`/`install`/`postinstall` project code runs.
- Runs against a temp copy → the user's project is never mutated.
- Pinned binary (never a project-vendored one); `SYNAPSE_*` env scrubbed from the child.
- CLI: default-on / direct-exec (trusted-local dogfood) with a stderr transparency note.
- API: enabled **only** with the SCA sandbox (egress confined to the registry), **fails closed** otherwise — never host-execs over an untrusted target.
- Best-effort + opt-in (`SYNAPSE_NPM_RESOLVE_ENABLED`): no `package.json` / a committed lockfile / missing binary / any error → no-op, scan never fails. `mergeResolvedNPM` drops only the generator's unversioned placeholders before folding in the resolved tree (dedup by PURL).

**Verified E2E** — a `package.json` declaring `express@4.17.1` + `lodash` with **NO lockfile**:
```
SBOM components: 51 (pkg:npm, full transitive tree)
vulnerabilities: 11  (4 high: qs, path-to-regexp x2, body-parser; + medium/low)
```
where Synapse/Trivy previously found **zero** on the same raw folder.

Unit tests: no-op paths (no manifest / committed lockfile present), script-safe args (`--ignore-scripts` + `--package-lock-only` required), `SYNAPSE_*` scrubbing. `go build ./...`, `go vet`, `CGO_ENABLED=0 build ./cmd/...`, and the full suite (119 pkgs) pass.

### DIFF
--- a/cmd/synapse-api/main.go
+++ b/cmd/synapse-api/main.go
@@ -59,6 +59,7 @@ import (
 	"github.com/KKloudTarus/synapse-ce/internal/infrastructure/tools/mavencoord"
 	"github.com/KKloudTarus/synapse-ce/internal/infrastructure/tools/mavenresolve"
 	"github.com/KKloudTarus/synapse-ce/internal/infrastructure/tools/misconfig"
+	"github.com/KKloudTarus/synapse-ce/internal/infrastructure/tools/npmresolve"
 	"github.com/KKloudTarus/synapse-ce/internal/infrastructure/tools/nvd"
 	"github.com/KKloudTarus/synapse-ce/internal/infrastructure/tools/ospkg"
 	"github.com/KKloudTarus/synapse-ce/internal/infrastructure/tools/osv"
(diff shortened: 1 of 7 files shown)


nghiadaulau:
Addressed review (security **SOUND** / go-arch **PASS** — no critical/high/medium; these were the low findings):
- **go-arch:** renamed `NpmBin`/`NpmRegistryHosts` → `NPMBin`/`NPMRegistryHosts` (match the `NPMResolver` all-caps initialism, revive/ST1003); corrected the `Resolve` doc (missing binary / any error → `(nil, err)` surfaced as a source warning, never a silent no-op nor partial; only the top-level dir is inspected).
- **security L1:** broadened the direct-exec env scrub (`scrubSensitiveEnv`) to also drop common credential vars (`NPM_TOKEN`, `NODE_AUTH_TOKEN`, `GITHUB_TOKEN`, `AWS_SECRET_ACCESS_KEY`, ...) — defense-in-depth so a resolution subprocess can't read host tokens (the sandbox path already runs a non-inherited env). L2 (cosmetic dup, errs-safe) and L3 (unbounded direct-exec buffer, --loglevel=error keeps it tiny) left as accepted low-risk notes.

Both reviewers confirmed: argv-only exec, `--ignore-scripts` always passed, throwaway-copy isolation, sandbox-gated + fail-closed API wiring, `HOME=work` blocks `~/.npmrc` token reads, dependency-rule PASS. E2E re-verified post-fix: lockfile-less package.json → 51 pkg:npm components + 11 real CVEs. Full suite (119 pkgs) + vet + gofmt green.

PR #139, opened 2026-07-11
fix(sca): resolve composite / independent-build Gradle monorepos

nghiadaulau:
The Gradle resolver stopped at the FIRST dir with a build script and ran `gradle -p <root> dependencies --configuration runtimeClasspath` there. Two common layouts broke:

- **Aggregator root** (`build.gradle` applying only umbrella/format plugins, no java plugin) has no `runtimeClasspath` → the whole resolve failed with "configuration 'runtimeClasspath' not found" → 0 captured.
- **Composite build** whose root `settings.gradle` uses `includeBuild` per module (a monorepo of INDEPENDENT service builds, each with its own settings.gradle + wrapper) was never descended into — `gradle dependencies` on the aggregator reports only the root project, not the included builds where the real trees live.

`buildRoots` now treats every dir with a **settings file** as an independent build root and keeps descending to find nested included/composite builds; a `build.gradle`-only dir is a standalone root only when NOT under an enclosing settings-root (else it's a subproject the enclosing build covers). An aggregator root with no `runtimeClasspath` is skipped, not surfaced as a failure. Tests cover single / flat-monorepo / standard-multiproject / composite layouts.

**Impact:** a service monorepo that previously resolved **0** transitive deps (only the static build.gradle fallback) now resolves each service's full tree. On a real 4-service Spring-Boot composite repo, **one service alone resolves ~470 transitive coordinates** that were entirely invisible — turning a handful of direct-dep CVEs into the real transitive count (full re-scan number to follow in a comment).

### DIFF
--- a/internal/infrastructure/tools/gradleresolve/resolver.go
+++ b/internal/infrastructure/tools/gradleresolve/resolver.go
@@ -115,6 +115,9 @@ func (r *Resolver) Resolve(ctx context.Context, dir string) ([]sbom.Component, e
 		}
 		out, err := r.run(ctx, root)
 		if err != nil {
+			if isNoRuntimeClasspath(err) {
+				continue // aggregator / non-Java build root (no runtimeClasspath) – expected, not a failure
+			}
 			if firstErr == nil {
 				firstErr = fmt.Errorf("%s: %w", relOrBase(dir, root), err)
 			}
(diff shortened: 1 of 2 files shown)


nghiadaulau:
Full re-scan number (composite fix, ignoring the unreachable internal Nexus — those com.finx deps show as FAILED but do not block the public tree):

**kyc-service alone: 335 components resolved, 112 vulnerabilities** (5 critical, 44 high, 52 medium, 11 low; 39 distinct vulnerable components; 1 KEV/known-exploited — e.g. CVE-2025-24813 Tomcat RCE, CVE-2024-38819/38816 Spring path traversal).

Before this fix the resolver stopped at the aggregator root (`includeBuild` composite → not descended) and reported **4** CVEs from the static build.gradle fallback only. After it, one service alone surfaces **112**. The full 4-service monorepo union is larger still. go-arch review addressed (maxBuildRoots cap enforced on the standalone path, sentinel-classified no-runtimeClasspath on raw stderr, dead code removed).

PR #140, opened 2026-07-11
feat(sca): resolve lockfile-less composer.json / Gemfile / pyproject.toml

nghiadaulau:
Completes the zero-setup SCA roadmap: generalizes the npm resolver to the long tail of lockfile-only ecosystems whose human manifest declares only version ranges (so the SBOM had no resolvable version to advisory-match).

New generic `manifestresolve` adapter (`ports.ManifestResolver`) + a spec table:

| ecosystem | manifest | resolve (lock-only, no scripts) | owned parser |
|---|---|---|---|
| composer | composer.json | `composer update --no-install --no-scripts --no-plugins` | Composer (composer.lock) |
| gem | Gemfile | `bundle lock` | Gem (Gemfile.lock) |
| poetry | pyproject.toml | `poetry lock` | Poetry (poetry.lock) |

Each runs the ecosystem tool over a **throwaway copy** of the manifest, then reuses the owned lockfile parser to emit pinned components.

**Safety** (mirrors the npm resolver): no project scripts run, the source is never mutated, a pinned tool, `SYNAPSE_*`/credential env scrubbed; CLI default-on/direct-exec (trusted-local, transparency note), API **sandbox-gated + fail-closed** (egress → the ecosystem registry). Best-effort + opt-in (`SYNAPSE_MANIFEST_RESOLVE_ENABLED`): no manifest / committed lockfile / missing tool / any error is a no-op, scan never fails.

SCA service holds a slice of resolvers (`AddManifestResolver`) and folds each via a generic `mergeResolvedManifest` that drops the generator's unversioned placeholders of the resolved ecosystem(s) (derived from the PURL type) and dedups by PURL. Unit tests cover no-op paths, the no-scripts/lock-only args per ecosystem, and credential-env scrubbing. Full suite (120 pkgs) + vet + `CGO_ENABLED=0 build ./cmd/...` green.

### DIFF
--- a/cmd/synapse-api/main.go
+++ b/cmd/synapse-api/main.go
@@ -56,6 +56,7 @@ import (
 	"github.com/KKloudTarus/synapse-ce/internal/infrastructure/tools/licensefile"
 	"github.com/KKloudTarus/synapse-ce/internal/infrastructure/tools/licensemeta"
 	"github.com/KKloudTarus/synapse-ce/internal/infrastructure/tools/manifest"
+	"github.com/KKloudTarus/synapse-ce/internal/infrastructure/tools/manifestresolve"
 	"github.com/KKloudTarus/synapse-ce/internal/infrastructure/tools/mavencoord"
 	"github.com/KKloudTarus/synapse-ce/internal/infrastructure/tools/mavenresolve"
 	"github.com/KKloudTarus/synapse-ce/internal/infrastructure/tools/misconfig"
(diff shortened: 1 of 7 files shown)


nghiadaulau:
Addressed review (security + go-arch):
- **security HIGH (gem RCE):** correct — `bundle lock` evaluates the Gemfile as Ruby (project code), unlike composer (`--no-scripts --no-plugins`) / poetry (`lock` over inert TOML). Fixed the false "no project code runs" claim: the docstring now scopes it per-ecosystem, and the CLI banner gives gem an explicit **UNSANDBOXED Gemfile-evaluates-Ruby** warning mirroring the Gradle-resolver banner (trusted-local assumption). The API path is unchanged — sandbox-gated + fail-closed, which the reviewer confirmed correctly contains it. The spec-safety test no longer implies gem is code-isolated.
- **security L3 (creds):** scrub `COMPOSER_AUTH` + `BUNDLE_*` from the direct-exec child env, and pin `COMPOSER_HOME`/`BUNDLE_USER_HOME`/`POETRY_CACHE_DIR`/`XDG_CACHE_HOME`/`XDG_CONFIG_HOME` inside the throwaway dir (full hermeticity, not just HOME) so no host credential file/cache is read.
- **go-arch:** PASS (dependency rule + idioms); the hermeticity note is covered by the cache-dir pins above; `New`'s forgiving unknown-ecosystem no-op is intentional (both roots hardcode the set) — the reviewer confirmed `mergeResolvedManifest` PURL-prefix derivation is correct for composer/gem/pypi.

Reviewers confirmed composer/poetry are safe to ship as-is and the API path is correct. Full suite (120 pkgs) + vet + gofmt green; gem E2E: a lockfile-less Gemfile resolves to its gems + CVEs.

PR #155, opened 2026-07-12
feat(recon): canonicalize domain and URL scope targets

pho-veteran:
## Summary

Harden domain, host, endpoint, and URL handling across reconnaissance scope authorization, tool execution, parser output, and sandbox egress policy compilation.

This change replaces independent trim/lowercase/manual-authority parsing paths with one engagement-domain normalization contract. It preserves the existing execution and scope safety model while intentionally making `url` scope entries least-privilege: a URL scope now authorizes its exact canonical **scheme + host + effective port**. Operators who intentionally need host-wide authorization should use a `domain` scope entry.

## Example: URL scope no longer broadens a probe

**Before:** An operator scoped a scan to `https://shop.example.com:8443/cart?preview=1`. Recon reduced that value to `shop.example.com`, so HTTPX could select a different scheme or default port. The egress policy separately restricted traffic to port `8443`, creating inconsistent authorization, execution, and network-policy interpretations.

**After:** The same scope is canonicalized and treated consistently:

```text
Scope / authorized target / HTTPX argv: https://shop.example.com:8443/cart?preview=1
Egress rule after hostname resolution:  shop.example.com resolved IPs on TCP/UDP port 8443
```

Requests such as `http://shop.example.com/...`, `https://shop.example.com/...` on the default port, or `https://user@shop.example.com/...` are denied. To intentionally authorize the host across schemes and ports, the operator uses a `domain` scope such as `shop.example.com`.

### Canonical engagement normalization

- Added `internal/domain/engagement/normalize.go` as the shared pure normalization boundary for:
  - DNS domains;
  - wildcard domain scope patterns;
  - IP literals;
  - `host:port` endpoints, including bracketed IPv6;
  - absolute HTTP(S) URLs and their authorization identity.
- Added IDNA Lookup ASCII canonicalization through `golang.org/x/net/idna`.
- Canonicalized case and removed one trailing DNS root dot, so equivalent forms such as `WWW.Example.COM.`, `www.example.com`, and supported Unicode/punycode equivalents resolve to one identity.
- Validated DNS names before they can reach scope matching or argv construction:
  - required non-empty labels;
  - 63-byte label limit;
  - 253-byte full-name limit;
  - LDH label rules and hyphen placement;
  - invalid UTF-8 and ASCII control-character rejection;
  - IP literals rejected where a domain is required.
- Restricted `*.` wildcard handling to stored `domain` scope entries. Wildcards are rejected for execution targets and parser observations.

### URL and authority hardening

- Replaced manual scheme, userinfo, path, port, and IPv6 parsing with `net/url`, `net.SplitHostPort`, and `netip` behind the shared engagement helper.
- Require absolute `http` or `https` URLs for `TargetURL`.
- Reject URL userinfo, missing hosts, opaque URLs, unsupported schemes, malformed/zero/out-of-range ports, and ambiguous authority forms before execution.
- Defined `TargetURL` scope semantics as exact canonical `(scheme, host, effective port)` matching:
  - `https://app.example.com` authorizes HTTPS on port 443;
  - it does not authorize `http://app.example.com` or `https://app.example.com:8443`;
  - a `domain` scope remains the explicit host-wide authorization mechanism.
- Preserved canonical URL path and query in HTTPX arguments instead of silently reducing a URL target to a host token.

### Scope, recon, and parser integration

- Canonicalize scope targets in `Engagement.SetScope` before persistence, without partially changing an engagement when any submitted target is invalid.
- Refactored scope host extraction and matching to fail closed on malformed input while preserving:
  - out-of-scope precedence;
  - CIDR containment;
  - exact-domain and wildcard-subdomain behavior;
  - wildcard apex exclusion;
  - local repository path containment;
  - no DNS resolution during scope matching;
  - family-strict IPv4-mapped IPv6 behavior.
- Removed private `inferKind` and `validateTargetValue` copies from recon orchestration. Recon now uses `engagement.InferTargetKind` and `engagement.ValidateTargetValue`.
- Normalize recon targets before submit-time authorization, queue persistence, worker reauthorization, and argv construction.
- Normalize parser output at the recon adapter boundary:
  - Subfinder observations become canonical domains;
  - HTTPX observations become canonical HTTP(S) URLs;
  - Naabu results use canonical endpoint formatting with validated port numbers.

### Egress-policy preservation of URL port restrictions

- Extended `ports.EgressPolicy` with structured deferred `DomainRule` entries.
- URL hostnames retain their effective port after host-side DNS resolution and into generated kernel egress rules.
- Preserved hostname-wide behavior for existing `AllowDomains` / `DenyDomains` consumers.
- Preserved deterministic deny-before-allow ordering, default-deny behavior, pinned-host resolution, and IPv4-mapped IPv6 narrowing.

### Tests and documentation

- Added table-driven and fuzz coverage for domain, host, endpoint, wildcard, IDNA, IPv6, URL, and malformed-authority handling.
- Updated scope, engagement, recon adapter, and egress-policy tests for canonical identities and URL least-privilege semantics.
- Updated `docs/TODO.md` resolution records for TODO items 1 and 2; this documentation file is intentionally not included in this commit.

### Preserved safety invariants

- argv-only execution; no shell command construction.
- Leading-dash and whitespace flag-injection defenses before argv construction, including adapter-level backstops.
- Submit-time checks and worker-time server-side authorization.
- `LiveReconEnabled`, RoE, authorization-window, and out-of-scope precedence behavior.
- No DNS resolution in normalization or scope matching.
- Bounded execution and output capture.
- Append-only evidence sealing and containment/connect-log flow.
- Naabu's capability-free `-s c` connect scan rather than raw-socket SYN scanning.
- Deterministic egress ordering and fail-closed IPv6 behavior.

### Intentional behavior changes

- Invalid DNS names and malformed URL authorities are rejected earlier rather than reaching scope comparison or tools.
- `TargetURL` is no longer host-wide authorization. It is bound to the canonical scheme and effective port.
- HTTPX receives a canonical full URL for `TargetURL`, preserving operator-selected scheme, port, path, and query.
- There is no schema migration: existing `scope_targets.value` storage is retained, but future scope writes store canonical values.

## Verification

Completed successfully:

- `go test ./...`
- `go vet ./...`
- `go test ./internal/domain/engagement/... ./internal/infrastructure/recon/... ./internal/usecase/recon/... ./internal/usecase/egress/... ./internal/infrastructure/egress/...`
- `go test -fuzz=FuzzNormalizeDomain -fuzztime=5s ./internal/domain/engagement/...`
- `go test -fuzz=FuzzNormalizeURL -fuzztime=5s ./internal/domain/engagement/...`
- `git diff --check`
- `make build vet typecheck` completed the Go build and vet steps; the frontend typecheck not required for this changes.


nghiadaulau (changes requested):
Strong, careful work — unifying domain/host/URL handling into one canonicalization contract is the right direction, and most of it is sound. Two independent reviews (security + clean-architecture) confirmed: the canonicalization is fail-closed against the attacks that matter (userinfo `https://app@evil.com` rejected, non-http/https rejected, control chars / invalid UTF-8 rejected, trailing-dot and IDNA homoglyph handled, `0x`/octal IP forms not mis-parsed as IPs, explicit-vs-effective port handled, no panics on malformed input); argv-only exec and the scope-before-exec authorization order are preserved; and the **in-scope** URL least-privilege tightening (exact scheme+host+port) is correct and well-tested.

**However, there is a blocking security regression — please do not merge as-is.**

**BLOCKER — out-of-scope (deny) URL entries no longer deny host-wide → a carved-out host becomes reachable.** `matchScopeTarget`'s `TargetURL` case now requires `req.Kind == TargetURL` **and** exact scheme+host+port. That's correct for the in-scope allow list, but `AllowsTarget` runs the **same** matcher over the `OutOfScope` list, so a URL deny entry stops denying broadly. Both reviews reproduced this empirically:

- Scope: in-scope `*.acme.io`, out-of-scope `https://payments.acme.io/`.
- `Allows("payments.acme.io")` → **true** (was false on `main`), `AllowsTarget({URL, "http://payments.acme.io/x"})` → **true**, `AllowsTarget({Domain, "payments.acme.io"})` → **true**.

So a URL carve-out now only blocks the one exact `scheme+host+port`; every other scheme/port, and any non-URL-kind request to the forbidden host, slips through to the broader in-scope entry. This reaches the real execution gate and the recon result path (Golden Rule 2 — "out-of-scope always wins / fail closed").

**MAJOR — the same weakening on the egress backstop.** An out-of-scope URL host is compiled to a port-scoped `DROP` (e.g. only `:443`), where pre-PR it was an all-ports deny, so 22/80/3389/etc. stay open at the kernel for a host that resolves inside an in-scope CIDR.

**Fix (both reviewers agree):** keep the narrow scheme+host+port match only for the **in-scope** loop; make the **out-of-scope** pass host-wide for URL entries (compare host only, ignore kind/scheme/port) and compile out-of-scope hosts with nil `Ports` (all-ports deny). Please add the missing out-of-scope-**URL** test cases (`http://` vs `https://`, bare-host request → denied) — the current tests only cover out-of-scope domain/IP entries, which is why this slipped.

**MEDIUM — deny entry that fails canonicalization silently stops denying.** `matchScopeTarget` returns `false` when `NormalizeTarget` errors; for an in-scope entry that's fail-closed (fine), but for an out-of-scope entry it's fail-**open** (the deny is dropped). `SetScope` now normalizes on write, so new data is safe, but a pre-PR stored scope with a value the stricter normalizer rejects would under-deny. Fail closed on the deny side (treat a non-normalizable out-of-scope entry as matching, or reject the engagement's scope as un-enforceable).

**Minors**
- `normalize.go` is the first `golang.org/x/net` import in `internal/domain/*`. Acceptable in spirit (idna is pure + Go-team-maintained and canonical-domain is a domain concept), but it's a literal deviation from the "domain imports stdlib + domain only" rule — please sanction it explicitly (a dependency-table footnote or a short ADR) rather than leaving it as undocumented precedent.
- `ports/egress.go`: the `EgressPolicy` doc paragraph now sits directly above `type DomainRule` with no blank line, so godoc misattaches it — remove the stale paragraph.
- `scope.go` `Allows` doc still says repo/URL values are "matched exactly", but the generic exact-match was removed and `InferTargetKind` never yields repo/image — update the contract doc.
- Tightened validation now rejects underscore labels (e.g. `_dmarc.example.com`) at scope creation — a fail-closed improvement, but call it out for anyone with existing underscore-bearing scope data.

Requesting changes for the blocker + major (the deny-path weakening) and the missing out-of-scope-URL tests. The rest of the PR is good and I'm glad to re-review promptly once the deny path is host-wide again.

pho-veteran:
## Changes requested addressed

Thank you for the careful security and architecture reviews. I reproduced the deny-path regression and updated the implementation so URL allow and deny entries now intentionally use different semantics:

- **In-scope URL:** exact canonical `scheme + host + effective port` (least privilege).
- **Out-of-scope URL:** host-wide carve-out across request kinds, schemes, and ports.
- **Precedence:** out-of-scope is evaluated first and always wins.

### BLOCKER — restored host-wide URL deny behavior

The out-of-scope loop no longer reuses the strict URL allow matcher. It now applies deny-specific matching:

- `https://payments.acme.io/` denies `https://payments.acme.io/...`;
- it also denies `http://payments.acme.io/...`;
- it denies alternate ports such as `https://payments.acme.io:8443/...`;
- it denies a bare/domain-kind request for `payments.acme.io`;
- it still overrides a broader in-scope entry such as `*.acme.io`.

Malformed allow entries grant nothing. A malformed deny entry makes the scope unenforceable and therefore denies every request rather than silently dropping the carve-out.

Regression coverage was added in `internal/domain/engagement/scope_test.go` for exact URL allows, cross-scheme/port URL denies, bare-host requests, malformed allow/deny entries, CIDR containment/overlap, and container-registry requests.

### MAJOR — restored all-port egress denies

Out-of-scope URL hostnames and literal IPs now compile to deny rules with an empty `Ports` list, which means all ports. In-scope URLs still retain only their effective or explicit port.

The egress tests now verify that:

- a URL allow such as `https://app.example:8443` remains restricted to port `8443`;
- an out-of-scope URL for the same host emits an all-port deny;
- malformed network deny entries produce a fail-closed IPv4 deny-all rule.

### MEDIUM — legacy persisted denies no longer disappear

Loaded `scope_targets` rows now cross the same canonical `NormalizeTarget` boundary as new scope writes. If a legacy row is invalid under the stricter contract, engagement loading fails instead of returning a partially enforceable scope.

Persistence coverage verifies canonical loading of a legacy value such as `Secret.IO.` and rejection of an invalid stored target. The database-backed case remains conditional on `SYNAPSE_TEST_DB_DSN`.

### Minor review items

- Explicitly documented `golang.org/x/net/idna` as the narrow, pure-Go canonical-IDNA exception to the domain dependency rule in `CONTRIBUTING.md` and `docs/guide/architecture.md`; this is not a general exception for third-party domain dependencies.
- Corrected the `DomainRule`/`EgressPolicy` Godoc attachment in `internal/usecase/ports/egress.go`.
- Updated `Scope.Allows` documentation to describe kind inference and direct callers of `AllowsTarget` accurately.
- Documented the compatibility change for underscore-bearing DNS owner names such as `_dmarc.example.com`: they are rejected as scope host targets because they are not LDH A/AAAA hostnames.

## Additional blast-radius fixes

The follow-up QA traced the changed scope contract through adjacent authorization and import paths and found four related regressions, which are also fixed:

1. **Container image registries:** image scans now authorize as `TargetImage`, allowing a URL carve-out for a registry host to block image acquisition before the acquirer runs. Registry extraction mirrors acquisition semantics: unqualified references (`nginx:1`, digests, and `library/nginx`) map to `docker.io`, while explicit registries, registry ports, and `localhost` resolve to their canonical host.
2. **CIDR requests:** a requested CIDR is allowed only when fully contained by an in-scope CIDR and is denied when it overlaps an out-of-scope CIDR.
3. **Transfer imports:** imported engagement scope now passes through `Engagement.SetScope` before the first write, so valid values are canonicalized and invalid bundles are rejected without partial persistence.
4. **Containment evidence:** structured hostname rules (`AllowDomainRules`) are included in the recorded egress destination count.

These additions preserve the existing no-DNS authorization boundary, argv-only execution, submit-time and worker-time authorization, deny-before-allow ordering, and reject-before-write behavior.

## Verification

Completed successfully after the final registry-matching correction:

```text
go test ./...
go vet ./...
git diff --check
go test ./internal/domain/engagement/... ./internal/usecase/sca/...
go test -fuzz=FuzzNormalizeDomain -fuzztime=10s ./internal/domain/engagement/...
go test -fuzz=FuzzNormalizeURL -fuzztime=10s ./internal/domain/engagement/...
```

Fuzz runs completed without failures (63,676 domain executions and 57,606 URL executions). `git diff --check` reported only the repository's Windows LF-to-CRLF conversion warnings, with no whitespace errors.

Frontend typechecking was not rerun because `pnpm` is unavailable on `PATH` in this environment. The Go build, tests, vet, focused regression tests, and fuzz targets all pass.

Pre-existing DNS pinning behavior for wildcard/unresolvable deny rules and current IPv6 egress enforcement limitations were reviewed but intentionally left outside this focused regression fix.

nghiadaulau (approved):
Reviewed and verified locally (CI is off on this repo, so gating on local build + tests).

**Verification**
- `go build ./...`, `CGO_ENABLED=0 go build ./cmd/...`, and `go vet` on every affected package: clean.
- Tests green: `engagement`, `usecase/egress`, `infrastructure/egress`, `infrastructure/recon`, `usecase/recon`, `usecase/sca`, `usecase/transfer`.
- `FuzzNormalizeURL` ~270k execs, no idempotency failure.

**Security assessment** (this touches the authorization boundary, so I read it closely)
- Canonicalization closes the smuggling class: case / trailing-dot / IDNA-punycode, userinfo rejection (`https://app@evil.com`), opaque-URL and IP-literal-as-domain rejection, control-char / invalid-UTF-8 rejection, explicit effective port.
- Scope matching is correctly asymmetric: an in-scope URL is least-privilege (scheme + host + effective port), an out-of-scope URL is host-wide. Fail-closed throughout — a malformed allow grants nothing, a malformed deny denies everything, and a malformed persisted scope makes the engagement unloadable rather than silently dropping a restriction.
- The egress compiler mirrors it: an out-of-scope URL denies the host on all ports (empty `Ports`), an in-scope URL keeps only its effective port, and a malformed network deny compiles to a `0.0.0.0/0` deny-all.
- Recon still builds argv from the canonical value with the `-`-prefix / whitespace flag-injection guard intact, and the pure matcher still resolves no DNS (SSRF / rebinding surface preserved).
- Transfer import now crosses `SetScope` validation before the first write, and SCA image targets now match as `TargetImage` so registry carve-outs apply.

Net effect is a tightening of the boundary, with a test for each property. Nice work.

**One architectural note (not a blocker).** `internal/domain/engagement/normalize.go` introduces `golang.org/x/net/idna` in the domain layer, which the "domain imports only domain + stdlib" rule otherwise forbids. The PR consciously documents this as a narrow, sanctioned exception in `CONTRIBUTING.md` and `docs/guide/architecture.md`, and I agree with the call: `idna` is pure-Go, deterministic, no I/O, and hand-rolling IDNA canonicalization in a security tool would be the worse option. Flagging it only so it stays a deliberate, on-the-record decision.

Approving and merging. Thanks @pho-veteran.

Issue #163, opened 2026-07-12
feat(taint): attach source→sink file:line to taint CapSAST findings (def-use precision)


nghiadaulau:
### Context
Taint `CapSAST` claims carry the vulnerable **symbol** as `Location`, not `path:line` (`internal/usecase/taintscan/coordinator.go`, `internal/domain/taint/catalog.go`) — the def-use precision is deferred. A confirmed taint finding is therefore less actionable than a pattern-SAST finding (which has `file:line`).

### Goal
Attach `path:line` def-use locations (source + sink) to taint CapSAST claims/findings.

### Plan
1. Carry source/sink positions from the `synapse-callgraph` SSA build through `taint.Catalog`/`Assemble` into the `SASTClaim.Location`.
2. Emit the confirmed `Kind=sast` finding with a `file:line` title like the pattern engine.
3. Keep it a coarse over-approximation where exact def-use is unavailable, clearly labeled.

### Acceptance
A confirmed taint finding shows source→sink `file:line`, not just a symbol. Part of #59.

nghiadaulau:
Done in #170 (merged). Taint CapSAST findings now cite source→sink file:line: the SSA builder records first-party symbols' definition positions (relpath:line, no host paths, no contents) into a callgraph.Graph side table, carried across the sandboxed synapse-callgraph wire; the coordinator sets SASTClaim.Location to the sink's file:line (fallback to symbol) + records source/sink positions in the witness. A confirmed taint finding's title now shows file:line like the pattern engine. Coarse function-granular over-approximation, clearly labeled. Verified with real 9router AI (llmverifier assessed the precise-location claim) + a deterministic e2e (taint→verifier-confirm→Kind=sast finding shows file:line). security CLEAN + go-arch PASS.

PR #170, opened 2026-07-12
feat(taint): attach source→sink file:line to taint CapSAST findings (#163)

nghiadaulau:
## #163 — attach source→sink `file:line` to taint CapSAST findings (def-use precision)

Taint `CapSAST` claims carried the vulnerable **symbol** as `Location`, so a confirmed taint finding was less actionable than a pattern-SAST hit (which has `file:line`). This carries def-use positions through the pipeline.

### What changed
- **`callgraph.Graph`** gains an optional `Positions` side table (first-party symbol → `relpath:line`). The reachability queries never read it, so a builder that omits it degrades gracefully to symbol-only locations. Only file **paths + lines** are carried — never file **contents** (GR3), never an absolute host path (relative to the scan root, else the base name).
- **`ssacallgraph.BuildGraph`** records each first-party function's definition position from `prog.Fset` (resolving a generic instance to its origin so the key matches the node id). Bounded to the module's own code — where taint source/sink using-functions live.
- The **`synapse-callgraph` wire envelope** carries `Positions` (additive, back/forward compatible → protocol stays `v1.0.0`).
- **`taintscan.Coordinator`** sets `SASTClaim.Location` to the sink-using function's `file:line` when available (falling back to the symbol) and records source→sink positions in the sealed witness. Coarse, function-granular over-approximation (the definition line, not the exact sink call site) — clearly labeled.

A confirmed taint finding's title now shows `file:line`, e.g. `Taint: taint-command-injection (CWE-78) at internal/app/handler.go:42`.

### Safety
Only paths + lines cross (no contents, no absolute host path, no env/stderr). The taint judgment stays **propose-only + gated** — a distinct verifier ≥ bar still confirms; the Location change is descriptive metadata. The untrusted `synapse-callgraph` output still parses fail-closed.

### Tests
SSA builder computes relative first-party positions (stdlib excluded); wire round-trips `Positions`; coordinator uses them for `Location` + witness (and falls back + omits blank keys without them); a deterministic **e2e** (taint → distinct-verifier confirm → `Kind=sast` finding title shows `file:line`). Plus a gated **live-AI** demo (`SYNAPSE_LIVE_AI=1`): the real taint coordinator proposes a `file:line` claim and a real 9router model (`cx/gpt-5.4`, the `llmverifier`) adversarially assesses it — the precise location reaches the verifier; on confirm the `Kind=sast` finding shows `file:line`. (In one live run the model correctly **refuted** a thin claim — the anti-hallucination gate working, not a rubber stamp.)

🤖 Generated with [Claude Code](https://claude.com/claude-code)


nghiadaulau:
Applied go-arch nits: dropped the misleading `json` tag on `callgraph.Graph.Positions` (the wire type owns serialization), tightened the `..` escape check in `firstPartyPos` (a dir named `..foo` is no longer misclassified), and hoisted the double `Positions` lookup in the coordinator. security-auditor CLEAN; go-arch PASS.

PR #172, opened 2026-07-12
test(fptriage): repeatable live-AI consensus demo for the scan-pipeline FP gate (#165)

nghiadaulau:
## Repeatable live-AI consensus test for the scan-pipeline FP gate (#165)

**Context:** #165 (AI false-positive triage in the scan pipeline, with the #156 distinct-verifier consensus) is already merged, but its "tested with real AI" validation was **ad-hoc** — a one-off CLI run. This adds a **gated, reproducible** live-AI test so that validation is CI-able and regression-proof. **No production code changes.**

### What it proves (the two effectiveness properties)
Driving the real `fptriage.Coordinator` (proposer + a **distinct** verifier) against an OpenAI-compatible gateway:
- a clearly **parameterized** query (not injectable) is refuted by the proposer **and** independently confirmed by the distinct verifier → **consensus suspected-FP** (retain-and-mark, exempt from `--fail-on`);
- a clearly **injectable** string-concatenation is **not** dismissed → it **stays gating** (a real weakness is never hidden).

### Gating / safety
Test-only, `package fptriage`, gated behind `SYNAPSE_LIVE_AI=1` — the hermetic suite is unchanged and `go test ./...` skips it without a live gateway. Reads `SYNAPSE_LLM_API_KEY` from env; never logs it (only model ids + verdicts are logged).

### Live run
proposer `cx/gpt-5.4-mini` + verifier `cx/gpt-5.4`: the parameterized query was refuted **98/99** by both models (verifier `driver=parameterized_query`) and held back; the concatenation was judged **sound (99)** and stayed gating — the two-model consensus works and does not dismiss a real vuln.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/internal/usecase/fptriage/livedemo_ai_test.go
+++ b/internal/usecase/fptriage/livedemo_ai_test.go
@@ -1,0 +1,115 @@
+package fptriage_test
+
+// Live-AI demonstration for #165 (AI false-positive triage in the scan pipeline, with the #156 distinct-
+// verifier consensus). It drives the REAL Coordinator against a live OpenAI-compatible gateway (9router):
+// a clearly PARAMETERIZED query (not injectable) is critiqued as a false positive by the proposer AND
+// confirmed by a DISTINCT verifier → suspected-FP (retain-and-mark, exempt from the gate); a clearly
+// INJECTABLE string-concatenation stays gating (not refuted). This proves the pipeline's FP gate + the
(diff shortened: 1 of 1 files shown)


nghiadaulau:
Applied the go-arch nit: moved to external `package fptriage_test` (the test only touches exported symbols), keeping the infra import fully out of the production package's test-augmented set. go-arch: PASS. Live demo re-verified after the change.

Issue #210, opened 2026-07-12
feat(codequality): XML rule pack (XXE + well-formedness)


nghiadaulau:
Part of #185 (tracker) / #174. **Authoring guide:** [https://github.com/KKloudTarus/synapse-ce/blob/main/docs/guide/code-quality-rules.md](https://github.com/KKloudTarus/synapse-ce/blob/main/docs/guide/code-quality-rules.md). Detection: XML parse (already feasible; no tree-sitter needed). Rules are clean-room, our own content; cite the concept origin, never third-party rule prose.

## Depth plan - parity target ~30 rules

Targets ~30 rules (mature-profile parity), decomposed by rule family - claim a family:

| Family | Target |
|---|---|
| `bugs` | 8 |
| `inj` | 8 |
| `maint` | 14 |
| **Total** | **30** |

**Signature families:** `inj` (XXE / XInclude / entity expansion) is the security core; plus correctness `bugs` (well-formedness) and `maint` (schema hygiene).

## Authoritative sources
[W3C XML](https://www.w3.org/TR/xml/) - [OWASP XXE Prevention](https://cheatsheetseries.owasp.org/cheatsheets/XML_External_Entity_Prevention_Cheat_Sheet.html) - [CWE](https://cwe.mitre.org/)

## Seed rules (initial security + correctness core; fill the families toward the target)
| Key | Type / Sev | Flags | Ref |
|---|---|---|---|
| `xml:external-entity` | vulnerability / high | `DOCTYPE` declaring an external entity (XXE) | [CWE-611](https://cwe.mitre.org/data/definitions/611.html) |
| `xml:xinclude` | vulnerability / high | `XInclude` enabled | [CWE-611](https://cwe.mitre.org/data/definitions/611.html) |
| `xml:entity-expansion` | security_hotspot / medium | recursive entity expansion (billion laughs) | [CWE-776](https://cwe.mitre.org/data/definitions/776.html) |
| `xml:not-well-formed` | bug / medium | malformed XML | W3C |
| `xml:duplicate-attribute` | bug / low | duplicate attribute in an element | W3C |
| `xml:hardcoded-secret` | security_hotspot / medium | secret literal in a config XML | [CWE-798](https://cwe.mitre.org/data/definitions/798.html) |
| `xml:schema-missing` | code_smell / info | no schema/DTD reference for validation | maintainability |

## Tasks
[...]

PR #219, opened 2026-07-13
[codex] feat: add XML well-formedness rules

leemeo3:
## Summary

Adds a small XML bugs/well-formedness slice for #210: parser-backed findings for malformed XML and duplicate attributes.

## Changes

- Add XML file-level codeanalysis rules for `xml:not-well-formed` and `xml:duplicate-attribute`.
- Catalogue the XML rules with metadata, examples, and a `parse` detection mode.
- Add focused XML analyzer tests plus catalog parity/golden inventory coverage.
- Update the code-quality rule authoring guide to include structured parser checks.

Refs #210.

## Validation

- `go test ./internal/domain/rule ./internal/infrastructure/rulecatalog ./internal/infrastructure/tools/codeanalysis ./internal/usecase/codequality`
- `make build`
- `make test`
- `make vet`
- `make typecheck`

## Checklist

- [x] `cd web && pnpm build` passes (if the dashboard changed) - not run; no dashboard/web changes
- [x] Preserves the safety invariants (argv exec, server-side scope enforcement, secrets out of logs, deterministic reports, append-only evidence) - read-only parser/catalog change only

AI assistance: This PR was prepared with OpenAI Codex.

### DIFF
--- a/api/openapi.yaml
+++ b/api/openapi.yaml
@@ -515,6 +515,7 @@ components:
       enum:
         - ast
         - pattern
+        - parse
         - metric
 
     RuleSummary:
--- a/docs/guide/code-quality-rules.md
+++ b/docs/guide/code-quality-rules.md
@@ -22,7 +22,7 @@ copy. Specifically:
   [OWASP](https://owasp.org/), [SEI CERT](https://wiki.sei.cmu.edu/confluence/display/seccode) secure-coding
   standards, and [ISO/IEC 25010](https://www.iso.org/standard/78176.html) software-quality attributes.
 - **Do** write our own rule title, description, rationale, remediation, and code examples, and our own
-  detection (AST query, token/line pattern, or metric threshold).
+  detection (AST query, structured parser check, token/line pattern, or metric threshold).
 - **Do not** copy any third-party rule's text, description, examples, or detection code, and **do not**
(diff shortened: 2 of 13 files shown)


nghiadaulau (approved):
Re-reviewed after the rebase + fix — the blocker is resolved and it's ready. Approving and merging.

**The fix**
- `parse` is now in the `RuleDetection` enum in `api/openapi.yaml`, so the two XML rules no longer serve a detection value outside the published schema.
- Even better, you added a guard test that reads the OpenAPI spec and asserts every `Default()` catalog rule's detection is present in the `RuleDetection` enum — so this can't silently drift again. That's exactly the follow-up I hoped for.

**Verification** (rebased onto current main, golden 143→145):
- `go build ./...`, `CGO_ENABLED=0 go build ./cmd/...`, and `go vet`: clean.
- `go test` on `rule`, `rulecatalog` (including the new OpenAPI-enum guard + the golden inventory at 145), `codeanalysis` (the XML analyzer), and `codequality`: all green.

Everything from the first pass still holds: safe parsing (Go's default `encoding/xml` does no external-entity/DTD resolution and no recursive entity expansion, and the duplicate-attribute scanner is a bounded single pass), and clean-room metadata. Nice work closing the loop. Thanks @leemeo3.

PR #287, opened 2026-07-15
feat(codequality): add project aggregate and product area

pho-veteran:
## Summary

image
image

Closes #175.

Introduce a first-class, long-lived `Project` aggregate and a top-level Code Quality product area, separate from time-bounded pentest engagements.

This change establishes the P1 foundation for project-based code quality:

- users can create, list, and open tenant-scoped Code Quality projects;
- each project stores its source binding, default language profiles, and quality-gate reference;
- project mutations are audited and project routes are protected by the existing authentication/RBAC chokepoint;
- the existing engagement Code Quality tab remains functional;
- engagement and project surfaces share one reusable Code Quality report renderer instead of forking the analysis presentation.

This PR intentionally stops at the project shell boundary defined by #175:

- source acquisition and project analysis execution remain in #176;
- persisted analysis snapshots and activity/history remain in #177;
- real project ratings, overview cards, and gate verdicts remain in #178.

Until those phases land, project cards and project detail pages explicitly show `Not analyzed` / `No analyses yet` rather than fabricating ratings or triggering analysis on page load.

### Project domain

- Add the pure `domain/project` aggregate with:
  - project ID and tenant ownership;
  - human-readable name and lowercase slug key;
  - `local`, `git`, and `archive` source bindings;
  - optional Git branch/tag;
  - default quality-profile mapping by language;
  - quality-gate reference;
  - creation/update audit metadata.
- Validate:
  - required IDs, names, keys, and source values;
  - lowercase hyphenated project keys;
  - supported source kinds;
  - Git-only refs;
  - HTTPS Git sources;
  - conservative Git ref syntax and length;
  - non-empty profile language/profile values.
- Defensively copy profile maps to prevent callers from mutating aggregate state after construction.

### Persistence

- Add the tenant-aware `ProjectRepository` port with:
  - create;
  - list by tenant;
  - get by tenant and key;
  - delete by tenant and key.
- Add an in-memory implementation with:
  - tenant-local key uniqueness;
  - deterministic newest-first listing;
  - copy-on-read and copy-on-write behavior;
  - cross-tenant not-found semantics.
- Add a PostgreSQL implementation using `pgx` with:
  - tenant-scoped reads and deletes;
  - JSONB source bindings and language-profile mappings;
  - PostgreSQL unique-violation mapping to `shared.ErrConflict`;
  - deterministic project ordering.
- Add migration `0045_projects.sql` with:
  - tenant foreign key;
  - unique `(tenant_id, key)` constraint;
[...]

PR #288, opened 2026-07-16
feat(codequality): ingest project sources for analysis

### LINKED ISSUE #176: feat(codequality): project source ingest — Git URL / archive / path → analysis (P2)
Part of #174.

## Context
There is no way to **get a project into Synapse to analyze**. Add project source ingest that reuses the existing sandboxed `acquire` (git-sandbox / image / dir) path.

## Tasks
- [ ] Source bindings: **Git URL** (branch/ref), **uploaded archive** (.zip/.tar.gz), and **local path** (dev). Reuse `internal/infrastructure/acquire` (git clone runs sandboxed; archive extraction is bounded + zip-slip-safe).
- [ ] `POST /api/v1/projects/{key}/analyses` → acquire the source → run the existing scan pipeline (quality + security) → persist an analysis snapshot (P3).
- [ ] Async job + status (reuse the scan-job runner; surface progress/errors).
- [ ] Web: "Add project" flow (choose source, kick off first analysis) + per-analysis status; drag-drop archive upload with size caps.
- [ ] CLI parity note: `synapse-cli scan` already covers the serverless path; document how it maps to a project analysis.

## Acceptance
From the UI a user can create a project from a Git URL or an uploaded archive and see an analysis run to completion, producing findings + metrics under that project.

pho-veteran:
## Summary
image

Closes #176 by making Code Quality projects ingestible and analyzable from the dashboard. Operators can create a project from a Git URL (with an optional ref), a bounded archive upload, or a development-only local path, then start and monitor a unified security and code-quality analysis without exposing the internal engagement used to enforce governance.

The implementation reuses the existing hardened acquisition and async SCA pipeline rather than adding a second scanner. Archive extraction remains bounded and zip-slip-safe through the existing acquirer; the new upload store only accepts `.zip`, `.tar.gz`, and `.tgz` archives and limits retained uploads to the configured workspace cap.

### Project source ingest

- Adds multipart project creation for archive uploads, alongside the existing JSON project creation endpoint.
- Supports Git URL sources with an optional branch/tag/ref, uploaded `.zip`/`.tar.gz`/`.tgz` archives, and local paths only outside production.
- Adds a file-backed `ProjectArchiveStore` that validates extensions, caps upload size, writes through a temporary file, publishes atomically, and removes retained archives when a project is deleted.
- Adds `SYNAPSE_PROJECT_UPLOAD_DIR` configuration and a persistent upload volume in the API image.
- Keeps `deploy/docker-compose.full.yml` intentionally out of this PR as requested.

### Governed async analysis

- Creates one hidden project-analysis engagement per project and persists its project association through migration `0046_project_analysis_context.sql`.
- Keeps hidden project-analysis engagements out of ordinary engagement reads and lists while allowing project-scoped lookup internally.
- Adds `POST /api/v1/projects/{key}/analyses`, `GET /api/v1/projects/{key}/analysis-status`, and `GET /api/v1/projects/{key}/analysis`.
- Runs the existing source acquisition, scope/window enforcement, async scan-job runner, findings persistence, scan-result cache, and code-quality report generation.
- Rejects duplicate starts while a project analysis is running and does not leak the hidden engagement ID in the project-analysis API response.

### Dashboard and documentation

- Adds the project source picker, Git ref input, archive drag/drop picker, client-side archive-size/type validation, initial-analysis start, status cards, polling progress, retry/run-again control, security metrics, and code-quality result rendering.
- Documents how `synapse-cli scan` maps to the project-analysis workflow and records the user-visible capability in `CHANGELOG.md`.
- Adds and updates handler, service, archive-store, SCA, RBAC, and frontend tests.

## Verification

- `make build` ✅
- `make test` ✅
- `make vet` ✅
- `corepack pnpm --dir web run typecheck` ✅
- `corepack pnpm --dir web build` ✅
- `git diff --check` ✅
- Exercised the running API with an isolated archive upload: accepted the AUP, uploaded a `.zip`, created the project, started `POST /projects/{key}/analyses`, observed the async job reach `succeeded`, and retrieved the persisted analysis from `GET /projects/{key}/analysis`.
- Probed an unsupported `.rar` upload and received `400 validation error: uploaded source must be .zip, .tar.gz, or .tgz`.
- Probed a nonexistent project analysis status endpoint and received `404 get project: not found`.

## Checklist

- [x] `cd web && pnpm build` passes (dashboard changed)
- [x] Preserves the safety invariants (argv exec, server-side scope enforcement, secrets out of logs, deterministic reports, append-only evidence) – see `CONTRIBUTING.md`
- [x] Documentation updated if needed (cli.md)

### DIFF
--- a/cmd/synapse-api/main.go
+++ b/cmd/synapse-api/main.go
@@ -330,7 +330,8 @@ func main() {
 
 	// Use cases.
 	engService := enguc.NewService(repo, clock, ids, auditLog)
-	projectService := projectuc.NewService(projectRepo, clock, ids, auditLog)
+	projectService := projectuc.NewService(projectRepo, repo, clock, ids, auditLog, !cfg.IsProduction())
+	projectService.SetArchiveStore(file.NewProjectArchiveStore(cfg.ProjectUploadDir, cfg.MaxWorkspaceBytes))
 	// Evidence artifact blob store: MinIO/S3 when configured, else in-memory (dev).
 	var blobStore ports.BlobStore
 	if cfg.BlobEndpoint != "" {
@@ -837,16 +838,19 @@ func main() {
 		os.Exit(1)
 	}
 	router := httpapi.NewRouter(log, auth, engService, scaService, aupService, findingsService, exportService, reportService, evidenceService, reconService, logBroker, transferService, auditService, vexService, usersService, credentialsService)
+	projectService.SetScanner(scaService)
 	router.SetProjects(projectService)
 	router.SetExploitation(exploitationService) // evidence-gated finding verify endpoint
 	// Read-only code-quality dashboard. Server-side analysis is PURE-GO and memory-safe only (pattern
 	// rules + duplication + Go-parser inventory); tree-sitter complexity is intentionally NOT wired here
 	// so the server never runs C parsers over untrusted source (that stays a local-CLI capability).
-	router.SetCodeQuality(codequality.New(
+	codeQualityService := codequality.New(
 		codeanalysis.New(),
 		codequality.WithDuplication(duplication.New(0)),
 		codequality.WithInventory(codeinventory.New()),
-	))
+	)
+	router.SetCodeQuality(codeQualityService)
+	scaService.SetCodeQuality(codeQualityService)
 	if ruleCat, rerr := rulecatalog.Default(); rerr != nil {
 		log.Error("rule catalog init failed", "err", rerr)
 		os.Exit(1)
--- a/deploy/Dockerfile
+++ b/deploy/Dockerfile
@@ -23,7 +23,8 @@ RUN CGO_ENABLED=0 go build -trimpath -o /out/synapse-api ./cmd/synapse-api
 # synapse-cli is bundled too so the image doubles as a cross-platform scanner
 # (`docker run … synapse-cli scan /scan`) – the SCA path runs identically on any host
 # that runs Linux containers (Windows/macOS via Docker Desktop, Linux natively).
-RUN CGO_ENABLED=0 go build -trimpath -o /out/synapse-cli ./cmd/synapse-cli
+RUN CGO_ENABLED=0 go build -trimpath -o /out/synapse-cli ./cmd/synapse-cli \
+    && mkdir -p /out/project-uploads
 
 # Pinned SCA tool binaries, shelled out at runtime.
 FROM anchore/syft:v1.45.1 AS syft
@@ -55,6 +56,7 @@ ENV SYNAPSE_SYFT_BIN=/usr/local/bin/syft \
     SYNAPSE_GRYPE_BIN=/usr/local/bin/grype \
     SYNAPSE_MVN_BIN=/usr/bin/mvn \
     SYNAPSE_GRADLE_BIN=/usr/local/bin/gradle \
+    SYNAPSE_PROJECT_UPLOAD_DIR=/home/synapse/project-uploads \
     JAVA_HOME=/opt/java-home
 EXPOSE 8080
 USER synapse:synapse
@@ -71,8 +73,11 @@ COPY --from=build /out/synapse-api /synapse-api
 COPY --from=build /out/synapse-cli /synapse-cli
 COPY --from=syft /syft /usr/local/bin/syft
 COPY --from=grype /grype /usr/local/bin/grype
+COPY --from=build --chown=nonroot:nonroot /out/project-uploads /project-uploads
 ENV SYNAPSE_SYFT_BIN=/usr/local/bin/syft
 ENV SYNAPSE_GRYPE_BIN=/usr/local/bin/grype
+ENV SYNAPSE_PROJECT_UPLOAD_DIR=/project-uploads
+VOLUME ["/project-uploads"]
 EXPOSE 8080
 USER nonroot:nonroot
 # Default to the API server; override for a one-off scan, e.g.:
--- a/docs/guide/cli.md
+++ b/docs/guide/cli.md
@@ -60,6 +60,10 @@ synapse-cli scan alpine:3.19 --image --offline
 The exit code is 0 when no finding meets the `--fail-on` threshold, and non-zero otherwise.
 Wire it straight into a pipeline step.
 
+### Project analysis parity
+
+`synapse-cli scan <local-path>` runs the same governed security scan path used by a Code Quality Project analysis and creates an ephemeral, scope-checked engagement around the target. The CLI does not attach the Project's combined code-quality report, create a Project, retain async job status, or persist the result; use `synapse-cli gate` for local code-quality gating. Git cloning and archive uploads are managed by the server-side Project flow, so clone or extract the source first for a serverless run.
+
 ## False-positive gate
 
 A scan of a real repository surfaces findings in test files and deliberately-insecure fixtures. Synapse
--- a/internal/adapter/httpapi/agent_handler_test.go
+++ b/internal/adapter/httpapi/agent_handler_test.go
@@ -62,6 +62,9 @@ func (f engRepoH) GetByID(context.Context, shared.ID) (*engagement.Engagement, e
 func (f engRepoH) GetByIDInTenant(context.Context, shared.ID, shared.ID) (*engagement.Engagement, error) {
 	return f.eng, nil
 }
+func (engRepoH) GetByProjectID(context.Context, shared.ID, shared.ID) (*engagement.Engagement, error) {
+	return nil, shared.ErrNotFound
+}
 func (engRepoH) List(context.Context, shared.ID) ([]*engagement.Engagement, error) { return nil, nil }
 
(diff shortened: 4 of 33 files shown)


nghiadaulau (changes requested):
Strong PR. The archive upload store, the reuse of the existing zip-slip-safe acquirer, the hidden-engagement governance model, and the scope pinning are all done carefully and hold up under review. One security fix is needed before merge, plus a couple of recommendations.

## Verified good (independent + reviewer pass)
- **Archive upload store**: server-side extension allowlist, size cap enforced two ways (`io.LimitReader(src, limit+1)` plus `written > limit`, and `http.MaxBytesReader` at the handler), write-through-temp then atomic rename, and the stored filename is `projectID + ext` from a server-generated ID, so the untrusted upload filename only contributes the extension and cannot traverse. Archive removed on create failure and on project delete.
- **No new extraction path**: archives are extracted only through the pre-existing zip-slip-safe acquirer.
- **Local source gated in the execution layer**, not just the UI (`create()` rejects `SourceLocal` when `!allowLocalSource`, wired from `!cfg.IsProduction()`), with a test.
- **Hidden-engagement governance**: still a full engagement carrying scope, so the execution gate is not bypassed. Scope is pinned to exactly the project source, tenant-scoped throughout, excluded from `List`/`GetByIDInTenant` in both memory and postgres, and reached internally only after tenant ownership is verified.
- **Scope + audit chokepoint intact**, RBAC correct (start = PermOperate, reads = PermView, machine role denied, cross-tenant 404), #287 credential-in-URL fix not regressed, migration `0046` additive and deploy-safe.
- Backend `build` (CGO on/off), `vet`, tests, and `web pnpm build` all pass here.

## Required before merge

**The hidden engagement ID leaks through `GET /api/v1/projects/{key}/analysis`.** `latestProjectAnalysis` (`internal/adapter/httpapi/project_handler.go`) writes the raw cached `ScanResult` JSON to the client. `ScanResult.Findings` is `[]finding.Finding`, and `finding.Finding.EngagementID` (`internal/domain/finding/finding.go:93`) is an **untagged** field that `buildFindings` sets to the hidden project-analysis engagement ID. So any analysis with at least one finding returns `"EngagementID":"<hidden-engagement-id>"`.

You already sanitize the job response (`projectAnalysisJobResponse`) and guard it with `TestProjectAnalysisJobHidesInternalEngagement`, but the result path is not covered and defeats the same stated invariant ("does not leak the hidden engagement ID"). The blast radius is limited today because `List`/`GetByIDInTenant` now filter `project_id IS NULL`, but it becomes dangerous the moment any future engagement-scoped endpoint authorizes without that predicate. Please return a sanitized project-analysis DTO (or strip/blank `EngagementID` from findings before writing the body) and add a test on the result path, not just the job path.

## Recommended (not strictly blocking, but please consider in this PR)
- **Promote `GetByProjectID` to the `ports.EngagementRepository` port.** Right now the service defines a private `analysisContextRepository` and does `contexts, _ := engagements.(analysisContextRepository)`, discarding the `ok`. There is no compile-time guarantee a repo implements it, so a future implementation that omits it would compile and silently degrade `Create`/`StartAnalysis` to a runtime error. This is the one place the PR steps outside the "all I/O through a declared port" convention; declaring it on the port and dropping the blind assertion restores that guarantee.
- **Add a focus ring to the archive drop-zone button** (`web/src/pages/CodeQualityProjects.tsx`). The raw `<button>` omits the `focus-visible:ring-2 focus-visible:ring-brand/50` every other control here uses, so keyboard users get no focus indication on the primary picker.

## Optional (fine as follow-ups)
- Duplicate-start guard is TOCTOU-racy (two concurrent `POST /analyses` can both enqueue). Not dangerous since both pin the same source, but an engagement-keyed run-lock would tidy it.
- The hidden analysis engagement has an open authorization window (`AuthorizedFrom/To == nil`), so the window control is a no-op here and scope pinning is the only active control. Defensible for static analysis of operator-supplied source, but worth a one-line comment saying so.
- Set `ProjectID` through a domain mutator/constructor variant rather than poking `analysis.ProjectID = p.ID` after `engagement.New`, so the invariant lives in the domain.
- `service.go` returns the `LatestJob` error unwrapped in one spot; wrap with `%w` for consistency.
- Frontend: a failed status fetch is `.catch(() => null)` and renders as "Not analyzed" (indistinguishable from genuinely un-analyzed); and `mapCodeQualityReport` defaults missing ratings to `'A'` (likely pre-existing, shared with the engagement view) which would show a fabricated top grade if a report is present but ratings are absent. A neutral/unknown default would be safer.

Fix the engagement-ID leak and I'll re-review and merge. Nice work on the upload and governance safety.

pho-veteran:
Thanks for the thorough review. I’ve addressed the blocking security issue and both recommendations, and also fixed one of the optional frontend issues after verifying the updated flow.

## Required fix addressed

- **Stopped the hidden engagement ID from leaking through `GET /api/v1/projects/{key}/analysis`.**
  - The cached analysis response is now sanitized before it is written to the client.
  - `EngagementID` is removed from findings in both locations where `finding.Finding` appears in `ScanResult`:
    - top-level `findings`
    - `code_quality.findings`
  - The sanitizer preserves unrelated and forward-compatible fields instead of rebuilding and potentially narrowing the cached result.
  - Malformed cached payloads fail closed with an internal error and are not echoed back to the client.

- **Added result-path regression coverage.**
  - The new handler test exercises `latestProjectAnalysis` directly with findings containing hidden engagement IDs.
  - It verifies that the IDs and their possible JSON key variants are absent from the response.
  - It also verifies that unrelated root, report, and finding fields are preserved.
  - A malformed-cache test confirms the original cached payload is not leaked through the error response.

## Recommendations addressed

- **Promoted `GetByProjectID` to `ports.EngagementRepository`.**
  - Removed the private `analysisContextRepository`.
  - Removed the unchecked type assertion and silent runtime degradation path.
  - `projectuc.Service` now calls the declared repository port directly.
  - Updated repository test doubles so missing implementations are caught at compile time.

- **Added a visible keyboard focus ring to the archive drop-zone button.**
  - The picker now uses the same `focus-visible` treatment as the other interactive controls.
  - Added a component test to guard the focus styling.

## Additional issue found and fixed

While checking the frontend recommendation around status-fetch failures, I found that the first fix could make one failed status request reject the whole `Promise.all` and hide an otherwise successfully loaded project list.

That flow is now handled per project:

- Projects are rendered as soon as `listProjects` succeeds.
- Each project status request is isolated, so one failure does not hide the other projects.
- A failed status is shown as **Unknown**, not **Not analyzed**.
- The actual error and a **Retry** action remain visible.
- Added a regression test proving the project card remains visible when its status request fails.

## Verification

- `go test ./internal/adapter/httpapi ./internal/usecase/projectuc`
- `corepack pnpm --dir web exec vitest run src/pages/CodeQualityProjects.test.tsx --reporter=verbose` — 9/9 passed
- `corepack pnpm --dir web exec tsc --noEmit`
- `git diff --check`

The remaining optional items — the duplicate-start TOCTOU guard, explicit authorization-window comment, domain-level `ProjectID` setter/constructor, `LatestJob` error wrapping, and neutral fallback for missing code-quality ratings — are unchanged and can remain follow-ups as suggested.

Ready for re-review. Thanks again for catching the result-path leak.

nghiadaulau (approved):
Approving. The blocking leak is fixed and both recommendations are in.

**Verified on `6782bd5`:**
- **Required (hidden engagement ID leak)** `latestProjectAnalysis` now runs `redactProjectAnalysisEngagementIDs`, which decodes to `map[string]json.RawMessage` (so unrelated/forward-compatible fields are preserved), strips `EngagementID` (and the `engagement_id`/`engagementId` key variants) from both the top-level `findings` and `code_quality.findings`, and fails closed on a malformed cache without echoing the raw payload. Covered by `TestLatestProjectAnalysisHidesInternalEngagement` and `TestLatestProjectAnalysisRejectsMalformedCache`, with the job-path test still green.
- **Recommended: `GetByProjectID` promoted to `ports.EngagementRepository`** the private interface and unchecked type assertion are gone, and the test doubles across the codebase now implement it, so a missing implementation is a compile error.
- **Recommended: focus ring** added to the archive drop-zone button, matching the other controls, with a component test.
- **Bonus** the frontend status fetch is now per-project isolated (one failed status no longer hides the whole list) and shows `Unknown` rather than `Not analyzed` on error, with a regression test.

Full gate green here: `go build` (CGO on and `CGO_ENABLED=0`), `go vet ./...`, the httpapi/projectuc/sca/engagement/execution/orchestrator/report/safety/vex tests, and `web pnpm build`.

Everything else was already clean on the first pass (safe archive store, zip-slip-safe acquirer reuse, execution-layer local-source gating, scope-pinned hidden engagement, tenant scoping, RBAC, additive migration 0046). The remaining optional items (TOCTOU duplicate-start, auth-window comment, domain `ProjectID` setter, `LatestJob` error wrapping, neutral rating fallback) are fine as follow-ups. Merging.

PR #290, opened 2026-07-17
feat(codequality): harden project analysis workflows

### LINKED ISSUE #177: feat(codequality): analysis snapshots + Activity timeline + Overall-vs-New-Code (P3)
Part of #174.

## Context
No analysis **history** is persisted, so there is no Activity timeline and no Overall-vs-New-Code comparison over time.

## Tasks
- [ ] `domain` + store: `Analysis` snapshot = {projectKey, ref/commit, timestamp, measures map, gate result, issue counts by type/severity/status, coverage, duplications, ratings}. Append-only.
- [ ] Define the **New Code period** (since previous analysis / since a fixed ref/date) and compute new-code deltas from consecutive snapshots (reuse the existing new-code gating logic).
- [ ] HTTP: `GET /api/v1/projects/{key}/analyses` (history) + `GET .../analyses/{id}` (one snapshot).
- [ ] Web: **Activity** page — timeline of analyses + metric trend charts (ratings, issues, coverage, duplications) using the existing chart primitives.

## Acceptance
Each analysis is stored as a snapshot; the Activity page shows the trend of ratings/coverage/issues over time, and New-Code metrics are derived from snapshot deltas.

pho-veteran:
## Summary

Hardens the Code Quality Project workflow across immutable analysis history,
scan admission, portfolio reads, managed quality-gate integrity, New Code
semantics, and frontend state.

Close #177 

## Closed issues

- Store each completed Project analysis snapshot with its matching scan result,
  so the latest-result endpoint cannot combine data from different runs.
- Enforce one `running` scan per engagement in the persistence layer:
  - atomic `CreateRunning` admission;
  - PostgreSQL partial unique index;
  - failed queue admission marks the claimed job failed when possible;
  - stale-scan recovery remains available for stranded jobs.
- Replace browser-side Project status fan-out with compact, server-enriched
  portfolio summaries backed by bulk analysis/context/job reads.
- Keep portfolio responses compact by omitting full scan results, detailed
  coverage/duplication payloads, gate conory.
- Compare coverage trends only against the immediate predecessor rather than
  skipping snapshots without coverage.
- Clear stale finding selection after filters remove the selected finding.
- Reset local gate-editor form state when
- Make managed custom-gate create/update/delete and Project gate assignment
  atomic with their audit append in PostgreSQL.
- Serialize custom-gate deletion, assignmh the
  same tenant/gate advisory-lock protocol:
  - creation first makes deletion conflict;
  - deletion first makes custom-gate Project creation fail;
  - no Project can retain a reference to  gate.
- Mirror the managed-gate request-visible behavior in memory mode.
- Stop deriving Project New Code LOC and maintainability from repository LOC
  deltas. New Code maintainability is expnged or
  added lines are measured from a source diff.
- Preserve New Code security/reliability grades and issue-based gate metrics.

## Migrations

- `0049_project_analysis_result.sql`
- `0050_scan_jobs_one_running.sql`
- `0051_project_gate_assignments.sql`

## Deliberate follow-ups

- The Project portfolio remains unpaginats
  server-side filtering when full tenant portfolios become materially expensive.
- Persist source-diff changed/addition LOC before restoring Project New Code LOC
  and maintainability.
- The existing Project creation saga still creates Project, engagement, and its
  audit record in separate steps; this PRreference
  invariant.

## Validation

```text
go test ./...
make build vet test typecheck
corepack pnpm --dir web run typecheck
corepack pnpm --dir web test -- --run
corepack pnpm --dir web run build
git diff --check


nghiadaulau (changes requested):
Big, high-quality hardening PR, and the hardest parts hold up. An independent security pass verified the atomic scan admission (partial unique index + `CreateRunning`, which actually closes the #288 TOCTOU), the advisory-lock quality-gate integrity (lock -> mutate -> audit in one tx, consistent ordering, no dangling gate reference), immutable snapshots stored with their matching result, tenant scoping, RBAC, append-only audit, the argv-only `git rev-parse` in the acquirer, and the #288 engagement-ID redaction all as correct. Migrations 0047-0051 are additive and clean. Nice work, and thanks for switching the other rating mappers to `?`.

There are two hard blockers plus two more required fixes before merge.

## Required before merge

**1. Dependency-rule violation: usecase imports infrastructure.** `internal/usecase/projectuc/service.go:21` imports `internal/infrastructure/tools/qualityprofile` and calls `qualityprofile.LoadGateBytes(result.GateFile)` at `:404`. That package pulls in `os` + `yaml.v3`; the rule is `usecase -> domain + usecase/ports` only, and this is the single usecase->infrastructure edge in the whole tree, new in this PR. Please parse the repo gate file at the infra edge (the SCA pipeline already reads the workspace) and put a domain `qualitygate.Gate` on `ScanResult` instead of the raw `GateFile []byte`, so the usecase consumes a domain type. (Or move the pure parse into `domain/qualitygate`, or behind a `ports` interface.)

**2. Fabricated best-grade `A` in the portfolio mapper.** `web/src/lib/api.ts:562-563` `mapProjectSummaryAnalysis` still defaults missing ratings to `'A'` (in `grade()` and in the hardcoded `newCode.rating`), while `mapCodeQualityReport` (`:550-552`) and `mapProjectAnalysis` (`:569`) were correctly moved to `'?'`. This one renders a green **A** in the portfolio `ProjectCard` (`MiniGrade`) whenever a summary omits a rating, which is the fabricated-data anti-pattern we care about. Default all three to `'?'` and make `newCode.rating` `'?'`/`null` here too.

**3. `aria-pressed` selection highlight is wrong for empty-id findings.** `web/src/components/codequality/FindingExplorer.tsx:21` uses `aria-pressed={selected?.id === finding.id}`. Scan-result findings are frequently unpersisted (empty/undefined id; the list key already falls back to `finding.id || finding.dedupKey`), so `undefined === undefined` marks every empty-id row as pressed and highlights them all. Compare by the same composite key the stale-selection effect uses (id + dedupKey) or by identity/index.

**4. Virtualize the FindingExplorer list.** `FindingExplorer.tsx:21` renders every match as a `<button>` in a fixed-height scroll container; a real scan easily exceeds the 50-row threshold. The sibling `CodeQualityReportView.tsx` already uses `VirtualTable` for findings; reuse it (or paginate), per the >50-rows convention.

## Recommended (not strictly blocking)
- `internal/usecase/projectuc/service.go` `RecordProjectAnalysis` now runs on every scan completion and returns an error if `engagements.GetByID` fails, which would flip an otherwise-successful non-project engagement scan to `failed`. Guard so a non-project scan is unaffected.
- `internal/usecase/sca/service.go:1099`: the `if s.jobs != nil` guard around job creation was dropped; keep a defensive nil-check (no-panic-in-library convention).
- `LatestWithResult` memory vs postgres drift: postgres filters `result IS NOT NULL` in-query and returns the latest result-bearing row; memory takes the absolute-latest row then returns `ErrNotFound` if it has no result. Make memory skip result-less rows to match (harmless today because `SaveWithResult` is always non-nil, but the port exposes a nil-result `Save`).
- Promote the batch-read fast paths (`ProjectContexts`, `LatestForEngagements`) to the `ports.EngagementRepository`/`ports.ScanJobStore` interfaces instead of runtime `interface{...}` type-assertions, so the contract is compile-checked.

## Optional / notes
- `tabular-nums` missing on a few numeric spots (`ProjectActivityView.tsx:49,56`, `QualityGates.tsx:23`, `qualityPresentation.tsx:34`).
- Portfolio card grid is unpaginated (you already noted this as a follow-up, fine).
- memory `quality_gate_mutator` audit-then-write is not atomic like the postgres tx version; negligible (dev-only backend, append-only audit, pre-checked), noting for parity.
- Scope note (advisory, not blocking): this bundles the quality-gate CRUD/mutator stack (P9/P10) with the P3 snapshot work. Cohesive at the seam, so fine here, but landing gate management on its own would have been easier to review.
- Heads-up so you don't chase it: the `web` suite shows one failing test, `RuleDetail.test.tsx "does not overwrite new rule with stale response"`. I confirmed it also fails on `main` (a local undici/AbortSignal environment quirk, unrelated to this PR), so it is not a blocker here.

Fix the four required items and I'll re-review and merge. The security and concurrency work in this PR is genuinely solid.

pho-veteran:
Thanks for the thorough review — especially for validating the scan-admission, advisory-lock, snapshot, tenant/RBAC, audit, acquirer, and redaction work.

Resolved all requested changes:

1. **Usecase dependency direction**
   - Removed the `projectuc -> infrastructure/tools/qualityprofile` import.
   - Repository gate loading now happens after source acquisition in SCA.
   - Parsing is injected through `ports.GateDecoder`, and `ScanResult` carries the decoded domain `qualitygate.Gate`.

2. **Portfolio ratings**
   - Missing portfolio and New Code ratings now map to `?`, never fabricated `A`.
   - Added regression coverage for absent ratings.

3. **Empty-ID finding selection**
   - `FindingExplorer` now uses a shared composite `(id, dedupKey)` identity for row keys, stale-selection reconciliation, and `aria-pressed`.
   - Added coverage for multiple unpersisted findings without IDs.

4. **FindingExplorer scaling**
   - Added pagination: 50 findings render initially, with an explicit **Load more findings** action.
   - This preserves the master/detail explorer while satisfying the >50-row pagination/virtualization requirement.

Also addressed the recommended items:

- Project snapshot recording runs only when `ScanOptions.ProjectAnalysis` is enabled, so ordinary engagement scans are unaffected.
- Restored the defensive `s.jobs != nil` guard around `CreateRunning`.
- Memory `LatestWithResult` now skips result-less snapshots, matching PostgreSQL’s result-bearing lookup.
- Promoted `ProjectContexts` and `LatestForEngagements` to typed repository ports, removing runtime interface assertions.
- Applied the requested `tabular-nums` polish.

Verification passed:

- Focused Go tests across project analysis, SCA, persistence, HTTP API, and affected use cases.
- `npm --prefix web run typecheck`
- Focused frontend tests: `api.projects.test.ts` and `FindingExplorer.test.tsx` — **12/12 passed**.
- `git diff --check`

nghiadaulau (approved):
Approving. All four required items are fixed and verified on `c29f87e`.

1. **Dependency-rule violation resolved** the `internal/infrastructure/tools/qualityprofile` import is gone from `usecase`; the gate is now parsed at the infra edge and the usecase consumes a domain type. Confirmed there is no `usecase -> internal/infrastructure` import anywhere in the tree, and `go vet ./...` is clean.
2. **Fabricated best-grade `A` removed** no `?? 'A'` remains in `web/src/lib/api.ts`; `mapProjectSummaryAnalysis` now defaults ratings to `'?'`/`null`, matching the other mappers.
3. **`aria-pressed` selection bug fixed** the row now compares with `findingKey()` (the composite id+dedupKey), so empty-id findings no longer all highlight.
4. **Findings list bounded** the explorer renders `slice(0, shown)` with a "Load more" control, so it no longer renders unbounded DOM past the 50-row threshold.

Recommended items also addressed: batch-read fast paths promoted to the `ports` interfaces (compile-checked now), the `s.jobs` nil-guard restored, the memory/postgres `LatestWithResult` drift aligned, and the `RecordProjectAnalysis` edge handled. The `tabular-nums` nits were picked up too.

Full gate green here: `go build` (CGO on and `CGO_ENABLED=0`), `go vet ./...`, the backend affected + changed-double package tests, `CGO_ENABLED=1` astwalk, and `web pnpm build`. Web tests: 125 passed; the single failure is `RuleDetail.test.tsx "does not overwrite new rule with stale response"`, which I verified also fails on `main` (a local undici/AbortSignal environment quirk), so it is unrelated to this PR.

The security and concurrency core was already verified clean on the first pass (atomic scan admission, advisory-lock gate integrity, immutable snapshots, tenant/RBAC/audit, #288 leak non-regression, additive migrations 0047-0051). Merging. Solid work.

Issue #181, opened 2026-07-12
feat(codequality): Issues explorer + accept/false-positive/won't-fix lifecycle (P7)


nghiadaulau:
Part of #174.

## Context
Synapse lists findings, but there is no per-project **Issues explorer** with a triage lifecycle (the standard *accepted* / *false-positive* / *won't-fix* buckets).

## Tasks
- [ ] Issue lifecycle on findings: **Open → Accepted / False-positive / Won't-fix / Reopen**, attributable + append-only (reuse the Judgment/evidence + existing finding-status model; a resolved issue is gate-exempt but retained + sealed, never deleted — mirrors the FP-triage retain-and-mark contract).
- [ ] Faceted explorer: filter by **type** (bug/vulnerability/code-smell/hotspot), **severity**, **rule**, **language**, **file/dir**, **status**, **new-code-only**.
- [ ] `GET /api/v1/projects/{key}/issues` (facets + pagination) + `POST .../issues/{id}/transition`.
- [ ] Web: Issues tab (faceted sidebar + VirtualTable + issue detail with rule link (P8), why-it-flagged, and the transition control).

## Acceptance
A user filters a project's issues by every facet and moves an issue to Accepted/FP/Won't-fix with an attributable, gate-affecting (exempting) transition.

VietSory:
I'd like to work on this.

Issue #179, opened 2026-07-12
feat(codequality): Security Hotspots + Reviewed % review workflow (P5)


nghiadaulau:
Part of #174.

## Context
Mature code-quality platforms separate **Security Hotspots** (security-sensitive code a human reviews) from confirmed vulnerabilities, and track a **Reviewed %** rating. Synapse has security findings but no hotspot concept or review workflow.

## Tasks
- [ ] Introduce a **Hotspot** projection: security-sensitive findings/patterns (e.g. crypto usage, request-controlled I/O, auth-sensitive code) flagged for review rather than asserted as exploitable. Seed from existing SAST/secret/misconfig signals + dedicated hotspot rules (authored in the language packs).
- [ ] Review workflow: **To review → Acknowledged / Fixed / Safe**, attributable + append-only (reuse the Judgment/evidence model, mirroring issue lifecycle in P7). A "Safe" review is a governed decision, not a silent suppression.
- [ ] `security_hotspots_reviewed` metric (% reviewed) + a Security Review rating; wire into the overview (P4) and gate metrics (P10).
- [ ] Web: Security Hotspots tab (list + review side-panel with rationale).

## Acceptance
Security hotspots are listed, reviewable through a governed workflow, and the Reviewed % appears on the overview and is gateable.

VietSory:
I'd like to work on this.

nghiadaulau:
Reopening: PR #294 landed the **foundation only** (the read-only hotspot projection + GET list/detail API, with hotspots partitioned out of vulnerability/gate metrics). The remaining P5 scope is not done yet and is tracked here: the governed review workflow (status transitions), Security Hotspots Reviewed %, Overview integration, Quality Gate hotspot metrics, and the Security Hotspots web experience. These land in the #179 follow-up PR (PR-B).

Issue #178, opened 2026-07-12
feat(codequality): project overview — quality gate + six rating cards (P4)


nghiadaulau:
Part of #174.

## Context
Assemble the **project overview** that mature code-quality dashboards lead with: the quality-gate verdict + the six headline cards, with **Overall Code** and **New Code** lenses.

## Tasks
- [ ] Overview endpoint aggregating the latest analysis: gate status (Passed/Failed + failing conditions), and six cards — **Security** (A–E), **Reliability** (A–E), **Maintainability** (A–E), **Security Hotspots** (Reviewed %, from P5), **Coverage** (%), **Duplications** (%).
- [ ] Overall vs New Code toggle; show **New issues** + **Accepted issues** counts (from P7).
- [ ] Web: overview page with the gate banner + the six cards + the Overall/New toggle; each card links to its detail (Issues/Hotspots/Measures).
- [ ] Ratings + non-color cues (icons/labels) per the design system; loading/empty/error states.

## Acceptance
Opening a project shows a gate verdict and the six ratings for both Overall and New code, each drilling into its detail view.

VietSory:
I'd like to work on this.

nghiadaulau:
Delivered by #291 (the read-model + API contract) and #292 (the Overview UI): the project overview now shows the Quality Gate passed/failed banner with ordered failing-condition evidence, the six headline rating cards (Security, Reliability, Maintainability, Security Hotspots Reviewed, Coverage, Duplications) for both the Overall Code and New Code lenses, and honest unavailable/not-supplied/not-applicable states (no fabricated grades or percentages). Cards whose drill-down destinations depend on later phases (hotspots #179, issues #181, changed-line metrics) render as explicitly unavailable rather than linking to placeholders. P4 complete.

PR #291, opened 2026-07-17
Feat/178 a project overview contract

VietSory:
## Summary

Adds the Project Overview read model and API contract required by #178.

This PR introduces a tenant-scoped endpoint that projects the latest immutable Project analysis into a compact, frontend-oriented response without loading or exposing the full scan-result payload.

## What changed

* Added `GET /api/v1/projects/{key}/overview`.
* Added an application-layer Project Overview read model.
* Added explicit `analyzed` and `not_analyzed` response states.
* Added stored Quality Gate verdict and ordered failed-condition evidence.
* Added Overall Code and New Code contracts for:

  * Security;
  * Reliability;
  * Maintainability;
  * Security Hotspots Reviewed;
  * Coverage;
  * Duplications.
* Added explicit availability and unavailable-reason semantics.
* Added unambiguous issue summary fields:

  * `new_code_total`;
  * `accepted_overall_total`.
* Added OpenAPI `oneOf` schemas for analyzed and not-analyzed responses.
* Added closed enums for gate metrics, operators, sources and statuses.
* Added a strict TypeScript response parser and API client method.
* Added backend, HTTP and frontend contract tests.
* Added frontend Vitest execution to CI.

## Contract behavior

An existing Project without a completed analysis returns:

```text
HTTP 200
state = not_analyzed
```

A missing or cross-tenant Project uses the existing not-found behavior.

The endpoint reads from the latest immutable Project analysis snapshot. It does not:

* load or decode the full scan result;
* re-evaluate the current Quality Gate;
* recalculate ratings from findings;
* load the current Quality Profile;
* reconstruct New Code from mutable state.

## Data integrity

The Overview mapper validates persisted evidence before exposing it:

* gate metrics, operators and sources must belong to the supported contract;
* gate verdict must agree with all condition results;
* failed conditions preserve snapshot order;
* numeric evidence must be finite;
* Project and analysis identities must be non-empty;
* analysis timestamps must be valid;
* New Code periods must be either a first analysis or a valid baseline comparison.

Invalid persisted evidence fails closed with a controlled server error.

## Availability semantics

Missing information is represented explicitly and is never converted into a successful-looking value.

In particular:

* missing coverage is not represented as `0%` or `100%`;
* unknown ratings are not represented as grade `A`;
* New Code coverage and duplication remain unavailable until changed-line producers exist;
* Security Hotspots Reviewed remains unavailable until #179;
* accepted-issue counts remain unavailable until #181.

## Security and tenancy

The endpoint does not expose:

* tenant IDs;
* internal Project IDs;
* hidden Engagement identifiers;
* source paths or repository credentials;
* raw findings;
* full scan results;
* `InternalIssues`.

Project lookup remains tenant-scoped and protected by Project view permission.

## Scope

This PR establishes the read model and frontend data contract only.

It does not add:

* the React Project Overview page;
* the Quality Gate banner UI;
* the six-card grid;
* the Overall/New Code toggle;
* Issues, Hotspots or Measures drill-down pages.

Those remain follow-up work for the next #178 implementation slice.

## Local validation

Passed:

* `git diff --check`
* targeted Project Overview use-case tests
* targeted Project Overview HTTP tests
* `go vet ./...`
* all Go packages except the environment-dependent sandbox package
* `pnpm install --frozen-lockfile`
* `pnpm test` — 15 files, 132 tests
* `pnpm run typecheck`
* `pnpm build`
* OpenAPI YAML parsing and local `$ref` resolution

Environment limitations:

* the full Go suite fails only in `internal/infrastructure/sandbox` under WSL with process/filesystem sandbox errors;
* the locally installed `golangci-lint` cannot parse the repository's Go version directive and must be validated using the CI-pinned version.

Final merge validation is delegated to the upstream pull-request CI running on its supported Linux and Windows environments.

Refs #178

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -94,6 +94,9 @@ jobs:
       - name: Install
         run: pnpm install --frozen-lockfile
 
+      - name: Test
+        run: pnpm test
+
       - name: Typecheck
         run: pnpm run typecheck
 
--- a/api/openapi.yaml
+++ b/api/openapi.yaml
@@ -30,6 +30,8 @@ tags:
   - name: findings
   - name: sca
   - name: export
+  - name: projects
+    description: Long-lived code-quality Projects and immutable analysis read models
   - name: rules
     description: Built-in code-quality and security rule catalog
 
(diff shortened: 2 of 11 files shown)


nghiadaulau (changes requested):
Excellent first contribution. I reviewed this strictly across security, architecture, and the frontend contract, and independently verified the claims. The read model is one of the cleanest additions to this epic: the data-honesty model is exactly right and the leak/tenancy discipline is solid.

## Verified strong (not just taking the description's word for it)
- **Data honesty (the priority)**: ratings map `''`/`'?'` to `unavailable/rating_not_available` (never defaulted to A) and hard-error on an unknown grade; coverage returns `not_supplied` when absent and `not_applicable` at zero executable lines (never a fake `0%`/`100%`); New-Code coverage/duplication, hotspots-reviewed, and accepted totals stay explicitly unavailable with typed reasons. The TS parser mirrors this and is materially stricter than the existing `mapCodeQualityReport`/`mapProjectAnalysis`.
- **No sensitive leak**: traced every field crossing the boundary. No tenant ID, internal project ID, hidden engagement ID, source path/credentials, raw findings, full scan result, or `InternalIssues`. The endpoint reads metadata-only (`LatestForProjects`), so the `ScanResult` blob behind the #287/#288 leak is never even loaded. The handler leak test asserts all of these are absent.
- **Tenant isolation + RBAC**: tenant-scoped `Get` + `LatestForProjects`, cross-tenant/missing → identical 404; route behind `authz(PermView)`, machine role 403 before the service runs.
- **Fail-closed**: invalid persisted evidence returns non-sentinel errors → generic 500, DSN not leaked (tested). The gate-consistency re-check and finite-number guards genuinely fail closed.
- **Contract alignment**: the gate-metric enum is identical across Go (`qualitygate.knownMetrics`), OpenAPI, and the TS `GATE_METRICS` (12/12/12), so the strict parser cannot spuriously reject a valid response.
- Dependency rule clean (usecase imports domain + ports only), `go build`/`vet`/tests green, `web build` green.

## Required before merge (both small)

1. **Reuse the domain invariant instead of reimplementing it.** `overviewGate` in `internal/usecase/projectuc/overview.go` re-derives `allPassed` and hand-filters the failed conditions, but `qualitygate.Result.Failures()` already returns exactly the non-passing conditions and `Evaluate` already guarantees `Passed` agrees with them. Build `FailedConditions` from `gate.Failures()`. Keeping a consistency re-check at the read boundary is good defense against corrupt persisted rows, but put it on the domain type as `qualitygate.Result.Validate()` and call that, rather than owning a second copy of the rule in a read-mapper.

2. **Lock the `not_supplied` / `not_applicable` contract with tests.** The parser (and backend) strictly require a non-null `unavailable_reason` for those two availability states, and today's fixtures only exercise `available`/`unavailable`. Add a positive case for each of `not_supplied` (coverage absent) and `not_applicable` (zero executable lines / no duplication) in both `overview_test.go` and `api.projects.test.ts`, plus a rejection case for the reason-missing branch. That is exactly the path that could silently regress.

## Recommended (not blocking)
- Doc-comment the new exported surface (`Service.Overview` and the `Overview*`/`Metric*`/`RatingMetric`/`PercentageMetric`/`CountMetric` types + reason/grade consts). The repo's revive config disables the `exported` rule so lint won't flag it, but this is a whole new public read-model API.
- The `[0,100]`/non-negative/ordering guards in `coverageMetric`/`duplicationMetric` are invariants of `measure.CoverageReport`/`DuplicationReport`; they belong on those domain types (constructor or `Validate()`) rather than re-checked in the read-mapper.
- Single-source the OpenAPI `ProjectOverviewGateMetric` enum against `qualitygate.ValidMetric` (they agree today, but it is a second hand-maintained list that can drift).
- `Overview` validates the project, then `analyzedOverview` validates it again; pass the validated value through or drop the second call.
- UI follow-up note: unavailable ratings are `grade: null` here (better than the `'?'` sentinel elsewhere); the cards PR must render `null` as unknown, not blank.
- LOW: `SourceRef`/`SourceCommit` are copied verbatim without a length/charset bound (server-generated, already surfaced by existing endpoints, so no new exposure).

## Note (not a fault of this PR)
Adding `pnpm test` to `.github/workflows/ci.yml` is a good improvement, but that workflow does not currently run on PRs (only GitGuardian + Dependabot execute), so it is inert today. When Actions is enabled it will surface a pre-existing failing test, `RuleDetail.test.tsx "does not overwrite new rule with stale response"`, which I confirmed also fails on `main` (a local undici/AbortSignal environment quirk), not anything you changed. Worth quarantining that flaky test separately before the CI step goes live.

Address the two required items and I'll merge. Genuinely strong, careful work — thank you.

nghiadaulau (approved):
Approving and merging. On reflection I'm reclassifying my two earlier "required" items as non-blocking follow-ups: neither is a defect. The re-validation in `overviewGate` is correct and actually a reasonable read-boundary defense against corrupt persisted rows, and the `not_supplied`/`not_applicable` code path is correct and matches the backend, it just lacks a dedicated fixture. A demanding review should not hold a clean PR on polish.

To be clear about the bar this cleared: security review is SHIP (I traced every field crossing the boundary, nothing leaks: no tenant/project/hidden-engagement ID, no source paths/creds, no raw findings, no full scan result, no `InternalIssues`, and the endpoint reads metadata only so the `ScanResult` blob is never loaded), tenant isolation and RBAC are correct, the fail-closed integrity checks genuinely fail closed, the dependency rule is clean, and the gate-metric enum is aligned across Go / OpenAPI / TypeScript. The data-honesty model is the best in this epic so far. `go build`/`vet`/tests and `web build` are green.

Please keep these as follow-ups (a tracking issue is fine, not required for this PR):
1. Build `FailedConditions` from `qualitygate.Result.Failures()` and, if you keep the read-boundary consistency check, move it onto the domain as `qualitygate.Result.Validate()` rather than reimplementing it in the mapper.
2. Add fixtures for the `not_supplied` and `not_applicable` availability states (plus the reason-missing rejection) in `overview_test.go` and `api.projects.test.ts`.
3. Doc-comment the new exported `Overview*` surface (revive's `exported` rule is off in this repo, so it's a convention nicety, not a lint gate).
4. Consider moving the `measure` range/ordering guards onto the domain types, and single-sourcing the OpenAPI metric enum against `qualitygate.ValidMetric`.

Note: the `pnpm test` addition to `ci.yml` is a good improvement, but that workflow does not currently run on PRs (only GitGuardian + Dependabot do), so it is inert today. When Actions is enabled it will surface a pre-existing flaky test (`RuleDetail.test.tsx`, which also fails on `main` due to a local undici/AbortSignal quirk) that should be quarantined separately.

Excellent first contribution. Merging.

Issue #180, opened 2026-07-12
feat(codequality): Measures explorer — metric tree + file drill-down (P6)


nghiadaulau:
Part of #174.

## Context
Expose the full **Measures** tree (the metric-explorer view mature analyzers provide), beyond the current three-rating summary.

## Tasks
- [ ] Aggregate the metrics the engine already computes: ncloc, files, functions, **cyclomatic** + **cognitive** complexity, **comment density** (add if missing), **duplicated lines/blocks %**, **coverage** (+ new-code coverage), issues by **type** (bug/vuln/smell/hotspot) and **severity**, tech-debt (remediation effort), ratings.
- [ ] Directory/file **drill-down** (roll metrics up the tree; reuse the AST inventory).
- [ ] `GET /api/v1/projects/{key}/measures` (+ per-path).
- [ ] Web: Measures tab (metric domains + a file tree with per-file metrics; VirtualTable for >50 rows).
- [ ] Add any missing first-class measures (comment density, complexity/cognitive as reportable measures, new-code coverage) so P4 + P10 can reference them.

## Acceptance
A user can browse every computed measure for a project and drill from project → directory → file.

VietSory:
I'd like to work on this.

PR #294, opened 2026-07-19
Feat/179 a hotspot projection api

VietSory:
## Summary

Introduces the read-only Project Security Hotspot foundation for #179.

Findings backed by catalog rules of type `security_hotspot` are now projected into stable, tenant-scoped and Project-scoped hotspot records instead of being treated as confirmed vulnerabilities in Project analysis metrics.

This PR intentionally delivers the projection and read API only. The governed review workflow, Reviewed %, Overview integration, Quality Gate metrics, and Security Hotspots web experience will be completed in the follow-up PR that finishes #179.

## What changed

* Added a dedicated Project Security Hotspot domain model and closed status vocabulary.
* Added deterministic hotspot classification using `Finding.RuleKey → catalog Rule.Type`.
* Added stable hotspot identities across repeated Project analyses.
* Added memory and PostgreSQL persistence implementations.
* Added migration `0052_project_hotspots.sql`.
* Added atomic persistence for the immutable Project analysis and its hotspot projections.
* Added rollback behavior when hotspot projection persistence fails.
* Preserved review state, version, first-seen metadata, and latest descriptive fields across rescans.
* Added deterministic handling for analyses arriving out of chronological order.
* Added read-only Project hotspot endpoints:

  * `GET /api/v1/projects/{key}/hotspots`
  * `GET /api/v1/projects/{key}/hotspots/{id}`
* Added status, rule, severity, and text filters.
* Added deterministic cursor pagination and filtered facets.
* Added tenant isolation, Project isolation, RBAC, response-leakage, persistence, and OpenAPI tests.
* Updated API startup wiring to require the shared rule catalog and hotspot-capable Project analysis store.

## Classification contract

The rule catalog type is the sole authority for hotspot classification:

```text
Finding.RuleKey
→ catalog lookup
→ Rule.Type == security_hotspot
→ Security Hotspot projection
```

Classification does not infer hotspot status from:

* finding kind;
* severity;
* CWE;
* title;
* tags;
* security quality alone.

Findings with missing or unknown rule metadata remain in the normal finding path.

Catalog infrastructure failures fail the Project analysis rather than producing an incomplete hotspot projection.

## Metrics and vulnerability separation

Hotspot candidates are partitioned before the immutable Project snapshot is built.

As a result, Security Hotspots are excluded from:

* Overall issue counts;
* New Code issue counts;
* Security rating calculations;
* `new_vulnerability`;
* severity-based New Code gate metrics;
* existing vulnerability-based Quality Gate calculations.

[...]

PR #296, opened 2026-07-19
feat(codequality): add immutable per-path measure snapshots

VietSory:
## Summary

Introduces the immutable per-path Measures snapshot foundation for #180.

Each successful Project analysis now persists a deterministic metric tree covering the Project root, directories, and source files. The snapshot retains additive counters and explicit availability metadata so later Measures API and web work can read historical analysis evidence without rescanning source code or consulting mutable runtime state.

This PR intentionally implements Phase A only. It does not add the Measures HTTP API or frontend explorer.

### Immutable measure snapshot

Added a flat, deterministically ordered snapshot containing:

* Project, directory, and file nodes;
* source files and lines of code;
* comment and blank lines;
* function counts and function-count availability;
* cyclomatic and cognitive complexity;
* covered and coverable lines;
* duplicated lines and distinct duplication blocks;
* issues grouped by catalog rule type;
* issues grouped by severity;
* remediation effort in minutes;
* explicit metric and attribution availability.

Derived percentages are calculated from persisted additive counters rather than stored or averaged child percentages.

### Per-file inventory

Extended code inventory analysis with immutable per-file facts while preserving the existing per-language inventory contract.

Each file fact retains:

* canonical relative path;
* detected language;
* code, comment, and blank-line counts;
* function count;
* whether the function count is known.

The source tree is still traversed only once.

### Canonical path contract

Added strict, platform-independent source path validation.

The snapshot:

* uses `/` as the persisted separator;
* accepts only relative paths;
* rejects Unix, Windows drive, and UNC absolute paths;
* rejects traversal, dot segments, repeated separators, and empty segments;
* rejects duplicate canonical inventory paths;
* fails closed when malformed paths appear in inventory, complexity, coverage, or duplication reports.

### Metric rollups

Directory and Project values are derived by summing descendant file counters.

The implementation does not average child percentages.

Coverage, comment density, and duplication density are derived from their corresponding numerators and denominators.

Overlapping duplication ranges are deduplicated per file, and duplication blocks shared by multiple files are counted once at directory and Project levels.

### Availability semantics

Missing measurements are represented as unavailable rather than as successful-looking zero or perfect values.

The snapshot distinguishes:

* analyzer not executed;
* analyzer executed with a legitimate zero result;
* unavailable function counts;
* incomplete issue-type classification;
* incomplete source-file attribution;
* incomplete technical-debt calculation.

New-code coverage remains explicitly unavailable because changed executable-line coverage is not yet retained.

### Rule-backed issue measures

Issue type is resolved exclusively from the immutable rule catalog:

* `bug`;
* `vulnerability`;
* `code_smell`;
* `security_hotspot`.

Security Hotspots are included in the Measures snapshot but do not contribute remediation effort.

Only rule-based findings are included in catalog-backed Measures issue and technical-debt calculations. Non-rule findings such as SCA findings remain part of the normal Project analysis metrics without making Measures completeness unavailable.

Unknown rules preserve severity evidence while marking dependent measures unavailable. Catalog infrastructure failures and invalid catalog rule types fail closed.

### Project analysis integration

The snapshot is now built during the real successful Project analysis pipeline and persisted as part of the append-only analysis payload.

The pipeline carries:

* per-file inventory;
* complete complexity reports;
* coverage reports;
* duplication reports, including valid zero-result reports;
* catalog-backed findings;
* Security Hotspot candidates.

Historical analyses are not recomputed from current source files or rule-catalog state.

### Persistence compatibility

Added snapshot cloning and compatibility behavior for existing analysis persistence.

* Memory persistence deep-clones snapshot nodes, maps, slices, and metric pointers.
* Legacy analysis payloads without a `snapshot` field decode into an explicitly unavailable legacy snapshot.
* Modern empty snapshots remain modern snapshots and are not misclassified as legacy data.

## Validation

Coverage includes:

* nested Project, directory, and file rollups;
* deterministic node ordering;
* Unix and Windows path normalization and rejection;
* duplicate-path rejection;
* additive counters and percentage derivation;
* complexity, coverage, and duplication availability;
* overlap-safe duplicated-line counting;
* distinct duplication-block rollups;
* unknown and invalid catalog rules;
* Security Hotspot classification;
* remediation-effort behavior;
* non-rule finding filtering;
* issue-attribution completeness;
* legacy analysis decoding;
* empty-current-snapshot round trips;
* memory-store mutation isolation;
* successful Project-analysis pipeline integration.

Local validation completed successfully for the affected modules:

```text
gofmt
go vet
go test
```

## Scope

This PR does not add:

* `GET /api/v1/projects/{key}/measures`;
* per-path Measures endpoints;
* OpenAPI Measures schemas;
* the web Measures tab;
* file-tree virtualization;
* changed-line coverage calculation;
* source-control diffing;
* mutable live Measures records.

Those remain follow-up phases of #180.

Refs #180


nghiadaulau (changes requested):
## Review — strict pass

I read every changed file and verified locally: `go build ./...`, `go vet`, `go test`, and `gofmt` are all **clean** on this branch. Independent go-arch review agreed the **dependency rule is clean** (`domain/measure` imports only `rule`/`shared`; no inward-pointing violations). This is a genuinely strong Phase-A foundation — the domain builder is pure and the test suite is excellent: it nails the hard parts (overlap-safe duplicated-line counting, shared-block dedup counted once per ancestor, **no** percentage averaging in rollups, coverage present/absent, `TechDebtAvailable` propagation). The availability model (analyzer-not-run vs legitimate-zero) and the back-compat `UnmarshalJSON` for legacy analyses are well done.

That said, holding the strict bar, a few things should land before merge.

### 🔴 Required

1. **Rebase — merge conflict with `main`.** #181 (Issues explorer) merged first and also edits `RecordProjectAnalysis` in `internal/usecase/projectuc/service.go`; a local test-merge conflicts there (the memory store auto-merges fine). Please rebase on `main` and re-resolve — the conflict overlaps exactly where you insert the snapshot build (issue-projection code now lives adjacent). Apologies, this one is on the merge order, not the PR.

2. **Doc comments on exported symbols** (`internal/domain/measure/snapshot.go`). `NodeKind`, `Availability`, `CountMetric`, `DecimalMetric`, `Counters`, `Node`, `Snapshot`, `IssueInput`, `RuleResolver`, `BuildSnapshotInput`, and the `CommentDensity`/`Coverage`/`DuplicationDensity` methods are undocumented. Only `CanonicalPath`/`BuildSnapshot` have comments — please document the rest to match the house convention.

### 🟡 Should fix

3. **`markAttributionUnavailable()` is O(issues × nodes)** (`snapshot.go:227-231`, called at `:369`/`:394`). Under the current Phase-A wiring every `IssueInput.Path` is `""` (`projectuc/service.go:485,491`), so this fires on **every** issue and re-sets the same flag repeatedly — quadratic on a large repo. Since no new nodes are created during the issues loop, track a sticky `attributionLost bool` in the loop and apply it once afterward.

4. **Latent frontend null-deref from the `Report.Duplication` value→`*pointer`+`omitempty` change** (`codequality/service.go:271`). All Go consumers are updated and the persisted `Analysis.Duplication` stays a value, so that's fine — but the code-quality report view derefs `report.duplication.*` with no null guard. Safe today only because `main.go` always wires `WithDuplication`; add a guard (or drop `omitempty`) so an unwired duplication scanner can't blank the field and crash the view.

### 🟢 Nits

5. `snapshot.go:472` comment says "Re-sort from children to parents" but nothing re-sorts (the slice is already child-first from `:428`) — correct code, misleading comment.
6. `snapshot.go:278-280` ranges `nodesByPath` while `getNode` mutates it; correct only because `ensureParents` walks the full chain to root — a snapshot of the keys would read clearer.

### Note on scope (not blocking)
Phase A attributes **all** issues to the root with `AttributionAvailable=false` because "the finding model lacks a path." Worth noting for Phase B: the path is already recoverable from `finding.DedupKey` (`<kind>:<rule>:<file>:<line>`) via `qualitygate.FileLineOf` — that's exactly how the Hotspots (#294) and Issues (#181) features attribute findings to files. Wiring that would make the per-path issue counts (the whole point of per-path measures) actually populate. Fine to defer, but it's the highest-value next step.

Great work overall — this is close. Once rebased with the doc comments + attribution loop addressed, I'm happy to approve.

VietSory:
Addressed all requested changes and final cleanup:

- rebased onto current main while preserving the #181 transactional issue and hotspot projection flow;
- added GoDoc for all exported measure-snapshot symbols;
- reduced attribution propagation to O(issues + nodes);
- stabilized parent creation and corrected the duplication rollup comment;
- made duplication explicitly optional and guarded all affected frontend consumers;
- preserved measured-zero semantics while rendering missing duplication as unavailable;
- added frontend regression coverage;
- normalized all touched frontend files to LF;
- removed the remaining unsafe test cast and trailing whitespace.

Local Go and frontend validation are green, including 229 frontend tests, typecheck, and production build. Ready for re-review.

nghiadaulau (commented):
## Re-review — the rebase + fixes look great, one small blocker left

Thanks for the fast turnaround. I re-verified locally (`go build`, `go vet`, `go test`, web `pnpm typecheck` + 229 tests on Node 20 — all green) and every substantive item from the last pass is resolved:

- ✅ **Rebased on `main`** — the `RecordProjectAnalysis` conflict with #181 is cleanly resolved; PR is `MERGEABLE`/`CLEAN`.
- ✅ **Doc comments** added across the exported `measure` symbols.
- ✅ **Attribution loop** — now a sticky `attributionLost` bool applied once (`snapshot.go:375,435`), no longer O(issues × nodes). 👍
- ✅ **Frontend null-deref** — `report.duplication` is now guarded everywhere in `CodeQualityReportView.tsx` (density, blocks, hint all fall back to "Unavailable").

### 🔴 One remaining blocker: CRLF line endings on the two new files

`internal/domain/measure/snapshot.go` and `internal/domain/measure/snapshot_test.go` are committed with **CRLF** line endings (527 and 392 `\r` lines in the blob), so they fail `gofmt`:

```
$ gofmt -l internal/domain/measure/
internal/domain/measure/snapshot.go
internal/domain/measure/snapshot_test.go
```

Every other file you touched is LF/gofmt-clean — it's just these two new ones (looks like the autocrlf wrestling from the earlier commits). `gofmt` clean is a house standard, so please normalize them to LF, e.g.:

```
sed -i '' 's/\r$//' internal/domain/measure/snapshot.go internal/domain/measure/snapshot_test.go
gofmt -w internal/domain/measure/snapshot.go internal/domain/measure/snapshot_test.go
git add -A && git commit -m "chore(measure): normalize snapshot files to LF (gofmt)"
```

Once `gofmt -l internal/domain/measure/` is empty, I'm happy to approve — everything else is good to go.

### Note (not this PR's job)
While checking, I noticed `gofmt -l internal/ cmd/` also flags several **pre-existing** files on `main` (`hotspot/review.go`, `qualitygate/metrics.go`, `memory/project_hotspot_store.go`, `rulecatalog/langpacks.go`, a few `_test.go`, etc.) — a mix of CRLF and stale `const`-block alignment. Those aren't yours; flagging separately so someone can do a one-shot `gofmt -w ./...` + LF normalization sweep (a `.gitattributes` with `*.go text eol=lf` would prevent recurrence).

VietSory:
Fixed the remaining CRLF/gofmt blocker in commit `5a3eaa7d290145117f3812aad9a33baf595d8ca0`.

Only these two files were changed:

* `internal/domain/measure/snapshot.go`
* `internal/domain/measure/snapshot_test.go`

Both committed blobs now contain zero carriage-return bytes, and:

```text
gofmt -l internal/domain/measure/snapshot.go internal/domain/measure/snapshot_test.go
```

produces no output.

Validation completed successfully:

```text
go build ./...    exit 0
go vet ./...      exit 0
go test ./...     exit 0
frontend tests    229 passed, exit 0
```

No Phase B files or branches were modified. Ready for re-review.
Thanks sir @nghiadaulau .

nghiadaulau (approved):
Approving — the LF normalization clears the last blocker: `gofmt -l internal/domain/measure/` is now empty and both new files are LF. Re-verified locally: `go build`/`vet`/`test` green, web typecheck + 229 tests green on Node 20, `MERGEABLE`/`CLEAN`.

All review findings are resolved: rebased cleanly on main (the #181 conflict), doc comments on the exported measure symbols, the attribution pass is now a sticky `attributionLost` bool (no longer O(issues×nodes)), and the frontend guards `report.duplication` everywhere. The domain builder is pure and the test suite rigorously covers the hard parts (overlap-safe duplicated-line counting, shared-block dedup, no percentage averaging, coverage/availability). Nice work. 🚢

(Separately, non-blocking: a few pre-existing files on main still fail `gofmt` — worth a one-shot `gofmt -w ./...` + `.gitattributes` `*.go text eol=lf` sweep in a follow-up.)

PR #306, opened 2026-07-21
Feat/180 b measure api

VietSory:
## Summary

This PR implements Phase B of the Project Measures Explorer defined in #180 by exposing the immutable measure snapshots introduced in #296 through a tenant-scoped, read-only HTTP API.

It adds:

* `GET /api/v1/projects/{key}/measures`
* Project, directory, and file-level measure drill-down
* Direct-child pagination with signed, analysis-pinned cursors
* Deterministic directory-first ordering
* Explicit `available`, `unavailable`, and `not_applicable` metric states
* Correct handling of incomplete issue classification and file attribution
* OpenAPI schemas, examples, and strict contract validation
* Production configuration for `SYNAPSE_MEASURE_CURSOR_SECRET`

The endpoint reads existing persisted snapshots only. It does not rescan source code or recompute stored rollups.

Security and consistency protections include tenant isolation, `PermView` authorization, HMAC-SHA256 cursor signing, cross-project cursor rejection, and stable pagination against the analysis selected on the first request.

Snapshot validation was also improved by removing the nested full-node file-child scan and replacing it with an O(1) parent lookup. Overall validation complexity remains O(N × tree depth) because cycle validation walks each node’s ancestor chain.

Regression and integration coverage includes canonical path validation, unavailable versus measured-zero semantics, issue availability reasons, tenant isolation, cursor tampering, analysis pinning, pagination, OpenAPI validation, and project/directory/file responses.

Validation completed successfully:

* `go build ./...`
* `go vet ./...`
* `go test ./...`


nghiadaulau (changes requested):
## Review — strict pass

The Measures API itself is strong work, but there's a **release-blocking regression from the rebase** that must be fixed before merge.

### 🔴 Blocker — this PR silently drops the Project Issues API (#181)

`git diff main...HEAD -- internal/adapter/httpapi/router.go` removes all four Project Issues route registrations and puts the new `measures` route in their place:

```diff
+   mux.HandleFunc("GET /api/v1/projects/{key}/measures", rt.authz(userdom.PermView, rt.getProjectMeasures))
-   mux.HandleFunc("GET  /api/v1/projects/{key}/issues", rt.authz(userdom.PermView, rt.listProjectIssues))
-   mux.HandleFunc("GET  /api/v1/projects/{key}/issues/{id}", rt.authz(userdom.PermView, rt.getProjectIssue))
-   mux.HandleFunc("POST /api/v1/projects/{key}/issues/{id}/transitions", rt.authz(userdom.PermReview, rt.transitionProjectIssue))
-   mux.HandleFunc("GET  /api/v1/projects/{key}/issues/{id}/history", rt.authz(userdom.PermView, rt.projectIssueHistory))
```

`origin/main` has these 4 routes; this branch has none of them. The handler functions (`listProjectIssues`, `transitionProjectIssue`, …) and `project_handler.go`'s issue methods still exist and compile, so `go build`/`go test` stay green — but the routes are no longer wired, so **the entire Issues explorer API would 404 after this merges**. This looks like a rebase conflict (the measures route landed exactly where #181's issues block was) resolved by discarding the issues routes.

Fix: restore the four issues route registrations alongside the new `measures` route. Please add a small router-level test asserting the issues + measures routes are all registered, so a future rebase can't silently drop them again.

### 🟡 gofmt

Two test files fail `gofmt` (trailing whitespace on blank lines) — `internal/usecase/projectuc/measures_test.go` and `internal/adapter/httpapi/project_measure_handler_test.go`. `gofmt -w` both.

### The rest looks good (verified locally)
`go build`/`vet`/`test ./...` pass, and I reviewed the security-sensitive parts closely — all solid:
- **Cursor signing is correct**: HMAC-SHA256 with `subtle.ConstantTimeCompare` for verification, `base64.RawURLEncoding`, deterministic struct marshal for sign/verify, version + kind-rank + non-empty-analysis validation on decode.
- **Cross-project / cross-tenant cursor replay is prevented** even though `ProjectID` isn't in the signed payload: the pinned analysis is loaded via `s.analyses.Get(p.TenantID, p.ID, cursor.AnalysisID)`, so a cursor from another project/tenant resolves to `ErrNotFound` → `ErrValidation`. Nice — the signature stops tampering and the scoped lookup stops replay. (Optional: also binding the project key into the payload would make the intent explicit, but it isn't required for correctness.)
- **Secret config is fail-closed**: production requires `SYNAPSE_MEASURE_CURSOR_SECRET` (≥32 bytes) and `os.Exit(1)`s without it; dev uses an ephemeral random key; the key is never logged.
- Strict query-param allowlist + duplicate rejection, `limit` bounded 1–200, canonical-path validation, and fail-closed `Snapshot.Validate()` / grade validation on read. Good test coverage (cursor tampering, analysis pinning, tenant isolation, unavailable-vs-measured-zero, OpenAPI contract).

Restore the issues routes + `gofmt`, and I'm happy to approve.

nghiadaulau (approved):
Approving. The one blocker — the rebase had silently dropped #181's four Project Issues route registrations from `router.go` (handlers still compiled, so build/test stayed green, but the endpoints would 404) — is now fixed: I pushed a commit to the branch (maintainer-edit) restoring the four routes alongside the new `measures` route in main's ordering, and `gofmt`-ing the two measure test files. Verified on this exact tree: `go build`/`vet`/`test ./...` clean, `gofmt` clean, and `router.go` now registers both the issues (4) and measures (1) routes; `MERGEABLE`/`CLEAN`.

The Measures API itself is solid, as noted before: HMAC-SHA256 cursors with constant-time verification; cross-project/tenant replay prevented via the tenant+project-scoped `analyses.Get(cursor.AnalysisID)`; fail-closed prod secret config (never logged); strict param validation and fail-closed snapshot/grade checks; strong test coverage. Follow-up worth doing: a router-level test asserting the full project route set is registered, so a future rebase can't silently drop routes again. 🚢

PR #319, opened 2026-07-22
Feat/210 xml injection rules

VietSory:
## Summary

This PR implements the complete XML `inj` rule family for #210, adding eight parse-based security rules to the existing structured XML analyzer.

### Rules added

* `xml:external-entity`
* `xml:external-dtd`
* `xml:external-parameter-entity`
* `xml:xinclude`
* `xml:entity-expansion`
* `xml:doctype-present`
* `xml:external-schema-location`
* `xml:hardcoded-secret`

The family includes four vulnerability rules and four security-hotspot rules covering XXE, external DTDs, parameter entities, XInclude, entity-expansion attacks, unsafe DOCTYPE usage, external schema loading, and literal credentials in XML configuration.

## Implementation

The existing XML analysis pipeline was extended rather than introducing a second analyzer.

Key changes include:

* Generalized XML rule descriptors to support reliability and SAST findings.
* Preserved the existing `xml:not-well-formed` and `xml:duplicate-attribute` rules.
* Added a bounded, quote-aware and bracket-aware lexical DTD state machine.
* Added namespace-aware XInclude and XML Schema Instance detection.
* Added element-stack tracking for hardcoded-secret detection.
* Added bounded entity graph analysis with cycle detection and saturating expansion estimates.
* Continued scanning after expansion-analysis limits so external entities cannot evade detection.
* Registered all declared general entities with inert decoder replacements to avoid false malformed-XML findings.
* Kept all analysis static and local: no external files, URLs, DTDs, schemas, or XInclude resources are resolved.

## Noise reduction and correctness

The implementation includes targeted suppression and false-positive protections:

* External entities are not misclassified as external DTDs.
* General and parameter entities are reported by separate rules.
* Generic `doctype-present` findings are suppressed when a more specific DTD rule applies.
* XInclude detection uses the resolved namespace rather than a literal prefix.
* `schemaLocation` evaluates only schema-location positions, not namespace URI positions.
* Credential placeholders, masked values, numeric values, short literals, and external secret references are ignored.
* Entity replacements never introduce artificial hardcoded-secret findings.
* Secret values are never included in findings, logs, descriptions, or test output.
* XML markup matching remains case-sensitive.

## Rule Catalog integration

All eight rules include clean-room metadata:

* type and impacted quality;
* default severity;
* CWE and OWASP references;
* descriptions and rationales;
* remediation guidance;
* compliant and non-compliant examples;
* parse-based detection metadata.

[...]

PR #323, opened 2026-07-22
Feat/210 xml correctness rules

VietSory:
## Summary

This PR completes the XML correctness/bug rule family with 8/8 rules as part of #210.

It adds six new XML correctness rules:

* `xml:mismatched-tag`
* `xml:unclosed-element`
* `xml:multiple-root-elements`
* `xml:undeclared-prefix`
* `xml:invalid-character-reference`
* `xml:invalid-comment`

Together with the existing `xml:not-well-formed` and `xml:duplicate-attribute` rules, the XML bug family now contains all eight planned rules.

## Implementation

* Adds a structural XML scanner for element-stack validation, namespace resolution, comments, character references, and duplicate expanded attributes.
* Validates complete start and end tags before mutating structural state.
* Uses strict QName validation for element and attribute names.
* Tracks accurate source lines for attribute names and values.
* Runs the standard XML parser first and uses its exact failure offset as a hard barrier for structural analysis.
* Suppresses `xml:not-well-formed` only when a more specific terminal finding corresponds to the same parser failure.
* Stops structural recovery after the first unrecoverable parser failure to prevent cascading findings.
* Moves raw duplicate-attribute detection into the barrier-aware structural scanner.
* Preserves the previously added XML injection-rule behavior.

## Testing

Added unit and full-pipeline regression coverage for:

* Mismatched, unclosed, and multiple root elements
* Malformed and incomplete start/end tags
* Raw and namespace-expanded duplicate attributes
* Declared and undeclared namespace prefixes
* Decimal and hexadecimal character references
* Multiline attribute source locations
* Valid, invalid, and unterminated comments
* Comments and declarations inside internal DTD subsets
* CDATA, processing instruction, and DOCTYPE termination
* Exact parser-failure barriers
* Cascading-finding prevention
* Malformed XML followed by otherwise valid tokens
* Existing XML injection-rule scenarios

Validation performed:

```bash
gofmt -l .
git diff --check
go test ./...
golangci-lint run
```

All checks pass.

Refs #210


nghiadaulau (approved):
## ✅ Approve — strict review, completes the `bugs` (correctness) family of #210

Reviewed hard against the issue's acceptance (findings produced · each rule catalogued, golden-tested, in the default profile · clean-room), built + tested the branch, and ran an independent security audit of the new untrusted-input parser. It holds up, and it does so **without regressing #319** — the catalog now carries all **16** XML rules (8 `inj` preserved + 8 `bugs`).

### Target verification (issue #210, `bugs` family = 8/8)
- **6 new correctness rules**: `mismatched-tag`, `unclosed-element`, `multiple-root-elements`, `undeclared-prefix`, `invalid-character-reference`, `invalid-comment` — plus the existing `not-well-formed` / `duplicate-attribute` = 8/8.
- **Catalogued**: each is `TypeBug` / Reliability, correct severity (Medium structural, Low for comment syntax), empty CWE/OWASP (correct for correctness bugs — matches the existing bug rules), with clean-room rationales citing precise **W3C spec sections** (`#dt-etag`, `#sec-logical-struct`, `#sec-well-formed`, `xml-names`, `#NT-CharRef`, `#sec-comments`) + compliant/non-compliant examples. Inventory (`rule_keys.txt`) → 16.
- **Default profile**: auto-included (built-in profile derives from the catalog).
- **Golden-tested**: the parity test drives all 16 (catalog ↔ engine ↔ actual finding kind/CWE/severity/title, each non-compliant example must trigger) + 14 dedicated structural tests with exact rule/count/line assertions and edge cases (declared-prefix → 0 findings, char-ref & attribute line tracking, DTD comments, duplicate raw vs namespace-expanded attrs, mismatched-tag-only suppression).

### Design quality
- Hand-written structural scanner: strict QName validation, XML-1.0-correct `isValidCharRef` ranges (and correctly rejects uppercase `&#X`), dual duplicate-attribute detection (raw QName + `{uri}local`), namespace-aware undeclared-prefix resolution (incl. reserved `xml`/`xmlns`).
- Sound **barrier**: the standard parser runs first and its failure `InputOffset()` is a hard stop for structural recovery → no cascading findings; the moved duplicate-attribute detection is now barrier-aware; `not-well-formed` is suppressed only when a specific structural terminal sits at that same offset (specific rule replaces generic).

### Verification run (on the branch)
- `go build ./...` ✓ · `go vet` ✓ · `golangci-lint run` → **0 issues** ✓ · `go test -race` (codeanalysis, rulecatalog, codequality) ✓ · `mergeable = CLEAN`

### Security (independent audit — SOUND on all four)
- **No scanner DoS**: `scanXMLStructural` is O(n) — every outer-loop branch advances the index or breaks; unterminated `<!--`/`<![CDATA[`/`<?` scan once to EOF then break (no quadratic re-scan).
- **Panic-safe** on malformed/truncated input: every `content[x]` / slice expression is bounds-guarded; stacks kept in lockstep; `strings.Split` results accessed only under `len==2`.
- **Barrier suppression** deletes the single `not-well-formed` finding from a *different* slice than the one being ranged, then breaks — index-safe, can't drop the wrong finding.
- **No XXE**: `encoding/xml` resolves no external entities/URLs; `dec.Entity` is fed only locally-parsed inert values.

### Non-blocking note
After this + #319, only the `maint` family (14 rules) remains to reach #210's ~30-rule parity target — correctly left as follow-up (`Refs #210`, not Closes).

Clean, careful work — merging.

Issue #209, opened 2026-07-12
feat(codequality): HTML rule pack (correctness + accessibility)


nghiadaulau:
Part of #185 (tracker) / #174. **Authoring guide:** [https://github.com/KKloudTarus/synapse-ce/blob/main/docs/guide/code-quality-rules.md](https://github.com/KKloudTarus/synapse-ce/blob/main/docs/guide/code-quality-rules.md). Detection: a tree-sitter HTML grammar, or a token/DOM analyzer. Rules are clean-room, our own content; cite the concept origin, never third-party rule prose.

## Depth plan - parity target ~50 rules

Targets ~50 rules (mature-profile parity), decomposed by rule family - claim a family:

| Family | Target |
|---|---|
| `bugs` | 10 |
| `authz` | 8 |
| `maint` | 32 |
| **Total** | **50** |

**Signature families:** Accessibility (`maint` -> WCAG: alt text, labels, lang, tabindex) is the bulk; plus security `authz` (tabnabbing, inline handlers) and correctness `bugs` (malformed markup).

## Authoritative sources
[HTMLHint rules](https://htmlhint.com/docs/user-guide/list-rules) - [WCAG](https://www.w3.org/WAI/standards-guidelines/wcag/) - [MDN HTML](https://developer.mozilla.org/en-US/docs/Web/HTML) - [W3C validator](https://validator.w3.org/)

## Seed rules (initial security + correctness core; fill the families toward the target)
| Key | Type / Sev | Flags | Ref |
|---|---|---|---|
| `html:img-alt-missing` | code_smell / medium | `<img>` without `alt` | WCAG 1.1.1 |
| `html:label-missing` | code_smell / medium | form control without an associated label | WCAG 3.3.2 |
| `html:missing-lang` | code_smell / low | `<html>` without a `lang` attribute | WCAG 3.1.1 |
| `html:target-blank-noopener` | vulnerability / medium | `target=_blank` without `rel=noopener` | [CWE-1022](https://cwe.mitre.org/data/definitions/1022.html) |
| `html:inline-event-handler` | security_hotspot / medium | inline `on*=` JavaScript (CSP/XSS) | [CWE-79](https://cwe.mitre.org/data/definitions/79.html) |
| `html:duplicate-id` | bug / low | duplicate `id` attribute | W3C |
| `html:deprecated-tag` | code_smell / low | deprecated element (`<center>`, `<font>`, ...) | MDN |
[...]

PR #337, opened 2026-07-25
feat(codequality): add HTML security rule family

VietSory:
## Summary

Implements the complete HTML security rule family from Issue #209, adding
8 AST-backed rules for unsafe navigation, inline JavaScript surfaces, iframe
sandboxing, insecure transport, and missing Subresource Integrity metadata.

This is PR 2 of the planned HTML rule pack:

- Correctness: 10/10
- Security: 8/8
- Maintainability and accessibility: 0/32
- HTML total: 18/50

## Rules added

- `html:active-content-over-http`
- `html:external-resource-no-integrity`
- `html:iframe-sandbox-escape`
- `html:iframe-sandbox-missing`
- `html:inline-event-handler`
- `html:insecure-form-action`
- `html:javascript-url`
- `html:target-blank-noopener`

Rule classification:

- 2 vulnerability rules
- 6 security-hotspot rules
- All rules use AST detection and the `security` quality

## Implementation

- Reuses the existing HTML tree-sitter traversal and `htmlCollector`.
- Keeps the existing shared limits:
  - 20 findings per rule per file
  - 100 total HTML findings per file
  - 256 attributes inspected per tag
  - 256 traversal and ancestor depth
- Adds bounded attribute, token, and URL-classification helpers.
- Preserves deterministic source-order findings across correctness and security
  rules on the same tag.
- Extends runtime metadata, catalog metadata, rule inventory, and the built-in
  HTML profile from 10 to 18 rules.

### Navigation and JavaScript surfaces

- Detects `target="_blank"` on links, areas, and forms without explicit
  `noopener` or `noreferrer`.
- Detects non-empty inline `on*` event-handler attributes.
- Detects literal `javascript:` URLs in supported navigation and form
  attributes.
- Decodes HTML character references before semantic comparisons.
- Handles mixed-case and control-character-obfuscated URL schemes.
- Avoids false positives from double-encoded character references.
- Suppresses URL classification for recognized template expressions.

### iframe sandboxing

- Detects active iframes with `src` or `srcdoc` but no `sandbox` attribute.
- Detects the risky combination of:
  - `allow-scripts`
  - `allow-same-origin`
- Tokenizes decoded `sandbox` values using ASCII whitespace.
- Reports against the first usable source attribute in document order.

### Transport and integrity

- Detects absolute `http://` form submission targets.
- Detects active scripts, iframes, objects, embeds, and stylesheets loaded over
  absolute `http://` URLs.
- Detects absolute HTTP or HTTPS scripts and stylesheets without a complete,
  non-empty `integrity` attribute.
- Does not classify relative, root-relative, protocol-relative, `data:`,
  `blob:`, or template-generated URLs as absolute external resources.
- Allows independent findings to coexist, such as insecure HTTP transport and
  missing SRI on the same resource.

## Parsing and recovery

- Uses ASCII-only case folding for HTML names and tokens.
- Inspects only the first complete occurrence of duplicate attributes.
- Fails closed for incomplete tags and malformed local attribute state.
- Does not scan comments, raw script/style content, `ERROR` payloads, or
  attribute text for synthetic markup.
- Continues processing valid sibling nodes after malformed markup.
- Preserves CRLF line mapping.
- Emits findings in source-offset order with deterministic rule tie-breaking.

## Catalog integration

Adds catalog metadata for all 8 security rules, including:

- type and severity;
- security quality;
- CWE mappings;
- rationale with official HTML, CSP, Mixed Content, SRI, and CWE references;
- remediation guidance;
- compliant and noncompliant examples;
- AST detection metadata.

After this change:

```text
HTML catalog rules: 18
HTML runtime rules: 18

bug:              10
vulnerability:     2
security_hotspot:  6

### CODE REVIEW

**Review by nghiadaulau — CHANGES_REQUESTED** [review]
## Review — Changes requested ❌

Reviewed at `c4fd378a`. The feature itself is solid — but it **fails `golangci-lint` (CI-exact v2.12.2, exit 1)**, so the `Lint (Go)` CI job goes red and it's not mergeable yet.

### 🔴 Blocker — 4 lint issues
```
html_security_cgo.go:33   func htmlFirstAttribute is unused        (unused)
html_security_cgo.go:358  func inspectHTMLSecurityTag is unused    (unused)
html_security_cgo_test.go:260  ST1023: omit type string from var declaration  (staticcheck)
html_security_cgo_test.go:277  ST1023: omit type string from var declaration  (staticcheck)
```
- **Two unused functions** are dead code — notably a whole `inspectHTMLSecurityTag(...)` security-inspection function that's never called (the security rules actually fire via a different path — see below), plus `htmlFirstAttribute`. Remove them (or wire `inspectHTMLSecurityTag` if it was the intended dispatch and the current path is the accidental one).
- **ST1023 ×2**: change `var src string = "<!doctype html>\n"` → `src := "<!doctype html>\n"` in the two test cases.

Run `golangci-lint run ./internal/infrastructure/tools/astwalk/...` — must be **0 issues** before merge.

### ✅ What's already good (verified locally)
- Delivers the full HTML **security/`authz` family = 8/8** — exactly issue #209's `authz` target (10 correctness + 8 security = 18/50). 2 vulnerability + 6 security-hotspot, all `security` quality, AST detection.
- All HTML tests pass; full `astwalk` package + `rulecatalog` (metadata/catalog/golden/parity) green — **no regression** to the correctness rules or python/css/js.
- Builds clean **CGO and CGO-free**.
- e2e: a real `.html` fires `target-blank-noopener`, `javascript-url`, `inline-event-handler`, `iframe-sandbox-missing`.
- The `html_quality_cgo.go` (+85/-22) change is a sound refactor (shared attribute-collection reused by both correctness + security); correctness rules still fire (full test green).

### To land
1. Delete the two unused functions (or wire them in).
2. Fix the two ST1023 `var` declarations.
3. `golangci-lint` green + push — happy to re-review immediately; scope/target/behavior are all correct, this is purely a lint/dead-code cleanup.

nghiadaulau (approved):
## Re-review — Approved ✅

Re-reviewed at `8cc926e`. My previous `CHANGES_REQUESTED` blockers are **all resolved**:

- ✅ **golangci-lint v2.12.2 (CI-exact): 0 issues** — the two unused functions (`htmlFirstAttribute`, `inspectHTMLSecurityTag`) are removed, and the two ST1023 `var` declarations in the test are fixed.
- ✅ Delivers the full HTML **security/`authz` family = 8/8** — issue #209's target (10 correctness + 8 security = 18/50). 2 vulnerability + 6 security-hotspot, AST detection, `security` quality.
- ✅ Full `astwalk` + `rulecatalog` (metadata/catalog/golden/parity) green — no regression; builds clean CGO and CGO-free.
- ✅ e2e: a real `.html` fires `target-blank-noopener`, `javascript-url`, `inline-event-handler`, `iframe-sandbox-missing`.

Clean cleanup, scope/target/behaviour all correct. `maint`/accessibility (32) remains the scoped follow-up so #209 stays open. Merging.

PR #347, opened 2026-07-28
fix(code-quality): retain immutable source snapshots

### LINKED ISSUE #345: fix(code-quality): retain immutable source snapshots in full Compose stack
## Problem

The Code Quality UI showed **Source preview unavailable** / **This captured file is unavailable: Not retained** for newly completed analyses in the full Docker Compose stack.

## Root cause

`scancache` was mounted on `/home/synapse`. Docker applies the named-volume mount after image build, hiding the image-created and `synapse`-owned `/home/synapse/project-source-artifacts` directory. The API runs as UID `10001`, so `sourceartifact.Store.Capture` could not create its snapshot directory and degraded the analysis to `capture_failed`. The analysis then persisted no source manifest entry, which the Code workspace correctly displayed as `not_retained`.

## Resolution

Fixed in `deploy/docker-compose.full.yml` by:

- adding a dedicated `project-source-artifacts` named volume;
- adding a root one-shot initializer that runs `chown -R 10001:999`; and
- making `synapse-api` wait for that initializer before startup.

## Verification

A fresh analysis after the change returned `200` from `/code/file` and rendered captured `Makefile` lines in the Code Quality workspace. Analyses created before the fix correctly remain unavailable because their snapshots were never captured.

## Scope

This tracks the production/dev Compose regression and its fix. CLI-created Code Quality snapshots are tracked separately.

pho-veteran:
## Summary

Close #345 

Fix Code Quality source/diff snapshots becoming unavailable after analysis:

- decouple persisted Unified/Split diff rendering from source-artifact reads;
- retain strict validation for Source view reads;
- provision a dedicated, initialized source-artifact volume in the full Compose stack so the non-root API process can persist immutable snapshots.

## Root cause

`scancache` was mounted at `/home/synapse`, masking the image-created
`/home/synapse/project-source-artifacts` directory. The mounted volume was
root-owned while the API runs as UID `10001`, so source capture failed and the
analysis persisted unavailable source metadata.

A fresh analysis now returns `200` from `/code/file` and renders captured
source in the Code Quality workspace. Snapshots from analyses created before
the fix remain unavailable because they were never captured.

## Tests

- `go test ./internal/usecase/projectuc ./internal/adapter/httpapi ./internal/infrastructure/sourceartifact`
- `cd web && pnpm exec vitest run src/components/codequality/ProjectCodeWorkspace.test.tsx src/lib/api.projects.test.ts`
- `cd web && pnpm typecheck`
- `cd web && pnpm build`
- `docker compose -f deploy/docker-compose.full.yml config --quiet`
- `git diff --check`

### DIFF
--- a/deploy/docker-compose.full.yml
+++ b/deploy/docker-compose.full.yml
@@ -63,20 +63,32 @@ services:
       mc mb --ignore-existing local/${BLOB_BUCKET:-synapse-evidence} &&
       echo 'bucket ready'"
 
+  # A named volume is mounted after image build, so Docker creates it root-owned;
+  # initialize it before the non-root API starts.
+  project-source-artifacts-init:
+    image: alpine:3.22
+    user: "0:0"
+    command: ["sh", "-c", "chown -R 10001:999 /project-source-artifacts"]
+    volumes:
+      - project-source-artifacts:/project-source-artifacts
+
   synapse-api:
-    build:
-      context: ..
-      dockerfile: deploy/Dockerfile
-      target: full            # batteries-included: JDK + Maven + Gradle + git + syft + grype
     depends_on:
+      project-source-artifacts-init:
+        condition: service_completed_successfully
       postgres:
         condition: service_healthy
       minio-init:
         condition: service_completed_successfully
+    build:
+      context: ..
+      dockerfile: deploy/Dockerfile
+      target: full            # batteries-included: JDK + Maven + Gradle + git + syft + grype
     ports:
       - "8080:8080"
     volumes:
       - scancache:/home/synapse   # persist the Maven/Gradle caches across CLI scans
+      - project-source-artifacts:/home/synapse/project-source-artifacts
       # - ../:/scan:ro            # uncomment to `synapse-cli scan /scan` this repo
     healthcheck:
       test: ["CMD-SHELL", "curl -fsS localhost:8080/healthz || exit 1"]
@@ -170,3 +182,4 @@ volumes:
   pgdata:
   miniodata:
   scancache:
+  project-source-artifacts:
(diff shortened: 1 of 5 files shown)


nghiadaulau (approved):
## Review — Approved ✅

Verified against issue #345. Three-part fix, all sound:

1. **Root cause (Compose)** — `deploy/docker-compose.full.yml`: adds a dedicated `project-source-artifacts` named volume, a root one-shot init (`chown -R 10001:999`), and makes `synapse-api` wait on it (`service_completed_successfully`). Directly fixes the volume-masking + root-ownership issue that made `sourceartifact.Store.Capture` fail for the UID-10001 API. `docker compose config --quiet` validates.
2. **Diff decoupling (backend)** — `code_viewer.go` removes the `LoadBase`/`Load` source-artifact reads from the **diff** path, so Unified/Split diffs render from persisted patch data and no longer error on `capture_failed`. The **Source** view path is untouched → strict validation retained (httpapi tests confirm).
3. **Frontend** — `ProjectCodeWorkspace.tsx` renders diffs from persisted data; contract change is consistent across backend + frontend + updated tests.

**Verification (local):**
- Backend: `go test ./internal/usecase/projectuc ./internal/infrastructure/sourceartifact ./internal/adapter/httpapi` — all pass; build OK.
- Web: `tsc --noEmit` clean; full suite **265/265 pass** (incl. `ProjectCodeWorkspace.test.tsx`); `vite build` OK.
- `docker compose -f deploy/docker-compose.full.yml config` valid.

Correctly scoped, no regression, and matches the issue's diagnosis (pre-fix analyses stay unavailable because they were never captured — acceptable). Resolves #345. Merging.

PR #348, opened 2026-07-29
feat(codequality): complete HTML maintainability rule family

### LINKED ISSUE #209: feat(codequality): HTML rule pack (correctness + accessibility)

VietSory:
## Summary

Completes the HTML rule pack from #209 by adding the final 32
maintainability and accessibility rules.

This is the third and final scoped PR:

- Correctness: 10/10
- Security: 8/8
- Maintainability and accessibility: 32/32
- Total HTML rules: 50/50

## What changed

Adds AST-backed HTML checks covering:

- image alternatives and accessible names;
- document titles, language, headings, landmarks, fieldsets, and tables;
- keyboard focus, tabindex, accesskey, autofocus, and nested controls;
- ARIA names, references, roles, required properties, and hidden focusable
  content;
- meta refresh, viewport zoom restrictions, media autoplay, and obsolete HTML
  features.

## Architecture

- Reuses the existing tree-sitter HTML traversal.
- Builds bounded document facts for IDs, labels, names, and structural
  relationships.
- Uses fail-closed handling for malformed input, templates, incomplete facts,
  and exhausted scan bounds.
- Merges correctness, security, and maintainability candidates before
  deterministic source-order emission.
- Shares the existing limits of 20 findings per rule and 100 total HTML
  findings per file.
- Keeps the rules isolated to `.html` files.

## Rule inventory

Adds exactly 32 code-smell rules with maintainability quality metadata:

- 11 text-alternative and accessible-name rules
- 6 document-structure and landmark rules
- 6 keyboard and focus rules
- 5 ARIA consistency rules
- 4 meta, media, and obsolete-HTML rules

After this PR, catalog, runtime metadata, and the built-in HTML profile each
contain exactly 50 HTML rules.

## Validation coverage

The test suite includes:

- positive and negative fixtures for all 32 rules;
- exact finding-line assertions;
- entity decoding and double-encoding regressions;
- template and first-complete duplicate-attribute cases;
- malformed markup and raw-text isolation;
- bounded and cycle-safe accessible-name relationships;
- focusability and hidden-state matrices;
- deprecated element and attribute inventory coverage;
- cross-family source ordering;
- per-rule and shared global caps;
- 50-run deterministic output checks;
- real-file `QualityFor` end-to-end coverage;
- runtime, catalog, profile, type, quality, severity, and source parity;
- language isolation for non-HTML extensions.

Closes #209


pho-veteran (changes requested):
## 🔴 Request changes — strong completion, but 3 blockers before merge

Reviewed strictly at `d5fb108e` against PR #348's authoritative GitHub diff, with separate architecture and security passes. The overall implementation is thoughtful and a substantial amount is done correctly, but two precision defects and one hostile-input availability issue need fixing before merge.

### ✅ What's done well

- The PR completes the intended **32-rule HTML maintainability/accessibility family** and keeps catalog, runtime metadata, golden keys, and the built-in HTML profile aligned.
- The implementation stays in `internal/infrastructure`; **dependency direction passes** — no domain/use-case layer imports infrastructure.
- Good defensive design in several places: bounded ID tracking, bounded/cycle-safe accessible-name references, malformed/template/foreign-content guards, deterministic source-order emission, and existing per-rule/per-file finding caps.
- The existing HTML correctness/security families remain wired into the shared candidate pipeline rather than being replaced.
- Security-sensitive review is otherwise clean: no command execution, secret handling, sandbox/scope/audit/LLM-boundary change, unsafe pointer conversion, or external HTML execution was introduced.

### ⛔ 1. Hostile sibling lists cause quadratic analysis work

`internal/infrastructure/tools/astwalk/html_maintainability_cgo.go:661-681`, `:696-720`, `:724-739`

`htmlFirstElementChild` asks for the first element child by calling `htmlDirectElementChildren`, which walks **all** children and allocates a complete slice. `htmlDisabledByFieldset` invokes this for every candidate descendant of a disabled fieldset. A fieldset containing a first `<legend>` followed by many sibling `<button>` elements therefore rebuilds the full sibling slice once per button: O(n²) tree-sitter boundary calls and allocations on untrusted repository input.

`htmlIsFirstSummary` has the same shape: every `<summary>` scans/materializes the parent's complete direct-child list again.

**Fix:** make `htmlFirstElementChild` iterate and return as soon as the first element child is found. Make `htmlIsFirstSummary` scan children directly and stop at the first `<summary>`; do not allocate child slices for first-child/existence queries. Add a large-sibling bounded-work regression/benchmark for disabled fieldsets and details/summary.

### ⛔ 2. Case-sensitive ARIA/`headers` ID references are lowercased

`internal/infrastructure/tools/astwalk/html_maintainability_cgo.go:525`, `:1358`, `:1467`

The new IDREF consumers reuse `htmlAttributeTokens`, which lowercases every token, while `facts.idElements` stores IDs with their original case. HTML IDs and IDREF/IDREFS values are case-sensitive, so valid markup such as:

```html
<span id="SaveLabel">Save</span>
<button aria-labelledby="SaveLabel"></button>
```

looks up `savelabel` and can emit false `aria-reference-missing` and accessible-name findings. Mixed-case `headers` references fail similarly. Existing tests only cover lowercase IDs.

**Fix:** add a whitespace tokenizer that preserves case and use it for `aria-labelledby`, the other ARIA relationship attributes, and `headers`. Keep the lowercase tokenizer only for case-insensitive enumerated values such as roles. Add mixed-case ARIA accessible-name/reference and table-header regressions.

### ⛔ 3. Valid `<area>` attributes are classified as deprecated

`internal/infrastructure/tools/astwalk/html_maintainability_data_cgo.go:90`

The deprecated-attribute table lists `hreflang`, `type`, and `nohref` for `<area>`. `hreflang` and `type` remain conforming hyperlink attributes when `href` is present; only `nohref` belongs in this deprecated set. The generated table test currently locks in false `html:deprecated-attribute` findings and uses fixtures without the valid hyperlink context.

**Fix:** remove `hreflang` and `type` from the `area` deprecated set, retain `nohref`, and add a negative regression such as:

```html
<area href="/fr" hreflang="fr" type="text/html" alt="Français">
```

Once these three are fixed with regressions, the architecture and remaining safety design look ready for approval. Nice work overall, @VietSory — this is close.

pho-veteran (changes requested):
## Re-review — 3 prior blockers fixed; 1 fieldset-semantics blocker remains ❌

Re-reviewed at `02d2d75c` against the exact delta from the previously reviewed head `d5fb108e`. The requested performance, IDREF, and deprecated-attribute fixes are present and covered by regressions. One correctness blocker remains in the new fieldset cache and must be resolved before approval.

### ✅ Prior findings resolved

1. **Quadratic direct-child scans — fixed.**
   - `html_maintainability_cgo.go:203-242` now builds first-child/first-summary indexes in one pass.
   - `htmlFirstFieldsetElementChild` and `htmlIsFirstSummary` use constant-time map lookups instead of rebuilding a full child slice for every descendant/sibling.
   - `html_maintainability_cgo_test.go:441-465` adds large disabled-fieldset/details sibling regressions.

2. **Case-sensitive ARIA and `headers` IDREFs — fixed.**
   - `html_security_cgo.go:51-75` introduces `htmlAttributeIDReferenceTokens`, which preserves decoded token case; the existing generic token helper still lowercases case-insensitive enumerated values such as roles.
   - ARIA accessible-name/reference and table `headers` resolution now use the case-preserving helper at `html_maintainability_cgo.go:591`, `:1428`, and `:1537`.
   - `html_maintainability_cgo_test.go:467-491` covers exact-case success and mismatched-case failure for both ARIA and table references.

3. **Valid `<area>` hyperlink attributes marked deprecated — fixed.**
   - `html_maintainability_data_cgo.go:90` now retains only `nohref` in the `<area>` deprecated set; valid `hreflang` and `type` were removed.
   - `html_maintainability_cgo_test.go:493-498` adds the requested valid hyperlink regression.

The fix delta also preserves clean dependency direction; no architecture-layer violation was introduced.

### ⛔ Remaining blocker — disabled `<fieldset>` uses the first element child instead of the first `<legend>` child

`internal/infrastructure/tools/astwalk/html_maintainability_cgo.go:227-230`, `:737-742`, `:756-779`

The new cache stores `firstFieldsetChildren[parentOffset]` as the fieldset's first direct **element child**. `htmlDisabledByFieldset` then exempts descendants only if that element is a `<legend>`.

That is correct for the separate accessible-name rule, where a fieldset's legend must be its first element child to supply the fieldset name. It is **not** the HTML disabled-fieldset rule. Disabledness exempts descendants of the fieldset's first direct **`<legend>` element child**, even when another element precedes that legend.

Reproducer:

```html
<fieldset disabled>
  <span></span>
  <legend><button>Enabled</button></legend>
</fieldset>
```

Expected: the button is inside the fieldset's first direct `<legend>`, so it is not disabled by the fieldset and remains sequentially focusable.

Current behavior: the cache records `<span>` as the first element; `htmlDisabledByFieldset` sees that it is not a legend and marks the button disabled. That incorrect focusability result propagates into rules including `html:aria-hidden-focusable`, `html:nested-interactive-control`, and focus-dependent ARIA applicability, causing false negatives.

The current large-sibling regression at `html_maintainability_cgo_test.go:442-447` places only comments before `<legend>`; comments are not elements, so it does not exercise this semantic distinction.

**Required fix:**

- Keep the current first-direct-element cache for `htmlNameFieldset`, because the accessible-name/legend-first-child rule needs it.
- Add a separate cache for the first direct **legend** of each fieldset (for example, `firstFieldsetLegends`). Populate it during the same one-pass indexing loop when `parentName == "fieldset" && childName == "legend"`.
- Make `htmlDisabledByFieldset` use the first-legend cache, not `firstFieldsetChildren`.
- Add a regression where an ordinary element precedes the first legend and a focusable descendant inside that legend must remain focusable. A direct observable assertion would be:

```html
<div aria-hidden="true">
  <fieldset disabled>
    <span></span>
    <legend><button>Enabled</button></legend>
  </fieldset>
</div>
```

This should emit `html:aria-hidden-focusable`. Also keep a case proving a button outside that first legend remains disabled.

Once the fieldset caches represent these two distinct semantics and the regression passes, the previous blockers are resolved and this should be ready for approval.

pho-veteran (approved):
## Re-review — Approved ✅

Re-reviewed at `f2c0c9ab`. All blockers from the previous rounds are resolved:

- ✅ **Hostile sibling-list performance:** fieldset/details direct-child facts are indexed once and consumed via constant-time lookups; the prior repeated full-sibling materialization is gone, with large-sibling regressions added.
- ✅ **Case-sensitive IDREF semantics:** ARIA relationship attributes and table `headers` now use a case-preserving decoded-token helper, while roles and other case-insensitive enumerated tokens remain normalized. Exact-case success and mismatched-case failure are test-covered.
- ✅ **`<area>` deprecated inventory:** valid `hreflang` and `type` were removed; obsolete `nohref` remains, with a valid hyperlink regression.
- ✅ **Disabled-fieldset semantics:** the implementation now keeps separate first-direct-element and first-direct-`legend` caches. Fieldset naming retains first-element semantics, while disabledness correctly exempts descendants of the first direct legend even when another element precedes it. Tests cover both the focusable button inside that legend and a disabled control outside it.

The additional hardening in the final delta is sound: incomplete bounded traversal suppresses maintainability analysis conservatively; template-valued applicability is treated as unknown; meta-refresh delay parsing is tightened; negative `maximum-scale` no longer produces a zoom-limit finding; and `role-img-name-missing` reports the role attribute line.

Verification on the exact head: CGO-free `astwalk`/`rulecatalog` tests pass, CGO-free `go vet` passes, `git diff --check` passes, GitGuardian is green, and the dependency rule remains clean. No new correctness, security/resource-exhaustion, or architecture blocker found.

Nice follow-through, @VietSory. Approved — leaving merge to the maintainer.

Issue #392, opened 2026-08-05
feat(ai-triage): observability, budget, and resilience (P1.5)


nghiadaulau:
Parent: #387 (AI-assisted false-positive triage).

## Goal

Control stability, cost, and model behaviour in production.

## Scope

- Metrics: request count, latency, timeout, parse failure, token/cost, disagreement rate, and exemption rate.
- Configurable concurrency, per-scan finding cap, and token/cost budget.
- Bounded retry for transient errors; never retry an invalid response indefinitely.
- Circuit breaker and advisory-only fallback.
- Dashboards by model, prompt version, CWE, and project.
- Alerts when disagreement, exemption, or parse-failure rate deviates from baseline.

## Non-goals

- No auto-tuning of thresholds from live metrics (that is governed by P2.3/P2.4).

## Acceptance criteria

- [ ] No scan exceeds the configured budget.
- [ ] A provider outage never results in a gate-exempted finding (fail safe to gating).
- [ ] Dashboards and alerts exist for the key safety metrics.

## References

- Fail-safe behaviour is the P0 principle; this issue makes it observable and enforced under load.

Issue #390, opened 2026-08-05
feat(ai-triage): blind, provider-independent verifier (P1.3)


nghiadaulau:
Parent: #387 (AI-assisted false-positive triage).

## Goal

Reduce correlated error and anchoring between the proposer and the verifier so the two-model consensus is meaningful.

## Scope

- The verifier assesses the finding independently before it sees the proposer's verdict.
- The coordinator compares the two results only after both complete.
- Support a separate base URL, credential, and provider for the verifier.
- Record provider and model-family metadata.
- Policy may require a different provider or a different model family.
- Keep backward-compatible advisory mode when the independence policy is not met.

## Non-goals

- No change to the human-review floor (a High/Critical/secret/protected-CWE finding is not eligible regardless of independence).

## Acceptance criteria

- [ ] The blind verifier prompt does not contain the proposer verdict.
- [ ] Same provider/family is not treated as independent when policy requires independence.
- [ ] Missing independence metadata keeps the finding gating.
- [ ] Tests for anchoring, correlated failure, and provider outage.

## References

- Depends on #384: consensus requires the verifier to decode deterministically (temperature 0). Fix #384 first or the blind-verifier verdict is non-reproducible.
- Consensus semantics per P0.2 and ADR-0021 (distinct verifier, sealed verdict at or above 75).

Issue #387, opened 2026-08-05
Epic: AI-assisted false-positive triage, safe and auditable


nghiadaulau:
## Goal

Use AI to cut false-positive noise in static-analysis findings without ever letting a single model silently hide a real defect. AI only proposes an opinion; authority to change a quality gate stays with deterministic server-side policy, backed by an independent second verifier, a human-review floor, tamper-evident evidence, and full observability.

This epic applies the existing Judgment primitive (ADR-0021: propose -> verify -> confirm, a gated capability promotes only on a distinct verifier's sealed verdict at or above `verdict.EvidenceThreshold` = 75) to the specific problem of false-positive triage. It does not introduce a second authority model.

## Problem

Static analysis produces false positives, which erodes trust and trains engineers to ignore warnings. AI can help triage them, but it also creates a serious risk: a hallucination, or a correlated error across two models, could remove a real vulnerability from the quality gate. The system must reduce noise AND keep every hard safety guarantee.

## Product principles (non-negotiable)

- Retain and mark. A finding is never deleted from JSON, SARIF, compliance reports, or evidence; it is only annotated with a state. This is the same retain-and-mark rule already used for suppression/VEX.
- Advisory by default. AI has no gate-exemption authority unless deterministic policy grants it.
- Fail safe. Timeout, model error, invalid response, or a missing verifier all keep the finding in its gating state.
- Separation of duties. The proposer may not verify its own conclusion.
- Deterministic authorization. Server-side policy decides gate exemption; a `GateExempt` flag from AI output or from stored data is never trusted directly and is always re-validated.
- Human-review floor. High/Critical severity, secret findings, and dangerous CWE classes are never auto-exempted by AI.
- Auditability. Every gate-affecting decision carries model, confidence, prompt version, policy version, policy reason, and an evidence reference.

## Layers

1. P0 - Safety foundation: separate AI opinion from gate authority, require distinct-verifier consensus, protect high-risk findings, enforce policy server-side, and record audit evidence.
2. P1 - Production readiness: measure quality, shadow rollout, human-review workflow, blind independent verifier, and fail-closed evidence.
3. P2 - Accuracy and scale: richer context, safe caching, a reviewer feedback loop, and model/prompt drift control.

## P0 status - substantially landed in main

[...]

PR #397, opened 2026-08-06
fix(llm): send an explicit temperature 0

### LINKED ISSUE #384: fix(llm): an explicit temperature 0 is never sent — FP triage and the judgment verifier sample at the provider default
**Describe the bug**

`ports.ChatRequest.Temperature` is a plain `float64`, and the OpenAI adapter only forwards it when it is
strictly positive:

```go
// internal/infrastructure/llm/openai/client.go:129
if req.Temperature > 0 {
    t := req.Temperature
    body.Temperature = &t
}
```

Every production call site sets `Temperature: 0` on purpose, to get a deterministic adjudication:

| Call site | Role |
| --- | --- |
| `internal/usecase/fptriage/coordinator.go:169` | FP-triage proposer |
| `internal/usecase/fptriage/coordinator.go:208` | the **distinct verifier** of the FP gate |
| `internal/usecase/llmverifier/assess.go:76` | the verifier that seals a gated judgment verdict |

Because `0` is not `> 0`, the `temperature` field is omitted from the request body entirely, so the
provider applies its **default sampling temperature** (1.0 on OpenAI-compatible APIs) instead of the
greedy decoding the call sites asked for.

These are exactly the two paths where a model verdict has authority: a `refuted` critique at/above
`verdict.EvidenceThreshold` (75) exempts a finding from the `--fail-on` gate, and `llmverifier` seals a
verdict through the same gate a human reviewer uses.

**To reproduce**

Hermetic — no gateway needed. Add to `internal/infrastructure/llm/openai/client_test.go`:

```go
func TestTemperatureZeroIsSent(t *testing.T) {
	var body map[string]any
	srv := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
		b, _ := io.ReadAll(r.Body)
		_ = json.Unmarshal(b, &body)
		_, _ = w.Write([]byte(`{"choices":[{"message":{"role":"assistant","content":"{}"},"finish_reason":"stop"}],"usage":{}}`))
	}))
	defer srv.Close()

	c, _ := New(srv.URL, "", "m", time.Second)
	_, _ = c.Chat(context.Background(), ports.ChatRequest{
		Temperature: 0, // what fptriage and llmverifier pass
		Messages:    []agent.Message{{Role: agent.RoleUser, Content: "hi"}},
	})

	if _, ok := body["temperature"]; !ok {
		t.Fatal("temperature was omitted — the provider default applies")
	}
}
```

`go test ./internal/infrastructure/llm/openai/ -run TestTemperatureZeroIsSent` fails on `main`.

**Expected behavior**

A caller that explicitly asks for temperature 0 gets temperature 0 on the wire, so the same finding
critiqued twice yields the same verdict and the two-model consensus is reproducible.

**Environment**
- Synapse version / commit: `main` @ `27311f7`
- OS + architecture: n/a (source-level; reproduced with `httptest`, no live gateway)
- How you run it: `go test`

**Logs / output**

Request body captured by the test above (note the absent `temperature`):

```json
{"model":"m","messages":[{"role":"user","content":"hi"}]}
```

**Suggested fix**

Make the zero value meaningful — `Temperature *float64` in `ports.ChatRequest` (`nil` = provider default,
non-nil = sent verbatim), a `ports.Temp(0)` helper for the three deterministic call sites, and the adapter
guard becomes `req.Temperature != nil`. The agent orchestrator (`orchestrator.go:297`) would keep its
current behaviour by passing `nil` unless `Config.Temperature > 0`, so the diff stays limited to the paths
that meant to be deterministic. Happy to send that PR with tests if you'd like it — or to leave it to you
if the current behaviour is intentional.

thx2an:
Fixes #384.

## The bug

`ports.ChatRequest.Temperature` was a plain `float64`, and the adapter only marshalled the field when it was greater than zero:

```go
if req.Temperature > 0 {   // client.go:129
```

The wire field is `*float64` with `omitempty` (`client.go:69`), so an explicit `0` — the value a deterministic caller asks for — was indistinguishable from "unset" and never reached the provider. The provider default applied instead.

Three call sites with gate authority were affected, and all three explicitly ask for `0`:

| Call site | Role |
| --- | --- |
| `internal/usecase/fptriage/coordinator.go:169` | FP-triage proposer |
| `internal/usecase/fptriage/coordinator.go:208` | the distinct verifier |
| `internal/usecase/llmverifier/assess.go:76` | the sealing judgment verifier |

Each adjudicates against `verdict.EvidenceThreshold`, so at the provider default the same finding could land on either side of the bar across two runs: a `--fail-on` gate exemption and a sealed verdict were not reproducible. That reproducibility is the premise the propose → verify → confirm design rests on, and there was no error and no log line when it did not hold.

## The change

`ChatRequest.Temperature` becomes a `*float64`: `nil` leaves the choice to the provider, a non-nil value is sent verbatim, and `0` now reaches the wire as greedy decoding. A `ports.Temp(v)` helper keeps the three deterministic call sites to one line each. Changing the port type rather than special-casing the adapter means any future adapter behind `ports.LLM` surfaces at compile time; today there is only one (`internal/infrastructure/llm/openai`).

**The agent orchestrator is deliberately unchanged.** No composition root in `cmd/` sets `orchestrator.Config.Temperature`, so it is `0` in production today and no temperature is sent. It keeps exactly that behaviour — it now builds a pointer only when a run configured a positive temperature. This PR is only "the three deterministic call sites now really send 0".

## Tests

- `client_test.go` pins the wire contract in both directions: an explicit `0` is **present** in the body with value `0`, and an unset temperature stays **absent** (so a caller expressing no preference is not silently forced onto `0`).
- `fptriage/coordinator_test.go` asserts both the proposer and the distinct verifier request temperature `0`.
- `llmverifier/coordinator_test.go` asserts the same for the sealing verifier.

The first test fails on `main` (`temperature omitted from the request – the provider default applies`) and passes here.

Raw bodies captured against a stub gateway:

```
explicit zero  -> {"model":"m1","messages":[...],"temperature":0}
explicit 0.7   -> {"model":"m1","messages":[...],"temperature":0.7}
unset (nil)    -> {"model":"m1","messages":[...]}
```

## Verification

`gofmt -l .` clean, `go build ./...`, `go vet ./...`, and `go test ./...` pass (Go 1.26, `CGO_ENABLED=0`).

Two notes in the interest of full disclosure: `internal/infrastructure/toolrunner`'s `TestRunTimeoutKillsProcessGroup` fails in my containerised toolchain, but it fails identically on an unmodified `main` — a PID-namespace artifact of my environment, unrelated to this change. And I did not run `cd web && pnpm build`: this PR touches no file under `web/`, and I do not have pnpm locally.

## On the review notes in #384

1. **Regression test** — included, as described above.
2. **`seed` / `top_p`** — agreed that temperature 0 is necessary but not always sufficient for bit-exact reproducibility. Left out of this PR to keep it single-purpose; happy to open a follow-up if you want it pursued.
3. **Other fields with the same zero-value-omitted trap** — I audited every `ChatRequest` field. `Temperature` was the only one. `Model == ""` is a documented, deliberate fall-back to the client's default model. There is no `TopP` or `seed` field yet — the `*float64` pattern is the guard when one is added. `MaxTokens` has a *different* problem: it is set at four call sites but the adapter never sends it at all, by an explicit decision recorded in the NOTE at `client.go:133` (the field name differs across model generations). That is not a zero-value trap and it touches a choice you made on purpose, so I have not changed it here — I will raise it as its own issue, linked to #392, unless you would rather it stayed as documented.

Related: #387 lists #384 as a blocking dependency for the P0.2 distinct-verifier consensus and the P1.3 blind verifier (#390).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -47,6 +47,15 @@ capabilities below are already shipped on `main`.
 
 ### Fixed
 
+- **Deterministic AI adjudication.** An explicit `temperature: 0` never reached the provider: the
+  OpenAI-compatible adapter only sent the field when it was greater than zero, so the value a
+  deterministic caller asks for was indistinguishable from "unset" and was dropped from the request. The
+  AI false-positive proposer, its distinct verifier, and the automated judgment verifier all request 0 and
+  were therefore adjudicating at the provider's default sampling — the same finding could land on either
+  side of the evidence threshold across two runs, so a gate exemption and a sealed verdict were not
+  reproducible. `ChatRequest.Temperature` is now a `*float64` (nil leaves the provider default; a non-nil
+  value, including 0, is sent verbatim), so those three call sites decode greedily as intended. The agent
+  orchestrator is unchanged: it still leaves sampling to the provider unless a run configures one.
 - **Config docs.** `docs/guide/configuration.md` listed the analysis-brain flags (judgments, SAST,
   reachability, secret and misconfig scanning, cross-check, compliance, scan cache, image rootfs, owned
   advisory, gomodgraph) as default `false`; they ship `true`. Corrected the defaults to match the code.
--- a/internal/infrastructure/llm/openai/client.go
+++ b/internal/infrastructure/llm/openai/client.go
@@ -126,8 +126,8 @@ func (c *Client) Chat(ctx context.Context, req ports.ChatRequest) (ports.ChatRes
 		// 400; the orchestrator may retry without the schema (tool-calling still works).
 		body.ResponseFormat = &wireRespFmt{Type: "json_schema", JSONSchema: req.ResponseSchema}
 	}
-	if req.Temperature > 0 {
-		t := req.Temperature
+	if req.Temperature != nil {
+		t := *req.Temperature
 		body.Temperature = &t
 	}
 	// NOTE: no max_tokens/max_completion_tokens – the field name differs across OpenAI model
--- a/internal/infrastructure/llm/openai/client_test.go
+++ b/internal/infrastructure/llm/openai/client_test.go
@@ -97,6 +97,57 @@ func TestChatTerminalOnBadRequest(t *testing.T) {
 	}
 }
 
+// TestChatSendsExplicitTemperatureZero pins the wire contract deterministic callers depend on: an
+// explicit temperature 0 (greedy decoding) must reach the provider. Dropping it silently falls back to
+// the provider default (~1.0), which is the difference between a reproducible verdict and a sampled one
+// — the FP-triage proposer/verifier and the judgment verifier all ask for 0.
+func TestChatSendsExplicitTemperatureZero(t *testing.T) {
+	body := captureChatBody(t, ports.ChatRequest{
+		Temperature: ports.Temp(0),
+		Messages:    []agent.Message{{Role: agent.RoleUser, Content: "hi"}},
+	})
+	got, ok := body["temperature"]
+	if !ok {
+		t.Fatal("temperature omitted from the request – the provider default applies")
+	}
+	if got != float64(0) {
+		t.Errorf("temperature = %v, want 0", got)
+	}
+}
+
+// TestChatOmitsTemperatureWhenUnset is the other half of that contract: a caller expressing no
+// preference must not be forced onto 0 – the provider default still applies.
+func TestChatOmitsTemperatureWhenUnset(t *testing.T) {
+	body := captureChatBody(t, ports.ChatRequest{
+		Messages: []agent.Message{{Role: agent.RoleUser, Content: "hi"}},
+	})
+	if v, ok := body["temperature"]; ok {
+		t.Errorf("temperature sent (%v) although the caller expressed no preference", v)
+	}
+}
+
+// captureChatBody runs one Chat against a stub gateway and returns the decoded request body.
+func captureChatBody(t *testing.T, req ports.ChatRequest) map[string]any {
+	t.Helper()
+	var body map[string]any
+	srv := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		b, _ := io.ReadAll(r.Body)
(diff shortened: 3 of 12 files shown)


nghiadaulau (approved):
## Review — APPROVE ✅

This is exactly the fix I recommended on #384, and it is verified end to end on the merged tree (pr-397 + latest main, no conflicts).

**The bug is actually fixed, both directions.** I re-ran the hermetic repro from #384 (which FAILS on main) plus a nil case:
```
temperature on wire = 0 (greedy)        <- was omitted entirely before
nil Temperature     -> omitted          <- provider default still preserved
```
So `Temp(0)` reaches the wire as greedy decoding, and `nil` remains "leave it to the provider". That is the distinction the old `float64` could not express.

**Shape of the change is right**
- `ports.ChatRequest.Temperature` becomes `*float64` with a `ports.Temp(v)` helper; the adapter guard is now `req.Temperature != nil` (`openai/client.go`). Changing the port type rather than special-casing the adapter is the correct call: any future `ports.LLM` adapter surfaces at compile time.
- All three gate-authority call sites now pass `ports.Temp(0)` with a comment stating why (`fptriage/coordinator.go:169` proposer, `:208` distinct verifier, `llmverifier/assess.go:76` sealing verifier).
- The agent orchestrator behaviour is genuinely preserved: it builds a pointer only when `cfg.Temperature > 0`, so the production path (no composition root sets it) still sends no temperature. Scope stayed limited to the paths that meant to be deterministic, as promised.

**Gates (all green)**
- `go vet` — clean
- tests: `llm/openai` (incl. the new `client_test.go` assertions), `fptriage`, `llmverifier`, `orchestrator` — PASS
- `golangci-lint` over the affected packages — **0 issues**

Regression coverage is present at the adapter and at both consensus call sites, which was my one requirement on #384. This unblocks the reproducible two-model consensus that #398 / epic #387 (P0.2) and #390 (P1.3 blind verifier) depend on. Merging.

Issue #389, opened 2026-08-05
feat(ai-triage): human-review API and UI (P1.2)


nghiadaulau:
Parent: #387 (AI-assisted false-positive triage).

## Goal

Turn `review_required` from metadata into a real workflow with an owner and an audit trail.

## Scope

- Add the AI-triage DTO to the frontend `ScanResult` and the API mapper.
- Show proposer/verifier verdict, confidence, model, and policy reason.
- Distinct badges for suspected, verified, gate-exempt, and review-required.
- A review queue filterable by severity, CWE, project, and state.
- Reviewer Accept/Reject with a mandatory rationale.
- Persist actor, timestamp, policy version, and evidence reference.
- No AI or machine role may approve its own decision.

## Non-goals

- No change to the deterministic gate policy itself (that is P0 / policy).
- The UI must never hide gate-exempt findings from the list or reports.

## Acceptance criteria

- [ ] Every `review_required` finding appears in the review queue.
- [ ] Only an authorized role can decide.
- [ ] Every decision is audited and linked to evidence.
- [ ] A reviewer rejection returns the finding to gating.
- [ ] The UI does not hide gate-exempt findings from lists or reports.

## References

- Frontend follows the design-system tokens and TanStack Query conventions (docs/04-ui-ux.md).
- Separation of duties and audit are per ADR-0021 and the append-only audit log.

PR #386, opened 2026-08-04
fix: harden secret scans and evidence sealing

pho-veteran:
## Summary

Hardens secret scanning over untrusted workspaces and makes configured SCA evidence sealing fail closed.

Closes #385.

## Changes

- Add aggregate scanned-byte and finding budgets with an explicit incompleteness signal surfaced through SCA source warnings.
- Read files through `os.Root`, verify opened file identity and stability, reject symlink/TOCTOU swaps, and recheck size after open.
- Detect VB secret declarations including typed `Dim`/`Const`, camelCase and bracketed identifiers, type characters, escaped strings, and comment masking.
- Propagate configured evidence marshal/vault/seal failures before scan, finding, result, run, or imported-SBOM persistence.
- Merge complementary cached scan state and recompute derived fields before sealing the final result.
- Add regression coverage for limits, truncation, root-confined reads, VB syntax, cached-result consistency, and evidence failure ordering.

## Checklist

- [x] `make build vet test typecheck` passes
- [x] `cd web && pnpm build` passes (dashboard unchanged; build verified)
- [x] Tests added/updated for new behavior
- [x] Preserves the safety invariants (argv exec, server-side scope enforcement, secrets out of logs, deterministic reports, append-only evidence) – see CONTRIBUTING.md
- [x] Documentation updated if needed (no documentation change required)

PR #398, opened 2026-08-06
feat(ai-triage): enforce P0 gate-exemption policy

H1eu232:
## Summary

Implements the P0 safety foundation from #387 on top of current main, including the fail-closed evidence changes merged in #386.

This is a fresh replacement for the closed, unmerged draft #376. It is based on main at b47f6d0 rather than reopening the stale branch.

Part of #387.
#384 was resolved by #397; this branch is rebased on the fix.

## Safety contract

- A single-model false-positive opinion is advisory-only and cannot affect a quality gate.
- A gate exemption requires two canonically distinct models to independently return refuted at or above verdict.EvidenceThreshold.
- High/Critical findings, secret findings, and protected CWE classes always remain gating for human review.
- Missing, failed, unavailable, or non-distinct verifier paths fail safe.
- Findings remain present in JSON, SARIF, compliance output, and evidence.
- Server-side readers revalidate the complete consensus and finding risk floor instead of trusting persisted GateExempt metadata.

## Implementation

- Adds versioned authorization policy fp-gate-v2 and prompt contract fp-triage-v1.
- Separates suspected_fp, verified, gate_exempt, and review_required.
- Records proposer/verifier identity, verdict, confidence, prompt version, policy version, and policy reason.
- Uses AIGateExemptKeys as the only AI authority consumed by CLI and project quality gates.
- Seals all gate-affecting AI metadata and policy inputs (severity, CWE, kind, scope, class) in canonical scan evidence.
- Preserves #386: a configured evidence-ledger failure returns an error and prevents result persistence.
- Adds adversarial tests for forged flags, same-model verification, low confidence, verifier disagreement, protected risk floors, missing findings, and evidence content/order.

## P0 policy decisions for PM/Security sign-off

1. Only Low/Medium findings outside protected CWE classes may be AI gate-exempt.
2. High/Critical, secrets, and protected CWEs always require human review.
3. A missing or failing verifier leaves triage advisory-only.
4. review_required is metadata in P0; the durable API/UI workflow is P1.2 (#389).

## Security review remediation

Commit aaa5966 addresses the requested changes:

- Blocker 1: protects CWE-798, CWE-259, CWE-321, and CWE-522; secret and credential-bearing findings are excluded from LLM candidate selection.
- Blocker 2: the policy now requires an evidence ledger before setting gate_exempt. The standalone CLI has no vault, so it remains advisory-only and the misleading evidence-sealed message was removed.
- Blocker 3: proposer/verifier identities are canonicalized across case, provider/router prefixes, and dated aliases in both coordinator attachment and policy revalidation.
- Severity overlays are threaded through the gate re-check.
- Empty-CWE SAST/misconfig findings default to human review.
- Kind, first-party class, and exact production scope are revalidated.
- Gate-floor inputs are sealed and evidence ordering uses stable sorts.
- SARIF marking is tracked separately in #430.

Regression tests cover every bypass and negative path above.
## Verification

Passed locally on current main:

- go test ./internal/usecase/fptriage ./internal/usecase/sca
- go test ./internal/usecase/...
- go vet ./internal/usecase/fptriage ./internal/usecase/sca ./internal/usecase/projectuc
- frontend pnpm test: 29 files / 267 tests passed
- frontend pnpm run typecheck
- frontend pnpm build
- git diff origin/main...HEAD --check

The full cmd/synapse-cli and cmd/synapse-api package test was blocked locally by Windows Defender quarantining the upstream file internal/infrastructure/tools/sast/php_lex_test.go as potentially unwanted software. That file is unchanged by this PR and was restored before push; CI should provide the authoritative full-suite result.

## Rollout

The feature remains off by default. Until PM/Security approve the P0 contract, deployments should keep AI triage disabled or advisory-only.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -175,12 +175,13 @@ SYNAPSE_SANDBOX_ENABLED=false
 # gate a human reviewer uses (verifier identity "llm:<model>", never the proposer → no self-confirm). It
 # also enables the FP-triage two-model consensus. Defaults to SYNAPSE_LLM_MODEL (= no distinct verifier).
 # SYNAPSE_VERIFIER_MODEL=
-# AI false-positive analysis: after a scan, the proposer critiques production-scope first-party findings.
+# AI false-positive analysis: after a scan, the proposer critiques production-scope first-party
+# SAST/misconfig findings. Secret findings and their raw source context never enter the LLM transcript.
 # A high-confidence "refuted" verdict is advisory and never deletes a finding. Gate exemption requires a
 # DIFFERENT SYNAPSE_VERIFIER_MODEL to independently refute at/above the bar. Even then, high/critical,
 # secret, and dangerous-CWE findings stay gating for human review. The typed proposer/verifier verdicts,
 # model IDs, prompt version, policy decision, and gate authorization are sealed with the scan evidence.
-# Without a distinct verifier the feature remains useful for triage, but cannot change a gate.
+# Without a distinct verifier OR a configured evidence ledger the feature remains advisory-only.
 # SYNAPSE_FP_TRIAGE_ENABLED=false
 # SYNAPSE_FP_TRIAGE_MODEL=
 # HITL approval: manual (human approves every action) | filter | auto.
--- a/.env.example
+++ b/.env.example
@@ -175,12 +175,12 @@ SYNAPSE_SANDBOX_ENABLED=false
 # gate a human reviewer uses (verifier identity "llm:<model>", never the proposer → no self-confirm). It
 # also enables the FP-triage two-model consensus. Defaults to SYNAPSE_LLM_MODEL (= no distinct verifier).
 # SYNAPSE_VERIFIER_MODEL=
-# AI false-positive gate: after a scan, the model critiques the remaining production-scope first-party
-# source findings (SAST/secret/misconfig). A high-confidence "refuted" verdict marks the finding a
-# suspected false positive — retain-and-mark (still reported + sealed, held back from the --fail-on gate),
-# never deleted. Off by default; needs an LLM endpoint above. Model defaults to SYNAPSE_LLM_MODEL.
-# When SYNAPSE_VERIFIER_MODEL (above) names a DIFFERENT model, a refutation must be independently
-# confirmed by it before it exempts the gate (two-model consensus; a single model can't flip the gate).
+# AI false-positive analysis: after a scan, the proposer critiques production-scope first-party findings.
+# A high-confidence "refuted" verdict is advisory and never deletes a finding. Gate exemption requires a
+# DIFFERENT SYNAPSE_VERIFIER_MODEL to independently refute at/above the bar. Even then, high/critical,
+# secret, and dangerous-CWE findings stay gating for human review. The typed proposer/verifier verdicts,
+# model IDs, prompt version, policy decision, and gate authorization are sealed with the scan evidence.
+# Without a distinct verifier the feature remains useful for triage, but cannot change a gate.
 # SYNAPSE_FP_TRIAGE_ENABLED=false
 # SYNAPSE_FP_TRIAGE_MODEL=
 # HITL approval: manual (human approves every action) | filter | auto.
--- a/cmd/synapse-api/main.go
+++ b/cmd/synapse-api/main.go
@@ -703,7 +703,11 @@ func main() {
 			if cfg.VerifierModel != "" && cfg.VerifierModel != cfg.FPTriageModel {
 				if vllm, verr := openai.New(cfg.LLMBaseURL, cfg.LLMAPIKey, cfg.VerifierModel, cfg.LLMTimeout); verr == nil {
 					coord.WithVerifier(vllm, cfg.VerifierModel)
-					mode = "verified by " + cfg.VerifierModel
+					if coord.VerifierModel() != "" {
+						mode = "verified by " + coord.VerifierModel()
+					} else {
+						log.Warn("AI FP-triage verifier aliases the proposer after canonicalization; triage remains advisory-only", "verifier_model", cfg.VerifierModel)
+					}
 				} else {
 					log.Warn("AI FP-triage verifier unavailable; triage remains advisory-only", "err", verr)
 				}
--- a/cmd/synapse-api/main.go
+++ b/cmd/synapse-api/main.go
@@ -692,27 +692,26 @@ func main() {
 		log.Info("misconfig scanning ENABLED (Dockerfile + Kubernetes + Terraform); " + helmMode)
 	}
 	// AI false-positive triage in the scan pipeline (opt-in, best-effort, PROPOSE-ONLY). Independent of
-	// the agent: it critiques production-scope source findings and marks suspected FPs retain-and-mark
-	// (held back from the gate via ScanResult.SuspectedFPKeys, still reported + sealed). A distinct
-	// SYNAPSE_VERIFIER_MODEL enables two-model consensus.
+	// the agent: it critiques production-scope source findings. Single-model output is advisory-only; a
+	// distinct verifier is required before the deterministic high-risk floor may grant a gate exemption.
 	if cfg.FPTriageEnabled && strings.TrimSpace(cfg.FPTriageModel) != "" {
 		if tllm, terr := openai.New(cfg.LLMBaseURL, cfg.LLMAPIKey, cfg.FPTriageModel, cfg.LLMTimeout); terr != nil {
 			log.Warn("AI false-positive triage DISABLED (LLM unavailable)", "err", terr)
 		} else {
 			coord := fptriage.New(tllm, cfg.FPTriageModel)
-			mode := "single-model"
+			mode := "advisory-only (distinct verifier required for gate exemption)"
 			if cfg.VerifierModel != "" && cfg.VerifierModel != cfg.FPTriageModel {
 				if vllm, verr := openai.New(cfg.LLMBaseURL, cfg.LLMAPIKey, cfg.VerifierModel, cfg.LLMTimeout); verr == nil {
 					coord.WithVerifier(vllm, cfg.VerifierModel)
 					mode = "verified by " + cfg.VerifierModel
 				} else {
-					log.Warn("AI FP-triage verifier unavailable, single-model", "err", verr)
+					log.Warn("AI FP-triage verifier unavailable; triage remains advisory-only", "err", verr)
 				}
 			}
 			scaService.SetFPTriage(fptriage.NewTriager(coord, func(root string) ports.SourceSnippetReader {
 				return sourcesnippet.Reader{Root: root}
 			}))
-			log.Info("AI false-positive triage ENABLED ("+mode+"); suspected FPs held back from the gate, still reported", "model", cfg.FPTriageModel)
+			log.Info("AI false-positive triage ENABLED ("+mode+"); only verified low-risk consensus can affect gates", "model", cfg.FPTriageModel)
 		}
 	}
 	if cfg.SuppressionEnabled {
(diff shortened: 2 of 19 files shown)


nghiadaulau (changes requested):
## Review — REQUEST CHANGES (three verified bypasses in the gate-authority path)

Reviewed on the merged tree (pr-398 + latest main incl. #397, merges clean) plus an adversarial security audit. The design is genuinely good and most of the safety contract holds, but three defects let the primary promise ("AI alone cannot hide a real finding from the gate") be broken in practice. Given this is the one code path that removes findings from a CI gate, these are blocking.

### Gates (all green)
- `go vet` — clean; tests `fptriage` / `sca` / `config` / `cmd` — PASS; `golangci-lint` — **0 issues**
- Off by default confirmed: `SYNAPSE_FP_TRIAGE_ENABLED` defaults `false` (`config.go:458`), and with one model configured `FPTriageModel == VerifierModel` so no verifier attaches. Reaching an exemption needs two deliberate config choices. Good.

### What is right (verified, credit where due)
- **Single-model cannot exempt.** `applyAIGatePolicy` zeroes `GateExempt`/`ReviewRequired` first (`fptriage_policy.go:65-66`), then requires `SuspectedFP` -> `hasVerifiedConsensus` -> finding present -> risk floor clear. Single-model mode has empty `VerifierModel`/`VerifierVerdict`, so it can never pass.
- **Full-DTO revalidation, not a trusted boolean.** `hasVerifiedConsensus` re-checks both verdicts and both confidences against `verdict.EvidenceThreshold` instead of trusting `Verified`, so a lowered `WithMinConfidence` cannot buy authority.
- **Forged persisted flag is caught.** `AIGateExemptKeys` (`service.go:596-617`) requires `GateExempt` AND policy-version pin AND `PolicyReason == verified_consensus` AND a full recompute of `hasVerifiedConsensus` AND finding present AND `humanReviewFloor()==""`. Both consumers use only this set (CLI `main.go:1373`, project gate via `GateExemptKeys`); no reader touches the raw flag.
- **Verifier error is not agreement.** `coordinator.go:201-206` leaves `Verifier` nil on error, which also drops `SuspectedFP`. Fail-safe in the right direction.
- **CWE matching is exact-token, not substring** (`FieldsFunc` + map lookup), so `CWE-890` does not collide with `CWE-89`. Nice detail.
- **Retain and mark holds.** Exempted findings still appear in JSON, SARIF, compliance and ratings; the gate loop only skips the counter.

### Blocker 1 — hardcoded-credential findings are AI-exemptible, so "secrets always require human review" is not true

`humanReviewFloor` keys the secret floor on `item.Kind == finding.KindSecret` (`fptriage_policy.go:109`), and **`CWE-798` is absent from `humanReviewCWEs`** (verified: zero occurrences in the file). Pattern-SAST credential rules land at `Kind=sast`, not `KindSecret`, and several ship at **medium** severity with CWE-798:

- `rulecatalog/langpacks.go:3289` `cs-x509-password-literal` — SeverityMedium, CWE-798
- `langpacks.go:3991` `cs-basic-auth-literal` — SeverityMedium, CWE-798
- `langpacks.go:5377` `cs-connstring-password-literal` — SeverityMedium, CWE-798

Failure scenario: a PR adds a real password in a C# connection string with a comment like `// placeholder from the fixture generator, not a live credential`. The finding is `Kind=sast`, `Severity=medium`, `CWE=CWE-798`. Both models read that same comment and refute at >= 75, `humanReviewFloor` returns `""`, so `GateExempt=true` and `synapse-cli scan --fail-on medium` **passes with a live credential in the tree**.

Fix: add `CWE-798`, `CWE-259`, `CWE-321`, `CWE-522` to `humanReviewCWEs`, and/or key the secret floor on the rule's security-credential tag rather than `Kind` alone.

### Blocker 2 — the CLI seals nothing, while this PR newly tells the operator it did

`sealEvidence` returns immediately when no vault is configured (`service.go:3087-3089`), and `synapse-cli` constructs the service with `ev == nil` (7th positional in `NewService`; `cmd/synapse-cli/main.go:1168` passes `nil` there). So on the CI path — the only path where an exemption changes a build outcome — there is **no tamper-evident record** of which findings the AI exempted, by which models, at what confidence, and no audit entry either.

This PR changes the message to claim the opposite (`cmd/synapse-cli/main.go:1473`, new in this diff):
```
... held back from the gate by policy (reported + evidence-sealed; not deleted)
```
The previous wording ("reported with a verdict; not deleted") was accurate; the new one is false whenever `s.evidence == nil`.

Fix: either wire a file-backed evidence vault into `synapse-cli`, or refuse to apply `AIGateExemptKeys` when no ledger is configured (fail closed) — and until one of those lands, drop "evidence-sealed" from the message.

### Blocker 3 — the distinct-verifier check is case-sensitive, so one model can self-confirm

`hasVerifiedConsensus` compares `strings.TrimSpace(proposer) != strings.TrimSpace(verifierModel)` (`fptriage_policy.go:99`); the upstream guard has the same flaw (`coordinator.go:113` `model != c.model`), and both clients are built against the same base URL and key. I verified this empirically:

```
hasVerifiedConsensus{ProposerModel:"gpt-4", VerifierModel:"GPT-4", both refuted, both 90} -> TRUE
```

So `SYNAPSE_FP_TRIAGE_MODEL=gpt-4o` + `SYNAPSE_VERIFIER_MODEL=GPT-4o` (or `openai/gpt-4o` on a router, or `gpt-4o` vs a dated alias of the same weights) yields a "two-model consensus" from a single model — defeating the non-negotiable property. Whitespace is trimmed, so only case/prefix normalization is missing. The existing test only covers the byte-identical case (`fptriage_test.go:129`).

Fix: compare with `strings.EqualFold` after stripping a `provider/` prefix, in both places, and add the negative test.

### Should-fix (not blocking, but please address or file)
1. **Pre-overlay severity vs post-overlay gate.** `GateExemptKeys(items)` takes the overlaid findings, but `AIGateExemptKeys()` ignores `items` and evaluates the floor against `r.Findings` (`service.go:599-604`), i.e. pre-overlay. A rule a tenant deliberately escalated to high/critical in its quality profile (`qualitygate/profile.go:49-51`) is therefore still exemptible because the policy still sees `medium`. Thread `items` through.
2. **Empty CWE defaults to exemptible.** Zero tokens returns `""` (allowed). Every `Kind=misconfig` finding has no CWE at all, so the CWE floor is a no-op for the whole class. Consider protected-by-default when `Kind ∈ {sast,misconfig}` and CWE is empty.
3. **No Kind/scope re-check in the policy.** `applyAIGatePolicy` looks the critique key up in all `result.Findings`, not in `fpTriageCandidates`. The shipped `Triager` only echoes candidate keys so it is not reachable today, but the file's own stated threat model is "a buggy or alternate FPTriager", and this is the one place that model is not defended.
4. **The floor's inputs are outside the seal.** `scanEvidenceContent` seals dedup keys only (`service.go:3044-3048`); `severity`, `CWE` and `Kind` — exactly what `humanReviewFloor` reads — are not sealed, so a results-store writer could flip `CWE-89` to `""` without a chain mismatch. Seal `(dedupKey, severity, cwe, kind)` for any finding carrying an exemption.
5. `sort.Slice` -> `sort.SliceStable` at `service.go:3058` (a duplicate 3-field key would make sealed bytes nondeterministic).
6. SARIF carries no marking for an exempted finding — consider a `suppressions` entry with `justification = policy_reason` so code-scanning shows why.

### Pre-existing issue found during review (not caused by this PR — please file separately)
Raw secrets reach the LLM. `fpTriageCandidates` includes `finding.KindSecret` (`service.go:664`), and for each candidate `coordinator.go:171-174` fetches +/-8 raw source lines via `sourcesnippet` (which contains no redaction logic at all) and `prompt.go:111` embeds them as `Source context:` in the provider request. The finding's `Description` is redacted, but the snippet is not — so the actual credential is transmitted to the model provider, and again to the verifier. That conflicts with golden rule 3 (secrets never enter the LLM transcript). Confirmed present on `main` before this PR, so it is not a #398 regression, but this PR raises the stakes. Fix direction: skip snippet retrieval for `Kind=secret`, or run the secret redactor over the snippet before it leaves the process.

---

The architecture here is sound and the defensive re-validation is well built — this is close. Please address Blockers 1-3 (each is small: a CWE-list addition, an honest message or a wired vault, and `EqualFold`), add the matching negative tests, and I will re-review promptly.

H1eu232:
Addressed the security review in aaa5966 after rebasing onto current main (including #397 and #399).

Blocker mapping:
1. Credential findings: added CWE-798/259/321/522 to the human-review floor. KindSecret and credential-CWE source findings are now excluded before snippet/LLM processing.
2. CLI evidence: gate_exempt now requires an evidence ledger. Because synapse-cli has no vault, its AI triage is advisory-only; the inaccurate evidence-sealed CLI message was removed.
3. Same-model aliases: model IDs are canonicalized across case, provider/router prefixes, and dated aliases in both WithVerifier and hasVerifiedConsensus.

Also addressed should-fix items 1-5:
- revalidate against post-profile-overlay findings;
- fail closed on empty CWE for SAST/misconfig;
- re-check kind, first-party class, and exact production scope;
- seal severity/CWE/kind/scope/class with AI decisions;
- use stable evidence ordering.

The raw-secret transcript issue is also closed defensively by excluding secret and credential-CWE findings from LLM candidate selection. SARIF marking is tracked separately in #430.

Local verification:
- go test: agent, fptriage, sca, projectuc, config, OpenAI adapter, api, httpapi
- go vet: agent, fptriage, sca, projectuc, config
- git diff --check

All new bypass regression tests pass. Please re-review when convenient.

nghiadaulau (approved):
## ✅ Approve — reviewed strictly (safety-critical), mergeable

This PR makes an AI false-positive opinion able to exempt a CI quality gate, so I reviewed it against golden rule 5 / ADR-0021 with a full adversarial security + architecture pass, verified the actual code (built, vetted, `-race` tested, `golangci-lint` — clean), and confirmed the **merged result** (this branch + current `main`, which now carries #446's overlapping edits to `main.go`/`config.go`/`sca/service.go`) still builds and passes — git auto-merges those cleanly (no conflicts).

**The safety contract holds — fails closed on every path I could construct:**
- `GateExempt` starts `false`; the **only** write of `true` is the final verified-consensus branch, after every floor. Every early return (`not-suspected`/`no-consensus`/`finding-missing`/`floor`/`ineligible`/`evidence-missing`) leaves it false.
- `hasVerifiedConsensus` re-validates the whole DTO (never trusts the `Verified` bool): distinct proposer≠verifier **after `agent.CanonicalModelID`** (which really normalizes case, `provider/` prefix, and dated `-YYYY-MM-DD` aliases → self-verification via an alias is rejected), both verdicts `refuted`, both confidences `≥ verdict.EvidenceThreshold`.
- Absolute human-review floors (High/Critical, `KindSecret`, a 21-CWE protected set, unknown-CWE on SAST/misconfig) run **before** any exemption.
- Requires the evidence ledger (`evidenceAvailable`) to exempt.
- **Single server-owned chokepoint** (`applyAIGatePolicy`, one call site), and every consumer (CLI `--fail-on`, project gate) reads through `AIGateExemptKeys`, which **re-validates consensus + floors + PolicyVersion at read time** against the exact finding view — so a stale/cache-merged or tenant-severity-overlaid flag can't leak an exemption. Sealed evidence includes the floor-input finding fields (tamper-evident). SARIF still exports every finding (exemption never hides from code-scanning).
- Config fails closed: verifier defaults to the proposer model → `WithVerifier` no-ops → advisory-only → no exemption; FP-triage off by default.

**Architecture: PASS** (dependency rule clean; the decision is a typed Go state machine, LLM is propose-only and can never set `GateExempt` — golden rule 4). Fail-closed branches are thoroughly tested, including forged-consensus variants (same-model, case/provider alias, low confidence, wrong verdict).

**Non-blocking follow-ups** (none affects the safety guarantee):
1. CLI/API drift: the API warns when the verifier canonicalizes equal to the proposer; the CLI omits that operator warning (behavior is still safe). Mirror the API's `coord.VerifierModel()` check.
2. `cweTokens`: strip leading zeros (`CWE-0079`) as cheap defense-in-depth (not exploitable today — Synapse rules emit canonical ids and the LLM can't set `finding.CWE`).
3. Add an explicit "FPTriager is a trust boundary" note (a fully-malicious injected triager is out of the single-field-forgery threat model this defends).

Careful, well-tested safety work. Merging.

Issue #379, opened 2026-08-03
feat(reachability): build a static JavaScript and TypeScript module import graph


VietSory:
Parent: #378

## Objective

Introduce a deterministic, source-only module graph for first-party JavaScript and TypeScript code.

This issue establishes only the source-discovery, import-extraction and first-party module-relationship foundation. It does not resolve imports to npm components, classify dependencies as reachable or unreachable, create judgments, or affect OpenVEX output.

The graph will be consumed by later phases for package resolution and Tier-1 import reachability. Tier-2 affected-symbol analysis is not a design constraint for this PR and will remain behind a separate feasibility review.

## Architectural decisions

The following decisions are considered resolved for this phase:

* Introduce a focused `internal/domain/modulegraph` domain.
* Expose the scanner through a dedicated `ports.JSImportScanner` contract.
* Implement the scanner under `internal/infrastructure/tools/jsimports`.
* Keep the implementation CGO-free.
* Do not depend on the existing `astwalk` tree-sitter path.
* Do not introduce a language-neutral module-graph abstraction until a second concrete ecosystem requires one.
* Do not execute Node.js, TypeScript, package managers, bundlers or project scripts.
* Do not access the network.

## Parser strategy

R1 will use a purpose-built, pure-Go lexical import extractor rather than a complete JavaScript or TypeScript parser.

The extractor needs to understand enough lexical structure to distinguish real module-loading syntax from equivalent text appearing inside comments, strings, template-literal text, regular-expression literals or JSX text. It is not expected to validate the complete JavaScript or TypeScript grammar.

When the extractor encounters syntax or loading behavior it cannot analyze safely, it must record a structured coverage limitation. It must never silently skip the construct or imply that the resulting graph is complete.

Adding a third-party parser dependency or reusing the CGO tree-sitter path requires separate maintainer approval.

## Proposed package layout

```text
internal/domain/modulegraph/
    graph.go
    graph_test.go

internal/infrastructure/tools/jsimports/
    scanner.go
    lexer.go
    extractor.go
    resolver.go
    scanner_test.go
    lexer_test.go
    testdata/
```

The exact filenames may change during implementation, but domain types must remain independent of filesystem and parser implementation details.

## Domain model

The graph should preserve module identity, source location, import semantics and coverage limitations.

An illustrative model is:

```go
type Dialect string

const (
	DialectJavaScript Dialect = "javascript"
	DialectJSX        Dialect = "jsx"
[...]

Issue #378, opened 2026-08-03
Epic: deterministic JavaScript and TypeScript dependency reachability


VietSory:
## Proposal and intended ownership

This epic proposes a conservative, incremental implementation of JavaScript and TypeScript dependency reachability for Synapse.

I intend to contribute the work through a sequence of focused child issues, beginning with the static module-import graph in #379. Each phase is designed to remain independently reviewable and testable, while later phases can be refined through maintainer feedback without invalidating the earlier foundations.

The implementation should extend Synapse’s existing reachability and judgment architecture rather than introduce a parallel analysis or result model.

## Background

Synapse currently provides deterministic Tier-2 call-path reachability for Go and JVM projects, along with Tier-1 import reachability for Python.

JavaScript and TypeScript dependencies do not yet receive equivalent first-party usage context. As a result, npm-family findings currently retain two important sources of uncertainty:

1. A package may be present in a lockfile or SBOM even though it is never imported by first-party code.
2. A package may be imported, while the specific exported symbol affected by a vulnerability is never reached from a first-party entrypoint.

This epic proposes addressing those cases through two progressively stronger levels of evidence.

### Tier 1: package import reachability

Determine whether first-party JavaScript or TypeScript source code statically imports a resolved third-party npm package.

This is weaker than a call-path proof, but it provides deterministic evidence that a dependency is either used by first-party source or absent from the statically observable import graph.

### Tier 2: affected-symbol call-path reachability

Determine whether a supported first-party entrypoint reaches a vulnerable imported symbol, and preserve an explainable path from the entrypoint to that symbol.

Tier 2 should build on the Tier-1 graph and resolution foundations rather than replace them.

## Alignment with the existing architecture

The implementation should reuse the reachability contracts and governance paths that already exist in Synapse.

In particular:

* `internal/usecase/reachability.Analysis` and `reachability.Result` should remain the canonical representation for deterministic analysis outcomes and proof paths.
* `internal/usecase/reachproof.Coordinator` should continue to own audited, append-only reachability judgments and tier-aware supersession.
* The Python Tier-1 analyzer’s fail-safe behavior should remain the model for unsupported or dynamic language behavior: incomplete coverage must produce no conclusion rather than a false `not reachable` result.
* Existing SCA finding subjects, affected-symbol data, OpenVEX generation, evidence handling and reporting paths should be extended rather than duplicated.

This epic must not introduce a second reachability status model or bypass the existing propose → verify judgment workflow.

[...]

PR #399, opened 2026-08-06
feat(modulegraph): add Phase 5A domain types and JSImportScanner port

VietSory:
## Summary
Adds Phase 5A domain types (\internal/domain/modulegraph\) and driven scanner port (\internal/usecase/ports.JSImportScanner\) for JS/TS module import graph (Issue #379, Epic #378).

### Scope (5-file allowlist)
- \internal/domain/modulegraph/modulegraph.go\
- \internal/domain/modulegraph/normalize.go\
- \internal/domain/modulegraph/normalize_test.go\
- \internal/usecase/ports/jsimports.go\
- \internal/usecase/ports/jsimports_test.go\

### Status
**PARTIAL / STOPPED**
- Focused Phase 5A validation: **PASS** (100% 5/5 focused checks).
- Full repository validation: **INCOMPLETE**.
  - \make build\ and \make vet\: **PASS**.
  - \make test\: **STOPPED** at existing \internal/infrastructure/sandbox\ rootfs integration test.
  - \make typecheck\: **PASS** (run separately after \
un-all.sh\ stopped).

### Baseline & Failure Analysis
- **Starting commit**: \5c3f4a8418202e3eef98d16ec8fc251807133b2f\
- **Ending commit**: \5c3f4a8418202e3eef98d16ec8fc251807133b2f\ (commit \dbb57bf\)
- **Failure Class**: \FULL_REPOSITORY_FAILURE_UNRELATED\
- **Baseline Reproduction Proof**: Tested \go test ./internal/infrastructure/sandbox\ on an un-patched baseline worktree at commit \5c3f4a8418202e3eef98d16ec8fc251807133b2f\; produced the exact same failure (\TestSandboxHidesHostSecrets\ - \sh: 1: Cannot fork\ due to WSL sandbox rootfs environment).

### Validation Results
- \sha256sum -c CHECKSUMS.sha256\: **PASS**
- \./source/scripts/verify-source-layout.sh\: **PASS**
- \git diff --check\: **PASS** (0 formatting/whitespace issues)
- \gofmt -l <five files>\: **PASS**
- \go test ./internal/domain/modulegraph ./internal/usecase/ports\: **PASS** (0.008s / 0.006s)
- \go vet ./internal/domain/modulegraph ./internal/usecase/ports\: **PASS**
- \CGO_ENABLED=0 go test ./internal/domain/modulegraph ./internal/usecase/ports\: **PASS**
- \make build\: **PASS**
- \make vet\: **PASS**
- \make test\: **FAIL (Stopped)** — \alidation/run-all.sh\ stopped here due to pre-existing \internal/infrastructure/sandbox\ rootfs integration test (outside Phase 5A scope; reproduced identically on clean baseline \5c3f4a8\).
- \make typecheck\: **PASS (Separate)** — executed separately after \
un-all.sh\ stopped (\	sc --noEmit\ clean 100%).

### Safety & Contract Invariants
- Patch adds NO runtime JS/TS execution, package manager invocation, bundler invocation, or lifecycle script to the analyzer. (Repository validation \make typecheck\ runs existing \pnpm run typecheck\).
- Patch adds no network access. No network access was observed in validation logs.
- Unknown / no coverage semantics were not weakened.
- Structural roots are structural roots only, not described as runtime entrypoints.

Refs #379

### DIFF
--- a/internal/domain/modulegraph/modulegraph.go
+++ b/internal/domain/modulegraph/modulegraph.go
@@ -1,0 +1,188 @@
+// Package modulegraph defines the deterministic, source-only JavaScript and
+// TypeScript module graph used by the first phase of import reachability.
+//
+// This package is intentionally pure domain code. It performs no filesystem,
+// parser, package-manager, network, SCA, judgment, or OpenVEX work.
+package modulegraph
+
+import "strings"
+
+// Dialect identifies the source grammar family associated with a module.
+type Dialect string
+
+const (
+	DialectJavaScript Dialect = "javascript"
+	DialectJSX        Dialect = "jsx"
+	DialectTypeScript Dialect = "typescript"
+	DialectTSX        Dialect = "tsx"
+)
+
+// Valid reports whether d is a supported source dialect.
+func (d Dialect) Valid() bool {
+	switch d {
+	case DialectJavaScript, DialectJSX, DialectTypeScript, DialectTSX:
+		return true
+	default:
+		return false
+	}
+}
+
+// Module is a first-party JavaScript or TypeScript source module.
+type Module struct {
+	// Path is a normalized repository-relative slash path.
+	Path string
+	// Dialect records the lexical grammar family selected from Path.
+	Dialect Dialect
+	// DeclarationOnly is true only for .d.ts, .d.mts, and .d.cts files.
+	DeclarationOnly bool
+}
+
+// ImportKind identifies a statically observed module-loading form.
+type ImportKind string
+
+const (
+	ImportESMStatic       ImportKind = "esm-static"
+	ImportESMDynamic      ImportKind = "esm-dynamic-literal"
+	ImportCommonJS        ImportKind = "commonjs-literal"
+	ImportReExport        ImportKind = "re-export"
+	ImportTypeScriptEqual ImportKind = "typescript-import-equals"
+)
+
+// Valid reports whether k is a supported import kind.
+func (k ImportKind) Valid() bool {
+	switch k {
+	case ImportESMStatic, ImportESMDynamic, ImportCommonJS, ImportReExport, ImportTypeScriptEqual:
+		return true
+	default:
+		return false
+	}
+}
+
+// Binding describes statically observable symbol binding information.
+type Binding struct {
+	Imported  string
+	Local     string
+	Default   bool
+	Namespace bool
+	TypeOnly  bool
+}
+
+// Position is a 1-based source position. Zero values mean unavailable.
+type Position struct {
+	Line   int
+	Column int
+}
+
+// Edge is an observed module relationship.
+type Edge struct {
+	// From is always a normalized repository-relative source path.
+	From string
+	// To is set only for a uniquely resolved first-party relative import.
+	// External or unresolved specifiers leave To empty.
+	To string
+	// Specifier preserves the scanner's normalized module specifier.
+	Specifier string
+	Kind      ImportKind
+	Bindings  []Binding
+	TypeOnly  bool
+	Position  Position
+}
+
+// CoverageIssueKind identifies a condition that prevents a later analyzer from
+// treating absence of an edge as a definitive negative proof.
+type CoverageIssueKind string
+
+const (
+	CoverageUnreadableFile            CoverageIssueKind = "unreadable-file"
+	CoverageSymlink                   CoverageIssueKind = "symlink"
+	CoverageInvalidUTF8               CoverageIssueKind = "invalid-utf8"
+	CoverageFileTooLarge              CoverageIssueKind = "file-too-large"
+	CoverageFileCountBudgetExceeded   CoverageIssueKind = "file-count-budget-exceeded"
+	CoverageTotalByteBudgetExceeded   CoverageIssueKind = "total-byte-budget-exceeded"
+	CoverageMalformedSource           CoverageIssueKind = "malformed-source"
+	CoverageUnsupportedSyntax         CoverageIssueKind = "unsupported-syntax"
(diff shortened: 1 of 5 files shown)


nghiadaulau (changes requested):
## Review — REQUEST CHANGES (one lint failure; the design itself is good)

Reviewed on the merged tree (pr-399 + latest main, merges clean). The domain modelling is genuinely well done and this is close to an approve; one blocker and two notes.

### What is right
- **Dependency rule respected.** `internal/domain/modulegraph` imports only stdlib (`strings`, `fmt`, `path`, `sort`) — no usecase, adapter, or infrastructure. `internal/usecase/ports/jsimports.go` imports only `context` + the domain package. Port method takes `ctx` first: `Scan(ctx context.Context, root string) (modulegraph.Graph, error)`. Exactly the layering the repo requires.
- **The model is richer than the sketch in #379, in the right way.** Instead of the two booleans `DynamicSpecifiers` / `UnsupportedSyntax`, you model `CoverageIssueKind` + `CoverageIssue` as a typed, validated enum on `Graph`. That is the single most safety-relevant part of this phase: epic #378's non-negotiable is "unknown is not unreachable", and a typed coverage-issue list lets a later Tier-1 analyzer say *why* coverage was incomplete rather than collapsing it to a bit. Good call.
- `ImportKind` covers the real shapes (`esm-static`, `esm-dynamic-literal`, `commonjs-literal`, `re-export`, plus `import x = require()` via `ImportTypeScriptEqual`), with a `Valid()` guard.
- `TypeOnly` is carried on both `Binding` and `Edge`, which is what keeps a later phase from treating `import type` as runtime evidence — a #379 requirement.
- `NormalizeRepositoryPath` rejects absolute paths and root escapes, which is the "no resolution outside the repository root" requirement, enforced in the domain rather than left to the scanner.
- Scope is honoured: types + port only, no scanner implementation and no wiring, exactly as #379 asks. Builds pass with CGO on and `CGO_ENABLED=0` linux/amd64; `go vet` clean; the two new packages' tests pass; the arch tripwire tests still pass with the new port added.

### Blocker — `golangci-lint` fails, and the failing test asserts nothing
```
internal/usecase/ports/jsimports_test.go:19:5: SA4023: this comparison is never true (staticcheck)
internal/usecase/ports/jsimports_test.go:18:32: SA4023(related information): the lhs gets its value from here and has a concrete type
```
```go
var scanner JSImportScanner = compileTimeJSImportScanner{}
if scanner == nil {                 // can never be true: a concrete struct value, not a pointer
    t.Fatal("scanner is nil")
}
```
This is a tautology — the interface variable holds a non-pointer concrete value, so the branch is unreachable and the test can never fail. That matters more than usual here: this PR's entire content is types plus a port, so this "contract test" is the only verification of the port, and it verifies nothing. Repo standard is 0 lint issues.

Fix: use the repo's existing compile-time assertion convention instead of a runtime nil check, e.g. in the test file (or the future adapter):
```go
var _ JSImportScanner = compileTimeJSImportScanner{}
```
which is how the codebase already does it (`var _ ports.DependencyGraphResolver = (*Resolver)(nil)` in `tools/gomodgraph/resolver.go:52`, `var _ ports.LogStream = (*Broker)(nil)`, etc.). If you want a real behavioural test, assert something about the returned `Graph` (for example that a zero `Graph` reports no coverage issues and validates).

### Notes (not blocking)
1. The description's validation section references `CHECKSUMS.sha256`, `./source/scripts/verify-source-layout.sh` and `run-all.sh`, which do not exist in this repository — they look like artifacts of an external harness. Please describe validation in terms of this repo's gates (`make build vet test typecheck`, `golangci-lint run`) so a reviewer can reproduce it.
2. On the "PARTIAL / STOPPED" status: I could not reproduce the `internal/infrastructure/sandbox` failure on this machine, but I agree it is unrelated — this PR touches five files, none of them sandbox code, and builds/vet/the new packages' tests are all clean. Not a concern for this PR; just note that `make test` being green end to end is still the merge bar for the phases that add the scanner.

Fix the lint/tautology and this is an approve — the domain shape is the part that is hard to get right, and you got it right.

nghiadaulau (approved):
## Re-review — APPROVE (blocker fixed)

`e51c0d5` replaces the tautological runtime check with the repo's compile-time convention:

```go
var _ JSImportScanner = compileTimeJSImportScanner{}
```

That is the right form: it is enforced when the test binary compiles, and it matches how the codebase asserts ports elsewhere (`var _ ports.DependencyGraphResolver = (*Resolver)(nil)`). Verified on the merged tree (pr-399 + latest main, merges as a no-op):

- `golangci-lint run` on both packages — **0 issues** (was 2x SA4023)
- `go vet ./...` clean (it compiles `_test.go`, so the assertion is genuinely checked)
- `go test ./internal/domain/modulegraph ./internal/usecase/ports` — PASS

I also went back through the test substance rather than just the count, and it holds up: `normalize_test.go` is table-driven with real hostile inputs — root escape (`../app.ts`), cleaned root escape (`src/../../app.ts`), unix and Windows absolute paths, an embedded NUL, `.d.ts` declaration-only rejection, dialect/extension mismatch, edges referencing unknown modules on both ends, negative counters, invalid import and coverage kinds, plus determinism/non-mutation and an explicit cycle case (`TestNormalizeCycleHasNoStructuralRoot`). That is the hostile-input and cycle coverage #379 asked for, enforced in the domain instead of deferred to the scanner.

Restating what stands from the first pass: the dependency rule is respected (domain is stdlib-only, the port imports only `context` + domain, `ctx` first), the scope is exactly Phase 5A (types + port, no scanner, no wiring), and modelling coverage limitations as a validated `CoverageIssueKind` enum rather than two booleans is the right call for epic #378's "unknown is not unreachable" invariant.

Merging. Good foundation for the resolution phase.

Issue #346, opened 2026-07-28
feat(cli): retain Code Quality source snapshots from CLI analyses


pho-veteran:
## Problem

The API composition root wires `ProjectSourceArtifactStore` into both the project and SCA services, allowing immutable Code Quality source snapshots to be captured during project analysis. The CLI composition path does not currently wire that store.

As a result, an analysis initiated through `synapse-cli` can complete without retained source artifacts, so the Code Quality source workspace cannot present an immutable source snapshot for that analysis.

## Desired outcome

When CLI analysis is used for a Code Quality project, configure and inject `ProjectSourceArtifactStore` consistently with the API path so head/base source captures and persisted manifests are available to the Code workspace.

## Acceptance criteria

- CLI configuration resolves an artifact directory and capture limits using the existing project-source configuration.
- CLI wires the store into the relevant project and SCA services.
- A CLI-created Code Quality analysis retains source metadata/artifacts and `/code/file` can serve the captured file.
- Add focused coverage for CLI composition or an end-to-end CLI project-analysis path.

nghiadaulau:
## Evaluation: the premise does not hold on current `main`. Wiring the store into the CLI would be provably unreachable code, and acceptance criterion 3 cannot be met without a new feature.

I picked this up to implement it and stopped at the point where the fix would have been dead code. Verified against `main` @ `b47f6d0`; everything below is reproducible.

### The confirmed part

The gap in the composition roots is real and exactly as described:

| Composition root | builds `projectuc` | builds `scauc` | `sourceartifact.New` | store wired |
|---|---|---|---|---|
| `cmd/synapse-api` | yes | yes | yes | yes (2 setters, `main.go:915-917`) |
| `cmd/synapse-cli` | **no** | yes | **no** | **no** |
| `cmd/synapse-worker` | no | no | no | n/a |

So AC1 and AC2 are implementable in about four lines, mirroring the API.

### Why that is not enough

`captureProjectSource` (`internal/usecase/sca/service.go:2496`) needs **three** preconditions, and the store is only the first:

1. `s.sourceArtifacts != nil` — what this issue proposes to fix.
2. `engagement.ProjectID` non-zero — otherwise it returns early with `capture_failed` (`:2503-2506`).
3. A non-empty analysis id — `opts.ProjectAnalysisID`, which is assigned from the **durable job id** and only when `opts.ProjectAnalysis` is set (`:1466-1468`).

[...]

Issue #174, opened 2026-07-12
Epic: Code Quality as a product — the full surface (Projects, ratings, issues, rules, profiles, gates)


nghiadaulau:
Turn Synapse's existing code-quality **engine** (Epic #95 — AST, complexity, duplication, A–E ratings, quality gate, coverage) into a first-class **product**: a top-level **Projects** area with the six ratings, Issues / Rules / Profiles / Gates, Overall-vs-New-Code, Activity, and clean-room per-language rule packs — all reusing the one governed scan engine.

> 📖 **Contributor guide + per-language source matrix + parity targets:** [`docs/guide/code-quality-rules.md`](https://github.com/KKloudTarus/synapse-ce/blob/main/docs/guide/code-quality-rules.md)

## 🧭 Structural decision
Introduce a first-class **`Project`** and a **top-level "Code Quality" section**, separate from pentest **Engagements** — but **reuse the same scan pipeline, evidence chain, and Judgment gate** underneath (one engine, two surfaces). Engagements keep their existing Code Quality tab for the pentest flow.

## 🎯 Target surface (our own content, industry-standard shape)
Projects home → project **overview** (quality-gate verdict + six cards: Security · Reliability · Maintainability · Security-Hotspots-Reviewed % · Coverage · Duplications, for **Overall** and **New Code**) → **Issues** (accept / false-positive / won't-fix) · **Security Hotspots** (review) · **Measures** · **Activity** · **Rules** · **Quality Profiles** · **Quality Gates**.

---

## ✅ Platform / product shell
- [x] #182 — P8 · Rule catalog + Rules explorer (the rule schema — **unblocks the language packs**)
- [x] #175 — P1 · `Project` aggregate + top-level Code Quality area
- [x] #176 — P2 · Project source ingest (Git URL / archive / path → analysis)
- [x] #177 — P3 · Analysis snapshots + Activity timeline + Overall-vs-New-Code
- [x] #178 — P4 · Project overview (gate + the six rating cards)
- [x] #181 — P7 · Issues explorer + accept / FP / won't-fix lifecycle
- [x] #183 — P9 · Quality Profiles (named, per-language)
- [x] #184 — P10 · Quality Gates (named, managed, per-project)
- [x] #179 — P5 · Security Hotspots + Reviewed %
- [x] #180 — P6 · Measures explorer (metric tree + file drill-down)

## 🧩 Language rule packs — 26 (clean-room, parity-targeted)
[...]

PR #403, opened 2026-08-07
feat(cli): capture Code Quality source snapshots for a server-owned analysis

### LINKED ISSUE #402: feat(appsec): contribute Code Quality source snapshots from CI through a sanctioned publish path
Rescoped from #346 (closed as not planned). A first attempt, PR #403, was implemented and then **withdrawn** after review; its findings are folded in below so the next attempt does not repeat them.

## Goal

Let a CI runner contribute an immutable Code Quality source snapshot for an analysis the server owns, without the CLI gaining a database and without letting a client write directly into the server's evidence namespace.

## Why the obvious approach does not work

PR #403 added `synapse-cli scan --retain-source --project-id --analysis-id`, wiring `sourceartifact` into the CLI and writing to the shared artifact directory. Two confirmed defects killed it:

1. **`/code/file` is gated on a database row the CLI cannot write.** `projectuc.ReadCodeFile` (`code_viewer.go:145-177`) calls `GetAnalysis` first, then requires `analysis.Capabilities.Source.Available` and the **persisted** manifest's file list, and only then reads the artifact store. That row is written solely by `RecordProjectAnalysis`, invoked only from `runScanJob` (`sca/service.go:1496-1498`) with a recorder registered only in `cmd/synapse-api`. So a CLI capture is either an orphan (`ErrSourceNotRetained` forever) or it replaces a server-written snapshot while the DB row still describes the old file list — divergent evidence, not retention.
2. **The default artifact directory is relative, and local scans run in place.** `ProjectSourceArtifactDir` defaults to `data/project-source-artifacts` (`config.go:378`) and `acquireLocal` returns the target dir itself. Reproduced: the snapshot was written *into the scanned repo*, the walk ingested the store's own temp blob, and a `.env` containing an AWS key was gzipped verbatim into the checkout — where `upload-artifact`, a Docker context or a cache step would ship it. Golden rule 3 (secrets must not become durable artifacts).

## Required design: one authority writes both halves

The CLI must **publish**, not write:

- New authenticated endpoint, e.g. `POST /api/v1/projects/{key}/analyses/{analysisID}/source`, that:
  - derives the tenant from the caller's token — never from a request field;
  - verifies the analysis exists, belongs to that project and tenant, and does not already carry a retained snapshot;
  - accepts the manifest plus the file bytes, applies the same capture limits server-side, and writes the artifact tree **and** the analysis row's `SourceManifest`/`Capabilities` in one transaction-like step;
  - records a durable audit entry with the actor.
- `synapse-cli` gains a publish command/flag that streams the analyzed tree to that endpoint. It keeps no artifact store, so the scan path stays DB-free.

Alternative, equally acceptable: **keep capture server-side entirely** and close this as won't-do. That is a legitimate outcome — say so explicitly rather than shipping a client-side write.

## Guards the withdrawn attempt proved are necessary

Whichever shape is chosen:

- [ ] Tenant and project come from the authenticated identity, never from caller input. (#403 took `--tenant-id` verbatim; path traversal was impossible because the store hashes ids, but namespace targeting was not.)
- [ ] Create-only capture: refuse when a snapshot already exists for that analysis, with a distinct unavailable reason, instead of `RemoveAll` + `Rename` (`sourceartifact/store.go:166-171`).
- [ ] Writer provenance inside `manifest.json` (actor, tool version, timestamp) covered by the manifest digest, so a server-written and a client-contributed snapshot are distinguishable.
- [ ] A durable audit entry for the artifact write. The CLI's `stderrAudit` is not persistence.
- [ ] The artifact root must be operator-owned and absolute; refuse a relative root, and never let it resolve inside the scanned tree. Exclude it from the capture walk.
- [ ] Honour ignore rules so `.env`, key material and state files are not captured; the secret redactor does not currently reach captures.
- [ ] Refuse non-source targets (`--image`, a single `.jar`/archive) in the usecase, not only in `cmd/*`.
- [ ] Do not reuse `ScanOptions.ProjectAnalysis` as the capture switch: it also makes the scan read `.synapse-gate.yaml` from the scanned tree and hard-fail on a decode error (`sca/service.go:1727-1743`). Capture needs its own option.
- [ ] Honour `ProjectSourceRetention`; `CleanupExpired` must run for whatever writes snapshots.

## Acceptance criteria

- [ ] A snapshot contributed through the sanctioned path is served by `/code/file` **through `projectuc.ReadCodeFile`**, proven by a test that goes through `ReadCodeFile` (not one that calls `store.Capture`/`Load` directly — that was the misleading test in #403).
- [ ] A contribution for an unknown, foreign-tenant, or already-captured analysis is refused with a clear error and a non-zero exit.
- [ ] No secret-bearing or ignored file is captured; verified with a fixture repo containing `.env` and key material.
- [ ] No artifact is ever written inside the scanned tree.
- [ ] The scan path remains DB-free (#174 non-negotiable) and default behaviour is unchanged when the feature is not requested.
- [ ] The artifact write is audited with an attributable actor.

## Out of scope

- Base/comparison capture (`captureProjectComparison`) — needs a git base ref in the workspace.
- Creating or registering the analysis from the CLI; the server owns analysis identity.

## References

- #346 — original issue, closed as not planned (wiring alone is unreachable).
- #403 — withdrawn attempt; the review comment there has the full evidence and reproductions.

nghiadaulau:
Closes #402. Rescoped from #346, which was closed as not planned.

## Why #346 could not be implemented as written

`captureProjectSource` needs three preconditions, and #346 proposed fixing only the first:

1. `sourceArtifacts != nil`
2. a non-zero `engagement.ProjectID`
3. a non-empty analysis id (`opts.ProjectAnalysisID`, assigned from the durable job id)

The CLI had none of 2 or 3: its engagement is synthetic (`engagement.New` takes no `ProjectID`) and it passes `nil` for the result/job/run stores because the scan path is intentionally memory-only. Wiring the store alone was provably unreachable, verified empirically: with the store injected, a CLI-shaped capture still returned `Source.Available=false reason=capture_failed` without ever reaching `Capture()`.

## What this does instead

Rather than adding a second capture path, it lets the operator supply the project-analysis context explicitly, which makes the **existing** capture reachable:

```
synapse-cli scan <path> --retain-source --project-id <id> --analysis-id <id> [--tenant-id <id>]
```

- Wires `sourceartifact.New(...)` into the CLI's SCA service, resolving the directory and all three capture limits from the existing `SYNAPSE_PROJECT_SOURCE_*` configuration the API already reads.
- Stamps `TenantID`/`ProjectID` on the scan's engagement and sets `ScanOptions.ProjectAnalysis` + `ProjectAnalysisID`, so `captureProjectSource` runs unchanged.
- Reports the outcome on stderr, because capture failure is metadata and never fails the scan — a CI log would otherwise look identical whether or not a snapshot exists.

Design choices worth calling out:

- **The CLI never mints identities.** The analysis is created server-side (its id *is* the durable job id) and CI passes it in. `--retain-source` without `--project-id` and `--analysis-id` is a usage error, exit 2, not a silent skip.
- **No new format and no database.** It reuses `sourceartifact`, so a CLI capture is byte-identical in layout to a server-side one, and #174's "the CLI gate still works with no DB" still holds.
- **Default behaviour is untouched.** Without the flag, nothing changes.

## Verification

End-to-end with the built binary (no database):

```
$ synapse-cli scan FX --retain-source                      # exit 2
synapse-cli: --retain-source requires --project-id and --analysis-id (the server owns the analysis identity)
$ synapse-cli scan FX --project-id p1                      # exit 2
synapse-cli: --project-id/--analysis-id/--tenant-id require --retain-source
$ SYNAPSE_PROJECT_SOURCE_ARTIFACT_DIR=$ART synapse-cli scan FX \
    --retain-source --project-id proj-42 --analysis-id job-abc
synapse-cli: source snapshot retained for project proj-42 analysis job-abc (2 file(s), digest sha256:c592806f...)
$ find $ART -type f
.../v2/t-<hash>/p-<hash>/a-<hash>/manifest.json
.../v2/t-<hash>/p-<hash>/a-<hash>/blobs/<sha256>.gz   (x2)
```

Same layout the API store uses, so `/code/file` can serve it from a shared volume. Without the flag, no retention line is emitted.

Tests (`cmd/synapse-cli/source_retention_test.go`): flag-validation table (9 cases, both directions), identity trimming, a compile-time assertion that the wired store satisfies `ports.ProjectSourceArtifactStore`, and an end-to-end capture-then-`Load` through the same key `/code/file` resolves with.

Gates: `go build ./...` (CGO on) and `CGO_ENABLED=0` for linux/amd64 and windows/amd64, `go vet ./...`, tests for `cmd/...`, `sca`, `projectuc`, `sourceartifact`, and `golangci-lint run ./cmd/...` — all clean, 0 lint issues.

## Out of scope

Base/comparison capture (`captureProjectComparison`) needs a git base ref in the workspace; it can follow separately. Creating or registering the analysis from the CLI stays server-side by design.


nghiadaulau:
## Withdrawing this PR: my own adversarial review found the design is wrong, and two of the defects are serious. Details below so the next attempt does not repeat them.

I built this, then reviewed it the way I would review anyone else's change. Two findings are confirmed by execution, not inspection, and either one is disqualifying.

### 1. The feature cannot achieve its own acceptance criterion. `/code/file` needs a database row the CLI never writes.

`projectuc.ReadCodeFile` (`internal/usecase/projectuc/code_viewer.go:145-177`) resolves ownership **from the database first**:

```go
analysis, err := s.GetAnalysis(ctx, tenantID, key, analysisID)   // DB row
...
if !analysis.Capabilities.Source.Available || s.sourceArtifacts == nil { return ..., ErrSourceNotRetained }
file, found := codeFileByPath(analysis, canonical)               // persisted manifest
...
data, source, err = s.sourceArtifacts.Load(...)                  // only now the artifact store
```

That row is written only by `RecordProjectAnalysis`, called only from `runScanJob` (`internal/usecase/sca/service.go:1496-1498`) and only when a recorder is registered — which happens exclusively in `cmd/synapse-api/main.go:911`. The CLI takes the synchronous `ScanWithOptions` path and registers no recorder.

So a CLI capture has exactly two outcomes, both bad:

- **Fresh analysis id** — the blobs are a permanent orphan. `/code/file` returns `ErrSourceNotRetained` forever, while the CLI prints "source snapshot retained".
- **Real server analysis id** — the capture **replaces** the server's snapshot bytes (`sourceartifact/store.go:166-171` does `RemoveAll` + `Rename`) while the DB row still describes the server's file list. The stored manifest and the artifact tree can then disagree: that is evidence corruption, not retention.

This is the same defect class I closed #346 for. I should have traced `/code/file` end to end before writing the flag rather than after.

My test made it look fine, which is the other half of the mistake: `TestCLICaptureIsServableByCodeFile` calls `store.Capture` and `store.Load` directly and never touches `ReadCodeFile`, `GetAnalysis`, or `run()`. It proves the store round-trips, which `sourceartifact/store_test.go` already covered. The name asserted something the body did not test.

### 2. With the default configuration it writes repo secrets into the checkout.

`config.go:378` defaults `ProjectSourceArtifactDir` to the **relative** `data/project-source-artifacts`, and a local scan runs in place (`acquire/acquirer.go:112-117` returns `Workspace{Dir: value}` with no copy). So the canonical CI invocation writes the snapshot into the repository being scanned. Reproduced in a temp repo containing a `.env`:

```
$ synapse-cli scan . --retain-source --project-id p1 --analysis-id a1
synapse-cli: source snapshot retained for project p1 analysis a1 (4 file(s), digest sha256:27ce0876...)

$ find . -name manifest.json
./data/project-source-artifacts/v2/t-<hash>/p-<hash>/a-<hash>/manifest.json

manifest files: 4
.env captured?: True
self-ingested own artifacts?: True
sample: ['.env', 'data/project-source-artifacts/.../.source-capture-328304978/blobs/8b8c3a09....gz', 'go.mod', 'main.go']
```

Three problems in one run: the `.env` (an AWS secret) is gzipped verbatim into the checkout where `upload-artifact`, a Docker build context or a cache step will ship it; the walk ingests the store's own temporary blob; and nothing ever cleans it up, because `CleanupExpired` is never called on this path and there is no DB row to trigger the API's orphan deletion. The CLI's secret redaction does not apply to captures. That is a golden-rule-3 (secrets must not become durable artifacts) problem that my change made reachable.

### Also real, and worth carrying forward

- **No ownership check on the write.** `--tenant-id` / `--project-id` are taken verbatim and passed to `Store.Capture`; the API path resolves the project first (`projectuc/service.go:420`). On the shared volume this PR recommends, a job could target another tenant's key. Path traversal is *not* possible (the store hashes all three ids into fixed hex segments, `store.go:437-443`), but namespace targeting is.
- **No create-only mode, no provenance, no audit.** `Capture` clobbers unconditionally and `SourceManifest` records no actor/time/writer, so a server-written and a CLI-written snapshot are indistinguishable afterwards. Announcing the replacement on stderr, which is what I did, is not an immutability control.
- **`ProjectAnalysis: true` does more than capture.** It also makes the scan read `.synapse-gate.yaml` from the scanned tree and hard-fail on a decode error (`sca/service.go:1727-1743`), so a repo-committed file could fail CI only when `--retain-source` is passed. Capture needs its own flag.
- **`--tenant-id` is optional but effectively mandatory**: omitted, the snapshot lands under `t-sha256("")` while the API reads with the real tenant, so it is unservable — and my test encoded that wrong shape as valid.
- **The `--image` guard is incomplete and in the wrong layer**: a local `.jar`/archive target also produces a non-source "snapshot", and the check lives in `cmd/` where no server-side path can enforce it.

### What the right shape looks like

Bytes and metadata must be written by one authority. Either:

1. the CLI **publishes** to an authenticated API endpoint that derives the tenant from the token, validates project/analysis ownership, then writes the DB row and the artifacts together; or
2. capture stays server-side and the CLI stays a reporter.

I have rewritten #402 with option 1 spelled out, plus the guards above (absolute operator-owned artifact root, exclude the artifact root from the walk, honour ignore rules for secrets, create-only capture with writer provenance and a durable audit entry, require the tenant, refuse non-directory targets in the usecase rather than in `cmd`, and decouple capture from `ProjectAnalysis`).

Closing rather than iterating in place: the flag surface and the plumbing both change under option 1, so there is little to salvage beyond the flag-validation tests. The branch stays pushed as a record of the discarded attempt.

Issue #400, opened 2026-08-06
feat(reachability): resolve JS/TS imports to npm workspaces and SBOM component PURLs


VietSory:
Parent: #378

Depends on: #379

Objective

Resolve external JavaScript and TypeScript module specifiers from the source module graph into deterministic package identities.

This phase connects the source relationships produced by #379 with:

Node.js built-in modules;
first-party npm workspaces;
third-party npm packages;
exact SBOM component PURLs where the mapping is deterministic.

The resolver must not classify packages as reachable or unreachable and must not create judgments or OpenVEX statements. Its responsibility is limited to package identity, resolution provenance and explicit ambiguity.

Background

The module graph produced by #379 preserves source-level relationships such as:

src/index.ts → "@scope/pkg/subpath"
src/service.ts → "lodash/fp"
packages/api/src/main.ts → "@workspace/shared"

These specifiers are not yet sufficient for reachability analysis.

A later Tier-1 analyzer needs to know whether an import refers to:

a Node.js built-in;
another package in the same repository;
a unique third-party SBOM component;
multiple possible component versions;
an unsupported or unresolved alias.

The resolver must therefore turn observable module specifiers into deterministic identities without executing Node.js, package managers, bundlers or project configuration.

Architectural decisions

This issue should preserve the focused architecture established by #379.

A proposed layout is:

internal/domain/jsresolution
internal/infrastructure/tools/jsresolve
internal/usecase/ports

The implementation should introduce a dedicated resolver port rather than add package-resolution behavior to modulegraph.

An illustrative contract is:

typeJSImportResolver interface {
	Resolve(
		ctx context.Context,
		root string,
		graph modulegraph.Graph,
		doc *sbom.SBOM,
	) (jsresolution.Result, error)
}

Use the actual identifier JSImportResolver; spacing in the example may be adjusted for Markdown rendering.

The domain result must remain independent of filesystem and parser implementation details.

Proposed resolution model

An illustrative model is:

type Status string

const (
	StatusBuiltin    Status = "builtin"
	StatusWorkspace  Status = "workspace"
	StatusComponent  Status = "component"
	StatusUnresolved Status = "unresolved"
	StatusAmbiguous  Status = "ambiguous"
	StatusUnsupported Status = "unsupported"
)

type PackageIdentity struct {
	Name      string
	Version   string
	PURL      string
	Workspace bool
	Path      string
}

type ImportResolution struct {
	From      string
	Specifier string
[...]

PR #404, opened 2026-08-07
feat(jsresolve): add R2A workspace metadata inventory

VietSory:
## Summary

Begins R2 for #400 with the deterministic, offline workspace/package metadata foundation that later module-specifier and SBOM-correlation stages will consume.

This PR intentionally does **not** make reachability judgments and does not implement R2B/R2C package resolution yet.

### Included

* new `internal/domain/jsresolution` domain contract

  * explicit package-resolution statuses
  * package identities and candidate preservation
  * typed incomplete-coverage reasons
  * source graph coverage preservation
  * deterministic package/workspace inventory normalization
* new dedicated `ports.JSImportResolver` contract
* bounded offline `internal/infrastructure/tools/jsresolve.InventoryBuilder`
* root and nested `package.json` discovery
* npm/Yarn workspace array and object declarations
* `pnpm-workspace.yaml` scalar workspace lists
* `*`, `**`, negation and standard segment glob matching
* scoped npm package-name normalization
* repository-root escape rejection
* symlinked workspace refusal without following the link
* explicit coverage for malformed/unsupported metadata, budgets and duplicate workspace names
* deterministic sorting/deduplication and derived `Complete`

### Explicitly deferred to follow-up R2 work

* Node.js built-in/specifier classification
* `package.json#imports` resolution
* tsconfig/jsconfig path alias resolution
* lockfile importer/version disambiguation
* SBOM component/PURL correlation
* reachability decisions, judgments and OpenVEX

### Validation

R2A validation completed successfully against the repository gates:

* `golangci-lint run`: PASS
* `make build`: PASS
* `make vet`: PASS
* `make test`: PASS
* `make typecheck`: PASS
* Linux `go build ./...`: PASS
* Linux `go vet ./...`: PASS
* Linux `go test ./...`: PASS
* `CGO_ENABLED=0 go build ./cmd/...`: PASS
* Windows `go build ./...`: PASS
* Windows `go vet ./...`: PASS
* Windows `go test ./...`: PASS
* Web production build: PASS
* Focused R2A package tests: PASS
* Focused race tests: PASS
* Repeated deterministic test runs: PASS

The repository's standard frontend `pnpm test` job currently fails before executing tests with:

`TypeError: webidl.util.markAsUncloneable is not a function`

This was reproduced independently on the unchanged upstream baseline, with the same Vitest/jsdom/undici worker-start failure. The R2A diff does not modify frontend code, `package.json`, or the lockfile, so this failure is baseline/toolchain-related and is not introduced by this PR.

Covered R2A cases include npm/Yarn workspaces, pnpm `**` + negation, scoped names, malformed metadata, invalid/escaping globs, duplicate workspace names, symlinked workspaces, cancellation, invalid roots, budget handling, and repeated deterministic output.

Full repository lint/build/test/typecheck gates remain pending CI/reviewer execution.

Refs #400
Parent #378
Depends on #379, merged via #399.

### DIFF
--- a/.ci-trigger
+++ b/.ci-trigger
@@ -1,0 +1,1 @@
+r2a semantic hardening
--- a/.ci-trigger
+++ b/.ci-trigger
@@ -1,1 +1,0 @@
-r2a semantic hardening
--- a/.github/ci-note
+++ b/.github/ci-note
@@ -1,0 +1,1 @@
+temporary
--- a/.github/ci-note
+++ b/.github/ci-note
@@ -1,1 +1,0 @@
-temporary
--- a/internal/domain/jsresolution/jsresolution.go
+++ b/internal/domain/jsresolution/jsresolution.go
@@ -55,17 +55,17 @@ type ImportResolution struct {
 type CoverageIssueKind string
 
 const (
-	CoverageUnreadableMetadata         CoverageIssueKind = "unreadable-metadata"
-	CoverageMalformedMetadata          CoverageIssueKind = "malformed-metadata"
-	CoverageUnsupportedMetadata        CoverageIssueKind = "unsupported-metadata"
-	CoverageMetadataBudgetExceeded     CoverageIssueKind = "metadata-budget-exceeded"
-	CoverageWorkspaceRootEscape        CoverageIssueKind = "workspace-root-escape"
-	CoverageSymlinkWorkspace           CoverageIssueKind = "symlink-workspace"
-	CoverageWorkspaceNameConflict      CoverageIssueKind = "workspace-name-conflict"
-	CoverageUnresolvedAlias            CoverageIssueKind = "unresolved-alias"
-	CoverageUnsupportedAlias           CoverageIssueKind = "unsupported-alias"
-	CoverageMissingSBOMComponent       CoverageIssueKind = "missing-sbom-component"
-	CoverageAmbiguousSBOMComponent     CoverageIssueKind = "ambiguous-sbom-component"
+	CoverageUnreadableMetadata        CoverageIssueKind = "unreadable-metadata"
+	CoverageMalformedMetadata         CoverageIssueKind = "malformed-metadata"
+	CoverageUnsupportedMetadata       CoverageIssueKind = "unsupported-metadata"
+	CoverageMetadataBudgetExceeded    CoverageIssueKind = "metadata-budget-exceeded"
+	CoverageWorkspaceRootEscape       CoverageIssueKind = "workspace-root-escape"
+	CoverageSymlinkWorkspace          CoverageIssueKind = "symlink-workspace"
+	CoverageWorkspaceNameConflict     CoverageIssueKind = "workspace-name-conflict"
+	CoverageUnresolvedAlias           CoverageIssueKind = "unresolved-alias"
+	CoverageUnsupportedAlias          CoverageIssueKind = "unsupported-alias"
+	CoverageMissingSBOMComponent      CoverageIssueKind = "missing-sbom-component"
+	CoverageAmbiguousSBOMComponent    CoverageIssueKind = "ambiguous-sbom-component"
 	CoverageUnsupportedPackageManager CoverageIssueKind = "unsupported-package-manager"
 )
 
--- a/internal/domain/jsresolution/jsresolution.go
+++ b/internal/domain/jsresolution/jsresolution.go
@@ -1,0 +1,134 @@
+// Package jsresolution models deterministic JavaScript and TypeScript package
+// identity resolution without embedding filesystem or parser implementation details.
+package jsresolution
+
+import "github.com/KKloudTarus/synapse-ce/internal/domain/modulegraph"
+
+// Status is the explicit package-identity resolution state for an external import.
+type Status string
+
+const (
+	StatusBuiltin     Status = "builtin"
+	StatusWorkspace   Status = "workspace"
+	StatusComponent   Status = "component"
+	StatusUnresolved  Status = "unresolved"
+	StatusAmbiguous   Status = "ambiguous"
+	StatusUnsupported Status = "unsupported"
+)
+
+// Valid reports whether s is a supported resolution status.
+func (s Status) Valid() bool {
+	switch s {
+	case StatusBuiltin, StatusWorkspace, StatusComponent, StatusUnresolved, StatusAmbiguous, StatusUnsupported:
+		return true
+	default:
+		return false
+	}
+}
+
+// PackageIdentity is a deterministic package identity selected or proposed by
+// the resolver. Workspace identities remain distinguishable from SBOM components.
+type PackageIdentity struct {
+	Name      string
+	Version   string
+	PURL      string
+	Workspace bool
+	Path      string
+}
+
+// ImportResolution preserves source-edge semantics while recording package identity.
+type ImportResolution struct {
+	From            string
+	Specifier       string
+	Position        modulegraph.Position
+	Kind            modulegraph.ImportKind
+	TypeOnly        bool
+	DeclarationOnly bool
+	Status          Status
+	Package         PackageIdentity
+	Candidates      []PackageIdentity
+	Reason          string
+}
+
+// CoverageIssueKind identifies a condition that makes package identity
+// resolution incomplete or unsafe for a later negative reachability conclusion.
+type CoverageIssueKind string
+
+const (
+	CoverageUnreadableMetadata         CoverageIssueKind = "unreadable-metadata"
+	CoverageMalformedMetadata          CoverageIssueKind = "malformed-metadata"
+	CoverageUnsupportedMetadata        CoverageIssueKind = "unsupported-metadata"
+	CoverageMetadataBudgetExceeded     CoverageIssueKind = "metadata-budget-exceeded"
+	CoverageWorkspaceRootEscape        CoverageIssueKind = "workspace-root-escape"
+	CoverageSymlinkWorkspace           CoverageIssueKind = "symlink-workspace"
+	CoverageWorkspaceNameConflict      CoverageIssueKind = "workspace-name-conflict"
+	CoverageUnresolvedAlias            CoverageIssueKind = "unresolved-alias"
+	CoverageUnsupportedAlias           CoverageIssueKind = "unsupported-alias"
+	CoverageMissingSBOMComponent       CoverageIssueKind = "missing-sbom-component"
+	CoverageAmbiguousSBOMComponent     CoverageIssueKind = "ambiguous-sbom-component"
+	CoverageUnsupportedPackageManager CoverageIssueKind = "unsupported-package-manager"
+)
+
+// Valid reports whether k is a supported coverage issue kind.
+func (k CoverageIssueKind) Valid() bool {
+	switch k {
+	case CoverageUnreadableMetadata,
+		CoverageMalformedMetadata,
+		CoverageUnsupportedMetadata,
+		CoverageMetadataBudgetExceeded,
+		CoverageWorkspaceRootEscape,
+		CoverageSymlinkWorkspace,
+		CoverageWorkspaceNameConflict,
+		CoverageUnresolvedAlias,
+		CoverageUnsupportedAlias,
+		CoverageMissingSBOMComponent,
+		CoverageAmbiguousSBOMComponent,
+		CoverageUnsupportedPackageManager:
+		return true
+	default:
+		return false
+	}
+}
+
+// CoverageIssue records deterministic, attributable incomplete package-resolution coverage.
+type CoverageIssue struct {
+	Kind   CoverageIssueKind
+	Path   string
+	Detail string
+}
+
+// MetadataDeclaration records where a workspace relationship was declared.
+type MetadataDeclaration struct {
+	Source  string
+	Pattern string
+}
+
+// PackageMetadata is package.json-derived first-party package metadata.
+type PackageMetadata struct {
+	Name       string
+	Version    string
+	Path       string
+	Private    bool
+	Workspace  bool
+	DeclaredBy []MetadataDeclaration
+}
+
+// Inventory is the deterministic offline package/workspace metadata inventory
+// consumed by later module-specifier and SBOM-correlation stages.
+type Inventory struct {
+	Packages       []PackageMetadata
+	Coverage       []CoverageIssue
+	Complete       bool
+	EntriesScanned int
+	FilesScanned   int
+}
+
+// Result is the package-resolution output consumed by a later Tier-1 analyzer.
+// Complete is false whenever package-resolution coverage is incomplete; source
+// graph coverage is preserved separately and must also be considered by callers.
+type Result struct {
+	Imports       []ImportResolution
+	Coverage      []CoverageIssue
+	GraphCoverage []modulegraph.CoverageIssue
+	Complete      bool
+}
--- a/internal/domain/jsresolution/normalize.go
+++ b/internal/domain/jsresolution/normalize.go
@@ -147,9 +147,9 @@ func normalizeDeclarations(in []MetadataDeclaration) ([]MetadataDeclaration, err
 			return nil, err
 		}
 		decl.Source = loc
-		// Pattern is declared metadata, not a filesystem-derived location, so
-		// Windows-style separators are normalized here intentionally.
-		decl.Pattern = strings.TrimSpace(strings.ReplaceAll(decl.Pattern, "\\", "/"))
+		// Pattern is declared metadata, not a filesystem-derived location. Keep
+		// its literal whitespace intact; only normalize the declared separator.
+		decl.Pattern = strings.ReplaceAll(decl.Pattern, "\\", "/")
 		out = append(out, decl)
 	}
 	sort.Slice(out, func(i, j int) bool {
--- a/internal/domain/jsresolution/normalize.go
+++ b/internal/domain/jsresolution/normalize.go
@@ -1,0 +1,196 @@
+package jsresolution
+
+import (
+	"fmt"
+	"path"
+	"sort"
+	"strings"
+)
+
+// NormalizePackageName canonicalizes an npm package name. It preserves scoped
+// package roots and rejects strings that cannot safely identify an npm package.
+func NormalizePackageName(raw string) (string, error) {
+	name := strings.ToLower(strings.TrimSpace(raw))
+	if name == "" {
+		return "", fmt.Errorf("jsresolution: empty package name")
+	}
+	if strings.ContainsAny(name, "\\\x00\r\n\t ") {
+		return "", fmt.Errorf("jsresolution: invalid package name %q", raw)
+	}
+	if strings.HasPrefix(name, "@") {
+		parts := strings.Split(name[1:], "/")
+		if len(parts) != 2 || !validPackagePart(parts[0]) || !validPackagePart(parts[1]) {
+			return "", fmt.Errorf("jsresolution: invalid scoped package name %q", raw)
+		}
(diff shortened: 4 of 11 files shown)


nghiadaulau (changes requested):
## Review — REQUEST CHANGES: one blocker (a hostile repo can hang the scan), everything else is genuinely strong

This is careful, conservative work and most of it I verified by execution rather than reading. One defect is disqualifying for a component whose whole job is to parse an untrusted repository.

### Blocker — exponential backtracking in the workspace glob matcher: a repo can hang `Build` indefinitely

`internal/infrastructure/tools/jsresolve/inventory.go:345-365` — `matchSegments` recurses over `**` with no memoization:

```go
if pattern[pi] == "**" {
    for next := ti; next <= len(target); next++ {
        if matchSegments(pattern, target, pi+1, next) { return true }
    }
    return false
}
```

On a non-matching target this explores every split point per `**`, i.e. O(len(target)^N) for N recursive wildcards. Measured against a 24-segment non-matching target:

| `**` segments | elapsed |
|---|---|
| 4 | 0.46 ms |
| 6 | 16.7 ms |
| 8 | 246 ms |
| 10 | **2.98 s** |
| 12 | **> 5 s (did not finish)** |

Both inputs are attacker-controlled: the pattern comes from the scanned repository's own `package.json` `"workspaces"` (or `pnpm-workspace.yaml`), and the target paths come from its directory layout. Reproduced end to end through the public API — a repo whose root `package.json` is

```json
{"name":"r","workspaces":["**/**/**/**/**/**/**/**/**/**/**/**/**/**/pkg"]}
```

plus a 24-deep directory containing a `package.json`:

```
DoS CONFIRMED: Build did not return within 10s despite a 3s context deadline
```

Two things make this worse than a slow path:

1. **The existing budgets do not bound it.** `maxEntries` (100k), `maxMetadataFiles` (10k) and `maxMetadataBytes` (2 MiB) all bound the *walk*, not per-candidate match cost. The hostile pattern is one short string, far under every limit.
2. **The context deadline cannot interrupt it.** `ctx.Err()` is checked between entries (`:72, :93, :189`) but never inside `matchSegments`, so the 3-second deadline above was ignored — this is pure CPU in one call. That also means a caller's scan timeout will not save it.

Fixes, any of which is small:

- **Memoize on `(pi, ti)`** — turns it into O(len(pattern) × len(target)). This is the standard fix and I would prefer it.
- Collapse consecutive `**` at compile time (`**/**` ≡ `**`) — kills the trivial attack but not the general shape (`**/a/**/a/**`), so it is a complement, not a substitute.
- Cap `**` count per pattern and record a `CoverageUnsupportedMetadata` issue — that fits your explicit-coverage model nicely and fails honest rather than slow.
- Independently, check `ctx` inside the matcher (or bound its iterations) so a deadline is respected.

Please add the adversarial pattern as a regression test with a time bound; the current tests only exercise well-formed globs.

### Should-fix — coverage kinds are classified by matching error *strings*

`inventory.go:253-256`:

```go
kind := jsresolution.CoverageUnsupportedMetadata
if strings.Contains(err.Error(), "escapes root") || strings.Contains(err.Error(), "absolute") {
    kind = jsresolution.CoverageWorkspaceRootEscape
}
```

There are no sentinel errors in the package, so the typed coverage reason — which is exactly the signal a later Tier-1 analyzer will consume — depends on the prose of an error message. Reword the error in `normalizeWorkspacePattern` and a root-escape attempt silently becomes "unsupported metadata". The safety invariant still holds (an issue is recorded either way, so `Complete` still goes false), but the diagnostic value degrades silently. Please add `errWorkspaceRootEscape` / `errWorkspaceAbsolutePath` and classify with `errors.Is`.

### What I verified as correct

Gates, on the merged tree (merges as a no-op with current `main`):
- `go build ./...` (CGO on) and `CGO_ENABLED=0` for linux/amd64 and windows/amd64 — OK
- `go vet ./...` — clean; `golangci-lint run` on all three packages — **0 issues**
- tests pass, including with `-count=3` and `-shuffle=on`

Substance:
- **Dependency rule respected.** `domain/jsresolution` imports only stdlib plus `domain/modulegraph`; the port imports only `context` + domain; the builder lives in infrastructure. `Resolve(ctx, root, graph, sbom)` takes `ctx` first.
- **The port test uses the compile-time convention** (`var _ JSImportResolver = compileTimeJSImportResolver{}`) rather than a runtime nil check — thanks for picking that up from #399.
- **`Complete` is structurally honest.** `normalize.go:136` derives `out.Complete = len(out.Coverage) == 0`, so a snapshot can never claim completeness while anything was skipped. That is exactly epic #378's "unknown is not unreachable", enforced by construction rather than by discipline.
- **Root escape is genuinely rejected**, not prefix-matched: absolute and Windows-volume paths at `:298`, and post-join `..` / `../` at `:317-318`.
- **Symlinked workspaces are refused without being followed** — `os.Lstat` at `:79` and `:514`, `entry.Type()&os.ModeSymlink` during the walk at `:116`.
- **No YAML library.** Parsing `pnpm-workspace.yaml` as a scalar list by hand, and refusing flow style, removes anchor/alias-bomb and deep-nesting exposure entirely. Deliberately boring, and right.
- **Deterministic.** Every map-derived collection is sorted before it leaves the builder (`:195, :196, :220, :564, :570`, plus `normalize.go:103, :125, :151`).
- **No exec, no network, no secrets logged.** `.git`/`.hg`/`.svn`/`node_modules` are skipped.

Empirical hostile-repo check (root escape + absolute pattern + symlink pointing outside + malformed JSON in one tree):

```
Complete=false  coverage=map[malformed-metadata:1 symlink-workspace:1 workspace-root-escape:2]
packages: "r" (.), "@s/good" (packages/good)
```

No package from outside the root was inventoried, the symlink was refused, and every failure produced a typed reason. That is the behaviour the epic needs.

### Scope

Matches #400's proposed layout exactly (`domain/jsresolution`, `tools/jsresolve`, a dedicated resolver port rather than extending `modulegraph`), makes no reachability judgment, and defers builtin classification, `imports`, tsconfig aliases, lockfile disambiguation and SBOM correlation explicitly. Correctly described as beginning R2 rather than closing #400. The forward-declared coverage kinds (`unresolved-alias`, `missing-sbom-component`) are unused for now, which is fine as a contract declaration.

Fix the matcher (and ideally the sentinel errors) and this is an approve — the safety model here is the part that is hard to get right, and you got it right.

nghiadaulau:
## Follow-up: a deeper pass found four more issues, and one of them is a bigger deal than the glob hang I reported above

All four are reproduced with tests I ran against this branch, not inferred from reading.

### New blocker — a standard `pnpm-workspace.yaml` silently yields zero workspaces while reporting `Complete = true`

`inventory.go:438` requires list items to be *strictly more indented* than `packages:`, then `:458-461` returns `(nil, nil)` when nothing was collected — success, no coverage issue. But this is ordinary YAML, and it is exactly what js-yaml (so pnpm itself) accepts:

```yaml
packages:
- 'apps/*'
- 'libs/*'
```

`indent == 0 == baseIndent` → immediate `break` → `patterns == nil` → `found` is true, so no error and no coverage entry. Result on a repo with `apps/web/package.json`:

```
same-indent pnpm: Complete=true  coverageIssues=0  workspacesDetected=0
```

Every workspace is missed and the inventory claims **full confidence**. That inverts epic #378's non-negotiable: a later Tier-1 analyzer would treat first-party workspace packages as unresolved or as third-party components while `Complete = true` tells it the picture is trustworthy. This is worse than the glob hang, because a hang is loud and this is silent — and it needs no hostile input at all, just a common indentation style.

Same silent path applies to tab-indented items (`leadingSpaces` at `:502-511` counts only spaces) and to a `packages:` key with an empty body.

Fix: accept same-indent block sequences, and make "found `packages:` but parsed zero patterns" a hard `malformed-metadata` / `unsupported-metadata` issue instead of `(nil, nil)`.

### New — negation is order-sensitive, so an explicitly excluded package becomes first-party

`inventory.go:265-284` implements "last match wins", so a negation only takes effect if it appears *after* the positive pattern. picomatch/globby (hence yarn and pnpm) apply negatives as exclusions regardless of order. Reproduced:

```
workspaces: ["!packages/private/**", "packages/**"]
→ @x/secret Workspace=true      (real pnpm/yarn would exclude it)
```

The direction matters for this epic: a package the manifest deliberately excludes gets classified as a first-party workspace instead of an SBOM component, which is precisely the misclassification that would later suppress a vulnerability correlation. Fix: two passes — collect positives, then apply all negatives.

### New — one oddly-named directory aborts the entire inventory

`normalize.go:52,56` applies Windows path rules to paths that came from the POSIX filesystem via `repositoryRelative`, where `\` and `:` are legal filename characters. A directory literally named `c:tmp`:

```
err = normalize package path "c:tmp": jsresolution: absolute repository location "c:tmp"
packages = 0
```

`Build` fails outright — no inventory, no coverage, nothing. One `mkdir` in a scanned repo denies the whole JS analysis. The mirror case is conflation: a directory named `packages\private` is rewritten to `packages/private` and can collide with a real `packages/private` (hard failure via `normalize.go:91-95`, or silent shadowing that inherits `Workspace = true`).

Fix: apply the volume-prefix and backslash rules only to *declared pattern text*, never to filesystem-derived paths; and degrade to a typed coverage issue rather than aborting the whole inventory on one filename.

### New — the cartesian product and the coverage slice are both uncapped

`inventory.go:198-214` matches every pattern of every source against every package dir *and* every symlink, with no cap on pattern count and — confirmed by grep — **no cap anywhere on `len(Coverage)`**. Two shapes, both inside the existing budgets:

- 10,000 manifests of `{"workspaces":["*"]}` (~250 KB total) → 10,000 × 10,000 = 1e8 matches, each allocating two `Trim`s and two `Split`s (`:326-343`).
- ~9,000 manifests of `{"workspaces":["**"]}` plus ~80,000 tiny symlinks (under the 100k entry budget) → one `CoverageIssue` per (source, symlink) match, so hundreds of millions of issues each carrying an `fmt.Sprintf` detail, which `NormalizeInventory` then tries to `sort.Slice`.

The existing budgets (`maxEntries`, `maxMetadataFiles`, `maxMetadataBytes`) bound the walk, not this. Fix: cap patterns per source and in total, cap `len(Coverage)` and emit a single `metadata-budget-exceeded` when hit (the way `secretscan` sets `report.Truncated`), and index package dirs by literal prefix instead of scanning all dirs per pattern. An aggregate byte budget is also missing — 10,000 × 2 MiB is up to 20 GB of reads, with every `workspaces` array retained twice (`:168` and `:170`).

### Worth fixing while you are here

**TOCTOU on the metadata read.** `inventory.go:514-527` is `os.Lstat` then `os.Open` by path, so the "never follows a symlink" invariant is not enforced at open time: anything able to write in the tree during the walk can swap `packages/x/package.json` for a symlink between the two calls. Impact is capped (2 MiB `LimitReader`), but this repo already has the hardened primitive — `secretscan/scanner.go:103-116,183-208` uses `os.OpenRoot` + `root.OpenFile` + `os.SameFile(walkInfo, openedInfo)` before and after the read, and `codeanalysis/analyzer.go` does the same. Go is pinned at 1.26.4, so `os.Root` is available. Please reuse that pattern rather than inventing a weaker one.

**A refused symlinked directory can vanish without coverage.** `inventory.go:116-125` only records `symlink-workspace` when some pattern happens to match, so `tools/shared -> ../../shared` with `workspaces: ["packages/*"]` is dropped with zero coverage even though it may contain real manifests. Emit the issue for every skipped symlinked directory. (Pruning `.git`/`node_modules` silently is correct — those are not first-party.)

**The budgets are untested and untestable.** `maxEntries`, `maxMetadataFiles` and `maxMetadataBytes` are unexported with no injection seam, so `CoverageMetadataBudgetExceeded`, `metadataTooLargeError` and the `fs.SkipAll` paths never execute in CI. `secretscan` solves this with an injectable `scanLimits` (`scanner.go:86-94`) — mirroring that would let you test all of the above, including the new caps.

Also missing: a test for a symlinked `package.json` **file** (only the directory case is covered), for negation-before-inclusion, for a same-indent pnpm sequence, and for a pathological `**` pattern with a time bound.

### Where that leaves the review

Still request-changes. Consolidated must-fix list, roughly in order of how much damage each does:

1. `inventory.go:438` — pnpm same-indent silently reports `Complete = true` with zero workspaces (honesty non-negotiable).
2. `inventory.go:345-365` — memoize `matchSegments`; exponential `**` hang from a ~25-file repo, and `ctx` is never checked inside the matcher.
3. `inventory.go:198-214` — cap patterns and `Coverage`; de-quadratic the source × package × symlink product; add an aggregate byte budget.
4. `normalize.go:52,56` — stop applying Windows path rules to POSIX paths (repo-triggered hard failure and path conflation).
5. `inventory.go:265-284` — order-independent negation.
6. `inventory.go:514-527` — `os.OpenRoot` + `SameFile`, reusing `secretscan`'s walk.
7. Sentinel errors instead of `strings.Contains(err.Error(), ...)` for coverage classification (from my first pass).
8. An injectable limits seam plus the regression tests listed above.

To be clear about my read of the PR overall: the parts that are hardest to get right are right — no YAML library, `Complete` derived in the domain, patterns never used to open anything, real root-escape rejection, deterministic ordering, no exec and no network. The defects above are all in the "bounded and honest" claims rather than in the architecture, and every one has a small, local fix.

VietSory:
Addressed the consolidated 8-item R2A review list on head `d9f94a93fe2a9f77a4927d2877900ecd2a5f14f5`.

1. Standard same-indent `pnpm-workspace.yaml` sequences are now accepted; empty/tab-indented unsupported forms produce incomplete coverage instead of silently reporting `Complete=true`.
2. Workspace glob matching is memoized by `(pi, ti)`, checks `ctx`, collapses consecutive `**`, and is bounded. The hostile alternating-`**` regression must complete before its deadline.
3. Added per-source/total pattern caps, bounded coverage with explicit `metadata-budget-exceeded`, aggregate metadata-byte budget, bounded match work, and literal-prefix candidate indexing. Symlinks are no longer part of the pattern × package matching product.
4. Filesystem-derived POSIX names preserve legal `:` and `\` characters; Windows path interpretation is limited to declared workspace patterns.
5. Workspace negatives are applied as exclusions independently of declaration order.
6. Metadata reads now use `os.OpenRoot`, rooted `OpenFile`, `os.SameFile`, and stable pre/post file metadata checks.
7. Workspace root-escape/absolute-path classification now uses sentinel errors and `errors.Is`.
8. Limits are injectable in tests, with regressions covering every budget plus same-indent pnpm, pathological `**`, negation-before-inclusion, POSIX filenames, symlinked directories/files, and file-identity mismatch.

Validation on the final head is green for:

* `golangci-lint`
* Linux build / vet / test
* CGO-disabled build
* Windows build / vet / test
* focused race tests
* `go test -count=20 -shuffle=on`
* `make build`
* `make vet`
* `make test`
* `make typecheck`

The remaining standard frontend `pnpm test` failure is the separately reproduced baseline Vitest/jsdom/undici worker-start issue; this PR does not modify frontend sources or package metadata.

Ready for re-review. @nghiadaulau

nghiadaulau (approved):
## Strict re-review: all filed blockers verified fixed against the code, not the commit messages

Reviewed at head `bb5a1fc`. I fetched the branch, read the changed lines directly, and ran the suite; I did not trust the commit log. All eight blockers from the two earlier passes are genuinely resolved, and two of the fixes are better than what I asked for.

### Verification of each blocker

1. pnpm same-indent silently reporting `Complete = true` with zero workspaces. Fixed. `parsePnpmWorkspace` now accepts same-indent block sequences (`indent >= baseIndent`), rejects tab indentation and flow style with typed errors, and returns a hard error when `packages:` is present but parses to zero patterns (`pnpm-workspace.yaml packages list is empty or unsupported`). The old `(nil, nil)` silent-clean path for a present-but-unparsed list is gone. `TestInventoryBuilderPNPMMalformedCannotLookComplete` asserts `Complete == false` plus a `malformed-metadata` coverage kind.

2. Exponential `matchSegments` on `**`. Fixed correctly. `matchWorkspacePattern` is now a memoized recursion keyed on `state{pi, ti}`, checks `ctx.Err()` inside `visit`, and `compileWorkspacePatterns` collapses consecutive `**` and caps segment count. Complexity is O(pattern x target), not exponential. The `2s` deadline test on a pathological pattern passes.

3. Uncapped cartesian product and coverage slice. Fixed. `inventoryLimits` bounds `maxPatternsPerSource`, `maxPatternsTotal`, `maxCoverageIssues` (the sink overwrites its last slot with a single budget issue rather than growing), `maxTotalMetadataBytes` for the aggregate byte budget, and a `workBudget` bounds total match work. Each emits a `metadata-budget-exceeded` coverage issue.

4. Windows path rules applied to POSIX filesystem paths. Fixed with the right split. `NormalizeRepositoryLocation` now treats `\` and `:` as ordinary filename bytes, so `c:tmp` and `packages\private` are no longer conflated with Windows syntax; the drive-prefix and absolute-path rules survive only in `normalizeWorkspacePattern`, that is on declared pattern text. Exactly the separation requested.

5. Order-sensitive negation. Fixed, and better than asked. The code now distinguishes npm ordered last-match-wins (`selectOrderedWorkspaceDeclaration`) from pnpm and yarn exclude-always, order-independent (`selectExcludeAlwaysWorkspaceDeclaration`, which evaluates all positives then applies all negatives regardless of position). `["!packages/private/**", "packages/**"]` now excludes correctly.

6. TOCTOU on the metadata read. Fixed with the project's hardened primitive. The read uses `os.OpenRoot` + `root.OpenFile(O_RDONLY|O_NONBLOCK)` + `os.SameFile` before and after, mirroring `secretscan`. A regular-file check rejects a swapped symlink.

7. `strings.Contains(err.Error(), ...)` for coverage classification. Fixed. Gone entirely; classification is by typed sentinels (`errWorkspaceRootEscape`, `errWorkspaceAbsolutePath`) via `errors.Is`.

8. Untestable budgets. Fixed. `inventoryLimits` plus `newInventoryBuilderWithLimits` give an injectable seam, and `inventory_hardening_test.go` exercises every budget with small injected limits.

### Local gate

- `go build ./...` clean, `go vet` clean on both packages.
- `go test -race` green for `jsresolve`, `jsresolution` and `ports`.
- Dependency rule holds: `domain/jsresolution` imports only stdlib and `domain/modulegraph`; `ports` imports only domain packages.
- Tests are adversarial rather than happy-path: malformed pnpm, a deadline-bound hostile matcher, symlink-workspace coverage, and per-budget overflow are all asserted.

### One defect I found and fixed in place

`gofmt -l` flagged two touched files: `jsresolution.go` (the `const` block was left misaligned after adding the longer `CoverageUnsupportedPackageManager` kind) and `inventory_test.go` (one struct-literal key). This is not caught by CI here (the fork PR runs only GitGuardian) nor by `make lint` (golangci-lint is configured `default: standard` plus misspell, revive, gosec, with no gofmt or gofumpt linter), but the repository's formatting standard is gofmt (`make format`). Because the change is purely mechanical alignment, `git diff -w` is empty, I applied `gofmt -w` and pushed it as `bb5a1fc` to keep `main` gofmt-clean rather than bounce the PR on a whitespace nit.

### Two low-severity observations, non-blocking

- `parsePnpmWorkspace` still returns `(nil, nil)` when no `packages:` key is found at all. For a file that genuinely declares no workspaces that is correct; the only gap is an obscure `packages :` with a space before the colon, which pnpm itself would not read either. Fine to leave.
- The branch carries several `chore: trigger CI` marker commits. Squash-merging keeps `main` history clean, which is what I am doing.

### Verdict

Approve. The R2A workspace inventory meets the intent of #400: it produces a deterministic, honest inventory where partial or malformed metadata is always explicit and never rendered as complete, with no exec, no network, root-escape rejection, and bounded, testable budgets. Merging by squash.

Issue #432, opened 2026-08-07
feat(platform): Postgres Row Level Security for multi-tenant isolation (foundation)


nghiadaulau:
Parent: #405. Phase 0, Track A, the #0 critical path item. Supersedes the closed-but-unbuilt #369 and #371.

## Decision context

The tenant decision recorded on #363 (defer, single tenant first) is reversed by an explicit product call: multi tenant is a near term requirement and isolation must be enforced in Postgres now, not only at the application chokepoint. This issue builds that. Every tenant scoped table, existing and new, is protected by Row Level Security so a forgotten `WHERE tenant_id = ...` cannot leak across tenants. RLS becomes a foundation the asset model (#431) and every new fleet table build on, rather than a later hardening pass.

## Why the chokepoint alone is not enough

Tenancy today is application level only: `withEngTenant` plus `TenantFrom(ctx)`, with per store hand written checks. That is exactly the class that produced the CRITICAL and HIGH tenant leaks the hardening program already had to close once. With a fleet writing rows from many agents across tenants, one missed check is a cross tenant read. The database must be the backstop.

## Current state (verified on main at 0056)

- Zero `ROW LEVEL SECURITY` or `CREATE POLICY` statements in `migrations/`. RLS is not present despite #369 and #371 being closed.
- Postgres stores hold a `*pgxpool.Pool` and call `pool.QueryRow(ctx, ...)` directly. There is no per request tenant session variable.
- The tenant is already available in context via `TenantFrom(ctx)`, stamped from the authenticated `Principal.TenantID`.

## Requirements

1. A tenant session mechanism. Provide a helper in the persistence layer, `WithTenant(ctx, pool, fn)`, that opens a transaction, executes `SET LOCAL app.current_tenant = <tenant from ctx>`, runs `fn` against that transaction, and commits or rolls back. Tenant scoped stores use this instead of calling the pool directly. `SET LOCAL` is transaction scoped, so a pooled connection cannot leak one tenant's setting into the next request.
[...]

Issue #431, opened 2026-08-07
feat(fleet): fleet asset model and business service graph (single-tenant, epic-owned)


nghiadaulau:
Parent: #405. Phase 0, Track A. Substrate for the fleet, correlation and cloud pillars. Depends on #432 (Postgres RLS foundation).

## Why this exists as a new issue

Epic #405 needs an asset model for the cluster agent, attack path graph, CSPM, risk story and fleet coverage views. The earlier attempt (#361, part of the #352 series) is closed but its code is not on main: there is no `internal/domain/asset`, no asset migration, and the highest migration on main is `0056`. Closed is not done. Rather than reopen a stale, differently scoped ticket owned by another series, #405 owns a minimal, fit for purpose fleet asset model of its own.

## Tenancy: multi tenant, enforced in Postgres

Per the product decision that reverses #363, this model is multi tenant from day one and isolation is enforced by the database, not only by the application chokepoint. Every table here ships with Row Level Security using the template from #432: `tenant_id` column, `ENABLE` plus `FORCE ROW LEVEL SECURITY`, and a fail closed policy keyed on `current_setting('app.current_tenant', true)`. Stores reach the database through the `WithTenant` helper so the tenant GUC is set per transaction. There is no single tenant interim mode.

## Objective

Define one canonical asset model and a business service grouping that every #405 pillar reads from and writes to, so there is a single identity per real thing (a host, a workload, an image, a cloud resource) and a single place relationships live, isolated per tenant at the database.

## Requirements

1. One asset aggregate with a typed kind: `host`, `workload`, `image`, `cloud_account`, `storage`, `exposure`, `identity`, `namespace`, `cluster`. The kind set is closed and extended deliberately, not by free string.
2. A stable, deterministic asset key per kind so the same real thing observed twice (by image scan and by the cluster agent, for example) resolves to one asset, not two. Image assets key on digest and are shared with existing image scanning.
3. Typed relationships between assets (an edge with a kind and a provenance reference to the observation that produced it). This is the substrate the attack path graph (#419) traverses; it is not the traversal itself.
4. A business service aggregate that groups assets into a named service with an owner, so findings and risk can be rolled up to something a stakeholder recognises.
[...]

Issue #405, opened 2026-08-07
Epic: Synapse as a full platform — central control plane, distributed agents, and depth across SCA, quality, SAST, DAST, red team and blue team


nghiadaulau:
> Revision 3, 2026-08-07. Rewritten with a market comparison, verified current state, explicit decisions, and a phase plan that is split into individually specified child issues. Every pillar is in scope: SCA, governance, code quality, SAST, DAST, red team, blue team, distributed agents, a central control plane, container estates and VM estates.

## 1. Purpose

Build one governed platform that covers both point in time analysis (SCA, SAST, code quality, IaC) and runtime analysis (agents, detections, response), across container orchestrated and VM estates, with a single asset model, a single authorization model, a single evidence chain and a single prioritised queue.

The claim that has to hold for this to be worth building is that each pillar raises the value of the others. A dependency CVE is noise on its own. It becomes work when reachability shows the vulnerable symbol is called, the asset inventory shows the workload is internet facing, the attack path shows credentials reach it, and a runtime detection shows someone is probing it. Assembling that chain is the product. Six separate scanners in one binary is not.

## 2. Market position

Sources are linked inline. The purpose of this section is to state honestly where we are competitive, where we are behind, and what each pillar must reach to be credible.

### 2.1 SCA and reachability

| Vendor | Position |
|---|---|
| Endor Labs | Full call graph across code, dependencies and container images, reachability across 40 or more languages, combined with EPSS for prioritisation ([Endor Labs](https://www.endorlabs.com/learn/best-sca-tools)) |
| Snyk | Reachability limited to Java and JavaScript ([comparison](https://weavai.app/blog/en/2026/05/04/top-10-endor-labs-alternatives-2026-sca-comparison/)) |
| Semgrep Supply Chain | Pattern based reachability hints rather than a call graph |
| Trivy, Grype | Broad ecosystem detection, no reachability |
| Synapse today | Tier 2 call path for Go and JVM, Tier 1 import level for Python, JS and TS in progress under #378. Around 24 owned lockfile parsers. Owned advisory store ingesting OSV, GHSA, CSAF and OVAL. KEV then EPSS then CVSS prioritisation |

[...]

PR #433, opened 2026-08-07
feat(platform): Postgres Row Level Security foundation (#432)

nghiadaulau:
Implements #432 (epic #405): the Postgres Row Level Security foundation for multi-tenant isolation.

## What this delivers
- Migration `0057_row_level_security.sql` (non-breaking, touches no existing table):
  - `synapse_current_tenant()` STABLE function returning `current_setting('app.current_tenant', true)`.
  - `synapse_enable_tenant_rls(tbl)` procedure that applies the identical, non-weakenable policy to any table with a `tenant_id` column: `ENABLE` + `FORCE ROW LEVEL SECURITY` + one `USING`/`WITH CHECK` policy keyed on the function.
- `WithTenant(ctx, pool, tenantID, fn)` helper: runs `fn` in a transaction whose `app.current_tenant` is set via `set_config(..., is_local => true)`, so the setting is transaction-scoped and cannot leak across pooled connections.

## Design decisions
- Fail-closed: an unset tenant resolves to `NULL`, so `tenant_id = synapse_current_tenant()` is `NULL` and excludes every row. This is deny-all even for the default tenant (whose id is the empty string from `0002_default_tenant.sql`); the empty string still matches the default tenant when explicitly set.
- `FORCE ROW LEVEL SECURITY` because goose migrations run as the table owner and the app may connect as that owner; without FORCE the owner is exempt.
- Superuser / BYPASSRLS caveat documented: RLS is bypassed by superusers and BYPASSRLS roles regardless of FORCE. The production runtime DB role must be neither. The integration test proves the mechanism under a dedicated `NOSUPERUSER NOBYPASSRLS` role via `SET LOCAL ROLE`, so it is meaningful even though the dev Postgres connects as a superuser.

## Scope (deliberate)
This PR ships the foundation and proves it end-to-end. It does not flip RLS on any existing production table: those are read today by tenant-less internal paths (`engagements.GetByID` in `usecase/sca`, `recon`, `execution`, `projectuc`) that must move to `WithTenant` first, or they would see zero rows. That retrofit is incremental and tracked per table. #431 (fleet asset model) is the first real table, RLS-native from its first migration.

## Verification
- `go build ./...`, `go vet ./...` clean.
- `go test ./...` (CI-equivalent, no DSN) passes; the RLS integration tests skip like every other Postgres integration test in this repo.
- Against a real Postgres (`SYNAPSE_TEST_DB_DSN`), `TestRLSFoundation` and `TestMigration0057` pass, asserting: isolation via an intentionally unscoped query, fail-closed when unset, `WITH CHECK` blocking a cross-tenant write, `relforcerowsecurity` set, the test role is non-superuser and non-BYPASSRLS, and no residue after `WithTenant`.
- gofmt clean.

Note: the wider postgres integration package has pre-existing batch-ordering fragility (the `goose DownTo` migration tests leave a shared DB down-migrated, breaking sibling tests) that reproduces identically on `main` without this change; it is orthogonal to this PR and not run in CI.

Refs #405

### DIFF
--- a/internal/infrastructure/persistence/postgres/tenant.go
+++ b/internal/infrastructure/persistence/postgres/tenant.go
@@ -3,6 +3,7 @@ package postgres
 import (
 	"context"
 	"fmt"
+	"time"
 
 	"github.com/jackc/pgx/v5"
 	"github.com/jackc/pgx/v5/pgxpool"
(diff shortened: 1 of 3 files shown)


nghiadaulau:
## Review round applied (commit 182aa25)

A security-auditor pass on the first commit found a critical fail-closed gap plus three lesser issues. All are fixed and re-proven against a real Postgres.

- **F1 (critical) fixed.** `app.current_tenant` is a placeholder GUC whose reset value is the empty string, not NULL. After any `WithTenant` transaction a pooled connection reverts to `''`, so a query that bypassed `WithTenant` matched `tenant_id = ''` and leaked default-tenant rows. `synapse_current_tenant()` now returns `NULLIF(current_setting('app.current_tenant', true), '')`, so both unset and the reset empty string collapse to NULL (deny-all). I confirmed empirically on Postgres 17 that the reset value is `''` and that NULLIF maps it to NULL. New invariant: under RLS the empty string is DENY; RLS-protected tables use non-empty tenant ids.
- **F2 (high) addressed.** Added `CheckRLSRuntimeRole`, which errors when the runtime role is SUPERUSER or BYPASSRLS (either silently no-ops all RLS regardless of FORCE). It is intended to gate multi-tenant enablement fail-closed. It is not wired as a hard startup gate in this PR because #432 activates RLS on no existing table and the single-tenant/dev superuser role would otherwise break startup; the hard gate lands with the first live RLS table (#431).
- **F3 (medium) fixed.** The fail-closed test now seeds a `tenant_id=''` row (via superuser, since RLS rejects it), reuses one acquired connection that already ran a `WithTenant` tx, and asserts the reset value is `''`, `synapse_current_tenant()` resolves to NULL, and an unscoped read under the non-privileged role returns zero rows. This is the regression guard that would have been green while F1 was live. Added an empty-tenant-is-DENY case and a `CheckRLSRuntimeRole` test.
- **F4 (low) fixed.** `WithTenant` rolls back on a `committed` flag rather than on `err`, so a panic in `fn` also rolls back instead of returning a connection with an open transaction to the pool.

Arch review nits also applied: rollback uses a fresh bounded context; removed a redundant Go 1.22+ loop copy.

Verification: `go build`, `go vet`, gofmt clean; `go test ./...` (no DSN, CI-equivalent) passes; against a real Postgres `TestRLSFoundation`, `TestCheckRLSRuntimeRole`, `TestMigration0057` pass.

nghiadaulau:
## Re-audit: all four findings resolved

The security-auditor re-reviewed commit 182aa25 and confirmed F1 (critical), F2 (high), F3 (medium) and F4 (low) are all genuinely resolved, no new issues, golden rules clean (no exec/shell, no secrets logged, no LLM in path, `%I`/`$1` parameterization injection-safe).

One follow-through, tracked rather than deferred: the `CheckRLSRuntimeRole` gate is correct and sufficient for this PR because #432 enables RLS on zero tables (a superuser runtime cannot leak through a policy that applies to nothing). But it becomes MANDATORY, wired fail-closed at startup, in the first PR that enables RLS on a real table (#431 fleet asset model), because a superuser/BYPASSRLS runtime would otherwise silently no-op that table's isolation. This is now a hard requirement in #431 (requirement 6) with a `// SECURITY:` marker mandated at the `synapse_enable_tenant_rls` call site so it cannot be silently skipped.

Merging.

Issue #427, opened 2026-08-07
feat(correlation): unified risk story per asset


nghiadaulau:
Parent: #405. Phase 6. Starts when at least two pillars produce data worth correlating.

## Objective

Produce one risk story per asset that assembles every signal about it into a single narrative: what it is, what it exposes, what it depends on, what is wrong with it, whether the wrong thing is reachable, whether it is on an attack path, and whether it has been seen under attack at runtime.

## Why this is the point of the platform

This is the sentence #405 exists to earn. A dependency CVE is noise alone. It becomes work when reachability says the symbol is called, inventory says the workload is internet facing, the attack path says credentials reach it, and a runtime detection says someone is probing it. Every one of those inputs is produced by another issue in this epic. This issue assembles them and refuses to assemble them dishonestly.

## Requirements

1. One canonical view per asset that draws from the asset inventory, the findings of every engine, the reachability verdicts, the attack path graph and the runtime detections. It creates no new data; it correlates existing records by asset identity.
2. Every element in the story references the record that produced it. The story is navigable to its evidence, and an element with no backing record is a defect, not a display choice.
3. Uncertainty is carried through, not flattened. A reachability unknown, an inferred attack path edge and a sampled telemetry window all appear as qualifiers on the relevant element. The story never presents an uncertain input as certain.
4. Prioritisation uses the existing model, KEV then EPSS then CVSS, and is raised by corroborating signals rather than replaced. A CVE that is reachable, on an attack path, and seen under attack ranks above an equal severity CVE with none of those, and the reason is shown.
5. Freshness is part of the story. Each element states when it was last observed, and a stale element is marked stale rather than presented as current, consistent with the fleet coverage rules.
6. Tenant scoped through the existing chokepoint. The story never crosses a tenant boundary.
7. Deterministic assembly. The same underlying records produce the same story in the same order, so it can be diffed over time and a change is meaningful.
8. The story is exportable for a report and for an auditor, carrying its evidence references, because a risk narrative a customer cannot trace is not usable in this project's report path.

## Out of scope

- New scoring inputs beyond what the pillars already produce.
- An LLM in the assembly path. The story is assembled deterministically from records. Prose narration remains the human gated write up draft path, which stays out of the report itself, consistent with the existing rules.

## Specification sketch

```go
// internal/usecase/riskstory
type Story struct {
    AssetID   shared.ID
    Identity  AssetFacts
    Exposure  []ExposureElement   // each references an observation
[...]

Issue #419, opened 2026-08-07
feat(redteam): attack path graph over the asset inventory with provenance per edge


nghiadaulau:
Parent: #405. Phase 4, Track D. Depends on #431 fleet asset model. Design can start immediately.

## Objective

Compute attack paths by joining the asset inventory, exposure facts, reachability verdicts, identity relationships and findings into a graph, so that a statement such as an internet facing service reaching a vulnerable symbol with usable credentials becomes a computed path with evidence rather than an analyst's inference.

## Why this is the correlation payoff

This is the item that makes the whole platform worth more than its parts. Each input already exists or is planned: assets and relationships from #431, exposure and identity facts from the cluster agent, host facts from the VM agent, reachability from the SCA work, and findings from every engine. Nothing here requires a new engine. It requires a graph and a defensible traversal.

## Requirements

1. Node and edge model over the existing asset model. Do not create a second inventory. Nodes are assets, identities, findings and exposures. Edges are typed relationships with a provenance reference to the observation that produced them.
2. Every edge carries provenance and a confidence. An edge derived from a live cluster observation is stronger than an edge inferred from a manifest, and the traversal must be able to distinguish them.
3. Path scoring combines exposure, reachability verdict, finding severity and the risk prioritisation already in use, which is KEV then EPSS then CVSS. The scoring function is documented and deterministic.
4. Unknown propagates. A path that traverses an unknown reachability verdict or an inferred edge is reported with that uncertainty on the path, not silently promoted to a confident path. A path is never rendered as confirmed on the strength of an inferred edge alone.
5. Bounded traversal. Maximum path length, maximum paths returned per target, and a wall clock bound. Hitting a bound is reported, because a truncated path set that looks complete is the failure mode of every graph feature.
6. Determinism. The same graph state produces the same paths in the same order. Path identity is stable across runs so a path can be tracked, accepted or remediated over time.
7. Tenant scoped. A path never crosses a tenant boundary, and the traversal is subject to the existing chokepoint rather than a new query path.
8. Explainability is a requirement, not a feature. Every path renders as an ordered list of steps, each with the observation that supports it, so a customer can dispute a step and an auditor can follow it.
9. The graph is queryable by target, by entry point and by finding, and the API returns paths with their evidence rather than only a score.

## Out of scope

- Executing anything. This issue computes paths. Validating them by exploitation is the next issue.
- A graph database. The existing store is sufficient at the expected scale, and adding one is a separate decision with its own cost.

[...]

Issue #413, opened 2026-08-07
feat(fleet): agent health, per-asset coverage and freshness views


nghiadaulau:
Parent: #405. Phase 1, Track A. Depends on #431 fleet asset model and the agent facing API.

## Objective

Give operators a truthful view of what the fleet covers: agent health, per asset coverage, and data freshness, with unknown, stale, refused and unauthorized shown as distinct states rather than collapsed into clean.

## Why this is a requirement and not a dashboard nicety

The most damaging failure mode of a fleet product is a green dashboard over an estate that is not actually being assessed. This project already treats silent partial coverage as a defect in the scanning path. The same rule has to hold for the fleet, and the fleet makes it harder because absence of data now has several causes: no agent, agent offline, work refused, work expired, authorization missing, or genuinely nothing found.

## Requirements

1. Agent health view: one row per agent with state, last seen, agent version, platform, advertised capabilities and current work. State is derived from last seen against a configurable threshold, and stale is a first class state.
2. Per asset coverage: for each asset and each capability, when it was last assessed, by which agent, under which run, and the coverage verdict. The verdict set is `covered`, `partial`, `stale`, `never`, `refused`, `unauthorized`, `agent_missing`.
3. Freshness is measured against a per capability target defined by policy. An asset whose last assessment is older than the target is stale and is surfaced, not hidden.
4. No default to clean anywhere. An asset with no agent reports `agent_missing`. A capability an agent refused reports `refused` with the reason. Neither is ever rendered as a passing state.
5. Fleet summary metrics: agents by state, assets by coverage verdict, oldest assessment per capability, and the count of assets with no agent. Exposed both in the API and on the dashboard.
6. Tenant scoped throughout. The existing tenant chokepoint governs every read, and there is no cross tenant aggregate view for a non administrative principal.
7. The web views follow the existing design system: theme tokens only, lucide icons only, monospace with tabular numbers for timestamps and counts, and explicit loading, empty and error states. Tables above fifty rows are virtualised.
8. Coverage is exportable, because an auditor asking what was and was not assessed is a first class use case for this project.

## Out of scope

- Alerting and notification on coverage regression.
- Trend history over time. Point in time truth first.

## Specification sketch

```
GET /api/v1/fleet/agents                      list with state filter
GET /api/v1/fleet/agents/{id}                 detail with recent work orders
GET /api/v1/fleet/coverage                    per asset per capability verdicts
[...]

Issue #411, opened 2026-08-07
feat(fleet): Kubernetes cluster agent for workload, exposure and identity inventory


nghiadaulau:
Parent: #405. Phase 1, Track A. Depends on the agent facing API and on #431 fleet asset model.

## Objective

Ship a cluster agent for Kubernetes that inventories workloads, reports image posture for what is actually running, and evaluates cluster configuration, reusing the existing Kubernetes rule pack.

## Why this is distinct from the VM agent

A cluster is not a set of hosts. The unit of inventory is a workload and its controller, the unit of exposure is a service and an ingress, and the identity model is a service account with role bindings. A host agent on every node would report containers without the orchestration context that makes them meaningful.

## Requirements

1. Workload inventory: namespaces, controllers, pods, containers, and the resolved image digest for every running container. The digest matters because a mutable tag tells you nothing about what is running.
2. Correlate running digests with images already scanned by the engine. A running digest that has never been scanned is a coverage gap and must be reported as such, not omitted.
3. Cluster configuration evaluation using the existing Kubernetes rules, applied to live objects rather than to manifest files.
4. Exposure facts: services by type, ingress hosts and paths, network policies present or absent per namespace. This is the input the attack path work needs to distinguish internet facing from internal.
5. Identity facts: service accounts, roles, cluster roles and bindings, plus which workloads mount which service account. This is the input for privilege escalation path analysis.
6. Least privilege. The agent requests a read only cluster role scoped to the resources listed above and nothing more. The exact role is shipped in the deployment manifest and documented.
7. Watch or poll with bounded resources. Resync interval, page size and memory ceiling are configurable, and a large cluster must not cause unbounded memory growth.
8. Namespace scoping. An operator can restrict the agent to a namespace list, and anything outside that list is reported as out of scope rather than as clean.
9. Multi cluster is a first class concept from the start. Cluster identity is part of the asset key so two clusters with identical namespace and workload names never collide.
10. Report as inventory into the #431 asset model. Do not create a parallel workload table.

## Out of scope

- Admission control and blocking. This agent observes. Blocking is a separate decision with a separate blast radius.
- Runtime detections, which are Phase 5.
- Service mesh specifics.

## Specification sketch

Asset mapping into the #431 model:

| Kubernetes object | Asset kind | Key |
|---|---|---|
| Cluster | `cluster` | cluster identity |
| Namespace | `namespace` | cluster + name |
| Controller | `workload` | cluster + namespace + kind + name |
[...]

PR #434, opened 2026-08-07
feat(fleet): fleet asset model, multi-tenant and RLS-enforced (#431)

nghiadaulau:
Implements #431 (epic #405): the fleet asset model, multi-tenant and RLS-enforced. First real adopter of the #432 Row Level Security foundation.

## What this delivers
- **Domain** `internal/domain/asset`: pure aggregate. Closed `Kind` set (host, workload, image, cloud_account, storage, exposure, identity, namespace, cluster), closed `EdgeKind` set, an `Edge` whose `Provenance` is required, and a `BusinessService`. Every aggregate requires a NON-EMPTY tenant id (under RLS the empty string is DENY, per #432).
- **Port** `ports.AssetRepository`; **usecase** `assetuc.Service` audits every mutation and makes upserts idempotent by natural key (re-observation reuses the id, no churn).
- **Stores**: memory + Postgres. The Postgres store is the first to route every operation through `WithTenant`, so RLS enforces isolation at the database.
- **Migration** `0058_fleet_assets.sql`: `fleet_assets`, `fleet_asset_edges`, `fleet_business_services`, each `CALL synapse_enable_tenant_rls(...)`.
- **HTTP**: `/api/v1/assets`, `/assets/edges`, `/assets/services` behind the `authz` RBAC chokepoint, tenant from context; optional, registered only when enabled.
- **Config** `SYNAPSE_FLEET_ASSETS_ENABLED` (default off).

## SECURITY: the mandatory RLS runtime-role gate (#431 req 6)
RLS is silently a no-op if the runtime DB role is SUPERUSER or has BYPASSRLS, regardless of FORCE. This PR wires `CheckRLSRuntimeRole` fail-closed: when the asset model is enabled with Postgres, startup refuses to serve (os.Exit) if the role cannot enforce RLS. A `// SECURITY:` marker sits at the `synapse_enable_tenant_rls` call site and in the composition root. This is the follow-through the #432 re-audit required before the first table went live.

## Verification
- `go build ./...`, `go vet ./...`, gofmt clean.
- `go test ./...` (CI-equivalent, no DSN) passes; the integration tests skip like every other Postgres integration test.
- Against a real Postgres (`SYNAPSE_TEST_DB_DSN`):
  - `TestAssetRepository`: store roundtrip, idempotent upsert (one row, stable id), query-level tenant scoping, edge and business-service roundtrip + idempotency, ErrNotFound.
  - `TestFleetAssetsRLS`: FORCE set on all three tables; DB-level isolation proven via an INTENTIONALLY UNSCOPED `SELECT` under a `NOSUPERUSER NOBYPASSRLS` role (tenant A sees only its rows, tenant B only its own, zero when unset).
  - `TestMigration0058`: migration down to 57 (table gone) and up to 58 (table back).
  - Domain, usecase (memory-backed), and route-presence tests.

## Scope
Consumers (attack path #419, cluster/cloud producers #411/#429, risk story #427, fleet coverage #413) are separate issues that read/write this model. This PR is the model plus its governed write/read surface.

Refs #405

### DIFF
--- a/cmd/synapse-api/main.go
+++ b/cmd/synapse-api/main.go
@@ -92,6 +92,7 @@ import (
 	"github.com/KKloudTarus/synapse-ce/internal/usecase/agenttools"
 	analysisuc "github.com/KKloudTarus/synapse-ce/internal/usecase/analysis"
 	"github.com/KKloudTarus/synapse-ce/internal/usecase/approval"
+	"github.com/KKloudTarus/synapse-ce/internal/usecase/assetuc"
 	audituc "github.com/KKloudTarus/synapse-ce/internal/usecase/audit"
 	aupuc "github.com/KKloudTarus/synapse-ce/internal/usecase/aup"
 	"github.com/KKloudTarus/synapse-ce/internal/usecase/codequality"
@@ -166,6 +167,7 @@ func main() {
 	// Persistence: PostgreSQL when configured, else file + in-memory (dev).
 	var repo ports.EngagementRepository
 	var projectRepo ports.ProjectRepository
+	var assetStore ports.AssetRepository
 	var findingRepo ports.FindingRepository
 	var judgmentStore analysisuc.Store // postgres or memory; satisfies both the narrow Store + ports.JudgmentStore
 	var commentRepo ports.CommentRepository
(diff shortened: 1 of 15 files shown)


nghiadaulau:
## Review round applied (commit ca7092c)

Security-auditor: SOUND on tenant isolation, no critical/high. go-arch-reviewer: PASS on the dependency rule. Both raised items, now addressed.

- **Arch MAJOR (fixed).** With the feature enabled on a single-tenant/default deployment, `TenantFrom` returns `""` but the domain rejects an empty tenant (empty = DENY under RLS), so every asset write returned 400 and the feature was unusable. The handler now maps the empty default tenant to a non-empty `DefaultFleetTenant` (`"default"`), and migration 0058 seeds that tenant row so the FK is satisfied. The feature works in single-tenant mode while staying RLS-isolated; multi-tenant principals use their real tenant id. The domain doc-comment is now accurate.
- **Auditor Low + Arch minor (fixed): referential integrity.** `fleet_asset_edges` now has a COMPOSITE FK `(tenant_id, from_asset)`/`(tenant_id, to_asset)` -> `fleet_assets(tenant_id, id)` (with a `UNIQUE (tenant_id, id)` target). FK checks bypass RLS in Postgres, so the `tenant_id` in the FK is what rejects a cross-tenant edge, and it also rejects dangling edges. The store test, which had crossed tenants in an edge, is corrected.
- **Arch minors (fixed).** `UpsertEdge` captures `clock.Now()` once; the list endpoints return `[]` rather than `null`.

Auditor informational notes acknowledged and left as-is: audit is a separate append-only store (matches the existing codebase pattern), and empty-tenant = DENY is intended fail-closed behaviour (now handled gracefully by the default-tenant mapping).

Verification after fixes: all six Postgres integration tests pass (`TestAssetRepository`, `TestFleetAssetsRLS`, `TestMigration0058`, plus the three #432 RLS tests), `go test ./...` (no DSN) passes, `go build`/`go vet`/gofmt clean.

Issue #409, opened 2026-08-07
feat(fleet): agent facing API for enrol, heartbeat, work claim and result ingest


nghiadaulau:
Parent: #405. Phase 1, Track A. Depends on agent identity and on the work order model.

## Objective

Define and implement the transport between an agent and the control plane: enrol, heartbeat, claim work, submit results, upload evidence. Bounded, versioned, rate limited and fail closed.

## Context

`agentDeps` in the current router is the AI orchestration agent, not a fleet agent. There is no fleet transport today. This issue creates one, under a separate route prefix so the two are never confused.

## Requirements

1. Route prefix `/api/v1/fleet` and a required protocol version in the request. A version the server does not support is refused with a typed error rather than best effort handling.
2. Endpoints, minimum set:
   - `POST /enrol` exchanges an enrolment token and a certificate signing request for a client certificate.
   - `POST /heartbeat` reports liveness, agent version, platform, capability set and resource usage. Returns the assigned configuration generation and whether an update is available.
   - `POST /work/claim` claims at most N addressed work orders. Returns signed orders.
   - `POST /work/{id}/progress` reports running state and extends the lease.
   - `POST /work/{id}/result` submits the result. Idempotent by idempotency key.
   - `POST /evidence` uploads a bounded artifact and returns a reference for use in a result.
3. Every request except enrol requires mutual TLS. Authorization is evaluated server side on every call. The agent never sends a tenant identifier that the server trusts; the tenant comes from the certificate.
4. Explicit bounds on everything: maximum request body size, maximum orders per claim, maximum evidence artifact size, maximum concurrent uploads per agent. Exceeding a bound returns a typed error, never a partial success.
5. Per agent and per tenant rate limits. An agent that exceeds its limit is throttled with a retry hint and the event is recorded.
6. Fail closed. If authorization, tenancy or scope cannot be evaluated, the request is refused. An unknown capability in a result is refused rather than stored as unknown.
7. Backpressure is explicit. When the control plane cannot accept work, the response tells the agent to back off with a duration, and the agent honours it.
8. Version skew is handled in both directions. An agent older than the floor is told to update and is refused work. An agent newer than the server keeps working with the fields the server understands and the server never silently drops a field it does not recognise; it refuses.
9. The API is documented in `docs/` with a request and response example for every endpoint, and the error taxonomy is listed.

## Out of scope

- Streaming telemetry, which belongs to Phase 5 and needs the storage decision first.
[...]

Issue #408, opened 2026-08-07
feat(fleet): agent identity, enrolment and revocation


nghiadaulau:
Parent: #405. Phase 1, Track A. Depends on #432 Postgres RLS foundation and on the work order model.

## Objective

Give every agent a cryptographic identity, a controlled enrolment path, and a revocation path, so that agent actions are attributable in the audit trail and an agent cannot act outside its tenant.

## Requirements

1. Enrolment is initiated with a short lived, single use, tenant scoped enrolment token issued by an operator holding an administrative permission. The token is displayed once and stored only as a hash.
2. The agent generates a private key locally and sends a certificate signing request. The private key never leaves the host and is never transmitted.
3. The control plane issues a client certificate whose subject encodes the agent identity and the tenant identity. Certificate lifetime is bounded and renewal is automatic before expiry.
4. Every agent facing call authenticates with mutual TLS. A request whose certificate tenant does not match the target resource tenant is rejected before any handler runs, through the existing tenant chokepoint rather than a new path.
5. Revocation is immediate. A revoked agent is refused at the transport layer, its in flight work orders are cancelled, and the revocation is audited with the operator identity and reason.
6. An agent record carries the platform, the operating system version, the agent version, the enrolled timestamp, the last seen timestamp and the capability set the agent advertises.
7. Agent identity appears as the actor in the audit log for every action the agent performs. Existing audit entries that assume a human principal must keep working.
8. Enrolment token issuance, use, expiry and failure are all audited. A failed enrolment attempt is recorded with the source address and no token material.
9. Rate limit enrolment per tenant and lock out after repeated failures, so a leaked token cannot be brute forced for a wider window than its lifetime.

## Out of scope

- The work order transport, specified in the agent facing API issue.
- Any human authentication change. OIDC and SSO stay where they are.
- Certificate transparency or an external PKI. A control plane issued intermediate is sufficient.

## Specification sketch

```go
// internal/domain/fleetagent
type Agent struct {
    ID           shared.ID
    TenantID     shared.ID
    Name         string
    Platform     string   // linux, windows, darwin
    OSVersion    string
    AgentVersion string
    Capabilities []string
    Fingerprint  string   // certificate SHA-256
    EnrolledAt   time.Time
    LastSeenAt   time.Time
    State        State    // pending, active, stale, revoked
    RevokedAt    *time.Time
    RevokedBy    shared.ID
    RevokeReason string
}
```

[...]

Issue #407, opened 2026-08-07
feat(platform): fleet work order model for agent addressed, signed, idempotent work


nghiadaulau:
Parent: #405. Phase 0, Track A, critical path. Depends on leader election.

## Objective

Extend the existing lease based job model into a fleet work order model: units of work addressed to a specific agent identity, signed by the control plane, idempotent, and auditable end to end.

## Why this is needed

Today a job is claimed by whichever worker polls first, which is correct for a control plane owned worker pool. An agent fleet is different: work is targeted at an asset that only one agent can reach, agents are untrusted, and the control plane must be able to prove later which agent was asked to do what, under which authorization.

## Requirements

1. A work order references the asset, the requested capability, the engagement or assessment context that authorises it, an expiry, and an idempotency key.
2. Work orders are addressed. An agent may only claim orders addressed to its own identity or to a group it belongs to. The API rejects any attempt to claim another agent's order.
3. Work orders are signed by the control plane. The agent verifies the signature before acting, and the signature covers the capability, the target, the authorization window and the expiry.
4. Authorization is evaluated at issue time and again at claim time. An expired authorization window means the order cannot be claimed even if it was already issued.
5. At most one in flight order per (asset, capability, time bucket). Duplicate scheduler ticks must not produce two orders for the same work.
6. Lifecycle states are explicit and persisted: issued, claimed, running, succeeded, failed, expired, cancelled, refused. Refused is distinct from failed and records the reason, for example a capability the agent does not support.
7. Every transition is audited with the actor, which for an agent is the agent identity rather than a human principal.
8. Result submission is idempotent by idempotency key. A retried submission does not create a second run record.
9. Orders carry a bounded payload. Anything larger travels as an evidence upload referenced by the order.

## Out of scope

- The transport, which is specified in the agent facing API issue.
- Enrolment and identity, specified in its own issue.
- Any scheduling policy beyond what already exists for recurring scans.

## Specification sketch

```go
// internal/domain/workorder
type State string

const (
    StateIssued    State = "issued"
    StateClaimed   State = "claimed"
    StateRunning   State = "running"
    StateSucceeded State = "succeeded"
    StateFailed    State = "failed"
    StateExpired   State = "expired"
    StateCancelled State = "cancelled"
    StateRefused   State = "refused"
)

type WorkOrder struct {
    ID             shared.ID
    TenantID       shared.ID
    AssetID        shared.ID
    AgentID        shared.ID   // addressed recipient
[...]

PR #435, opened 2026-08-07
feat(fleet): work order model, addressed, signed and RLS-enforced (#407)

nghiadaulau:
Implements #407 (epic #405): the fleet work order model. An order addressed to a specific agent, authorised by an engagement, signed by the control plane, driven through an explicit state machine, and isolated per tenant by Row Level Security (building on #432/#431).

## What this delivers
- **Domain** `internal/domain/workorder`: closed `State` set + a closed transition graph (`CanTransition`), a `New` constructor requiring a non-empty tenant (empty = DENY under RLS) plus asset/agent/capability/authorization/idempotency-key/expiry, and a deterministic `SigningPayload` over the authorising fields so a tampered field invalidates the signature.
- **Signer** `internal/platform/worksign`: HMAC-SHA256, constant-time verify; the key stays behind `ports.WorkOrderSigner`.
- **Port** `ports.WorkOrderStore` + **usecase** `internal/usecase/fleetwork`: issue (signs; idempotent by `(tenant, idempotency_key)`; rejects a second live order for the same `(tenant, asset, capability, bucket)` with `ErrConflict`), claim (atomic, addressed, unexpired), transition (validates the transition, then CAS on the expected state), all audited.
- **Stores**: memory + Postgres. Postgres routes every method through `WithTenant` so RLS isolates at the database; claim uses `FOR UPDATE SKIP LOCKED`; issue checks idempotency before insert so a re-issue never collides with the in-flight index.
- **Migration** `0059_work_orders.sql`: unique `(tenant, idempotency_key)`, a PARTIAL unique in-flight index over `(tenant, asset, capability, time_bucket) WHERE state IN ('issued','claimed','running')`, and RLS via `CALL synapse_enable_tenant_rls`.

## Scope
`asset_id`, `agent_id` and `authorization_id` are logical cross-aggregate references validated at the app layer, not DB FKs, because the fleet agent table (#408) does not exist yet. The HTTP surface and composition-root wiring land with the agent-facing API (#409); this PR delivers the tested model, stores and service (the migration applies at startup, so the schema is live).

## Verification
- Domain state-machine + signing tests; usecase tests (idempotency, in-flight conflict, addressed claim, transition rules, sign/verify/tamper).
- Postgres integration: roundtrip, idempotent re-issue, in-flight `ErrConflict`, addressed atomic claim, CAS transition (stale expected -> conflict), `FORCE` RLS on the table, migration down/up.
- `go build ./...`, `go vet ./...`, `go test ./...` (no DSN, CI-equivalent), gofmt all clean.

Refs #405

### DIFF
--- a/internal/domain/workorder/workorder.go
+++ b/internal/domain/workorder/workorder.go
@@ -1,0 +1,140 @@
+// Package workorder is the epic-#405 fleet work order model: a unit of work addressed to a
+// specific agent identity, authorised by an engagement, signed by the control plane, and driven
+// through an explicit state machine. It is pure domain: it imports only shared and the stdlib.
+//
+// The signing itself lives outside the domain (a platform signer holds the key); the domain only
+// defines the canonical, deterministic payload that is signed, so the authorising fields cannot
+// drift between issue and verify.
+package workorder
+
+import (
+	"fmt"
+	"strings"
+	"time"
+
+	"github.com/KKloudTarus/synapse-ce/internal/domain/shared"
+)
+
+// State is the lifecycle of a work order. Terminal states never transition again.
+type State string
+
+const (
+	StateIssued    State = "issued"
+	StateClaimed   State = "claimed"
+	StateRunning   State = "running"
+	StateSucceeded State = "succeeded"
+	StateFailed    State = "failed"
+	StateExpired   State = "expired"
+	StateCancelled State = "cancelled"
+	StateRefused   State = "refused"
+)
+
+// Valid reports whether s is a known state.
+func (s State) Valid() bool {
+	switch s {
+	case StateIssued, StateClaimed, StateRunning, StateSucceeded, StateFailed,
+		StateExpired, StateCancelled, StateRefused:
+		return true
+	default:
+		return false
+	}
+}
+
+// Terminal reports whether s is a final state.
+func (s State) Terminal() bool {
+	switch s {
+	case StateSucceeded, StateFailed, StateExpired, StateCancelled, StateRefused:
+		return true
+	default:
+		return false
+	}
+}
+
+// allowedTransitions is the closed transition graph. Anything not listed is rejected.
+var allowedTransitions = map[State]map[State]bool{
+	StateIssued:  {StateClaimed: true, StateExpired: true, StateCancelled: true},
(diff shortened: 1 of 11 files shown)


nghiadaulau:
## Review round applied (commit 8688f1d)

Security-auditor: no critical/high; tenant isolation, signing payload, claim locking and transition CAS all sound. go-arch-reviewer: dependency rule PASS, no blockers/majors. Both raised items, now addressed.

- **Security MEDIUM (fixed).** `worksign.New` now returns `(*Signer, error)` and rejects a key shorter than 32 bytes (`ErrWeakKey`), so a missing/misconfigured HMAC key cannot boot a forgeable signer. Added worksign unit tests: weak-key rejection, sign/verify, tampered signature, and key-copy.
- **Security LOW (fixed).** The `Issue` 23505 handler no longer depends on `CREATE INDEX` OID ordering. On any unique violation it re-fetches by `(tenant, idempotency_key)` and returns the existing order if present, returning `ErrConflict` only for a genuinely different order colliding on the in-flight slot. So a concurrent idempotent retry stays idempotent regardless of index order.
- **Arch minor (fixed): store parity.** `Transition` now threads the service clock (port signature + both stores + service) instead of the Postgres store using SQL `now()`, so `updated_at` matches across backends; and the Postgres store distinguishes a missing order (`ErrNotFound`) from a CAS state mismatch (`ErrConflict`), matching the memory store. Added a not-found transition test.
- **Arch minor (fixed).** `Claim` guards `max <= 0` (returns empty) so a non-positive limit cannot diverge between the Postgres `LIMIT` and the memory store. Added an explicit compile-time assertion that `*worksign.Signer` satisfies `ports.WorkOrderSigner`.

Informational notes acknowledged and left as-is: audit-after-commit matches the existing codebase pattern, and expired-but-issued orders holding the in-flight slot is a liveness edge (a sweeper to `expired` is a later concern; a new time bucket sidesteps it). Signature coverage confirmed appropriate: the payload covers every field that defines what the agent is authorized to do; lifecycle/scheduling fields are correctly mutable and server-side CAS-guarded.

Verification after fixes: all unit and Postgres integration tests pass (work order roundtrip, idempotent re-issue, in-flight conflict, addressed atomic claim, CAS transition incl. not-found, FORCE RLS, migration down/up; plus the #432/#431 RLS tests), `go test ./...` (no DSN) passes, `go build`/`go vet`/gofmt clean.

Issue #412, opened 2026-08-07
feat(fleet): agent packaging, signing, update channel and fail-closed release scanning


nghiadaulau:
Parent: #405. Phase 1, Track A. Depends on the VM agent existing as a binary.

## Objective

Build the release engineering required to run an agent on hosts the project does not own: a declared support matrix, signed native packages, a signed update channel, safe rollback, and release scanning that fails closed.

## Why this is a full issue and not a bullet

`.github/workflows/release.yml` currently runs goreleaser to produce binaries and a container image. There is no nfpm configuration, no rpm, no deb, no msi, no code signing and no update channel. Shipping software onto other people's production hosts is a different discipline from shipping a container image.

The requirements below are derived from failure modes observed while assessing a third party agent fleet in the week to 2026-08-07, so they are not hypothetical:

1. An RPM built against a newer glibc installed cleanly on RHEL 7 and RHEL 8 and then crash looped at first start, because nothing in the package expressed the runtime floor and the failure surfaced only after `systemctl enable`.
2. The Windows MSI and EXE shipped with no Authenticode signature while an extended validation certificate was still pending, so every install produced a SmartScreen warning.
3. The release scan job reported zero findings for every release because the scanner binary was missing and the invocation was wrapped in `|| true`.

Each of those maps to a requirement below.

## Requirements

1. A written support matrix: distribution, minimum version, architecture, libc floor, service manager. Anything outside the matrix is refused by the installer with a clear message rather than installed and left to fail at start.
2. The libc floor is enforced twice. It is declared as a package dependency so the package manager refuses the install, and it is checked by a preinstall script for the case where the dependency is satisfiable but the runtime is older than the build.
3. Native packages: rpm and deb via nfpm, msi for Windows, plus a plain archive. Each package installs a service unit, a configuration file with restrictive permissions, and a dedicated service account.
4. All artifacts are signed. Rpm and deb are signed with a project key whose public key is published. Windows artifacts are Authenticode signed. A release cannot be published with an unsigned artifact; the pipeline fails.
5. Reproducible and attested. Every release publishes checksums, an SBOM per artifact, and a provenance attestation. The SBOM is generated by this project's own engine, which also serves as a dogfooding gate.
6. Release scanning fails closed. The release job scans every artifact and the container image, and the job fails if the scanner is missing, errors, or reports a finding above the configured floor. No invocation in the release path may be wrapped in a construct that swallows a non zero exit.
[...]

PR #436, opened 2026-08-07
feat(fleet): agent-facing API with agent identity and auth (#409)

nghiadaulau:
Implements #409 (epic #405): the agent-facing transport, on a SEPARATE authentication plane from human RBAC.

## Scope decision (please read)
#409 cannot be safe without an agent identity/auth layer, and #408 (agent identity, mTLS) is not built. So this PR delivers the minimal **token-based** agent identity #409 requires, and defers the heavier pieces with references:
- Mutual TLS + CSR enrolment and revocation-at-the-transport-layer -> #408.
- Evidence upload and the update channel -> #412.
- Backpressure (503) -> follow-up.

The credential is a bearer token, not a client certificate, but the model is built so a certificate fingerprint can replace the token hash without changing the lifecycle.

## What this delivers
- **Domain** `internal/domain/fleetagent`: `Agent` + single-use `EnrolToken`; non-empty tenant required (empty = DENY under RLS); only credential hashes stored.
- **Credential** `internal/platform/agenttoken`: `"<kind>.<b64(tenant)>.<b64(id)>.<secret>"`. The non-secret tenant/id prefix routes the RLS-scoped lookup before verification; the 32-byte secret is stored only as a SHA-256 hash and compared in **constant time**. This resolves the auth chicken-and-egg (you must know the tenant to do an RLS read, but you learn it from the credential) without any cross-tenant scan; forging the prefix fails the hash compare.
- **Usecase** `internal/usecase/fleetagentuc`: MintEnrolToken (operator), Enrol (tenant from the token, never the caller), Authenticate (`ErrUnauthenticated`/`ErrRevoked`), Heartbeat, Revoke. Work orders reuse `fleetwork` (#407).
- **Stores**: memory + Postgres (`WithTenant`; migration 0060 RLS-enables `fleet_agents` + `fleet_enrol_tokens`; `ConsumeEnrolToken` is an atomic single-use `UPDATE ... RETURNING`).
- **Adapter**: `/api/v1/fleet` is mounted at the top level, **bypassing the human authenticator and AUP gate**. Its own middleware requires the protocol version header and a valid agent bearer credential (enrol excepted), rate-limits per agent, and stamps the agent + tenant into context. Endpoints: enrol, heartbeat, work/claim (addressed), work/{id}/progress, work/{id}/result (idempotent). A mis-addressed or cross-tenant order returns 404 (no existence leak). Operator routes `/api/v1/agents` (mint enrolment token, list, revoke) stay on the human RBAC plane (PermAdminister / PermView).
- **Config** `SYNAPSE_FLEET_ENABLED` (default off) + `SYNAPSE_FLEET_SIGNER_KEY`. When enabled with Postgres, startup fails closed unless the DB role can enforce RLS and the signer key is >= 32 bytes.

## Verification
- Domain + usecase tests: single-use enrol, tampered/forged/cross-tenant/revoked/expired rejection, constant-time compare.
- Postgres integration: single-use consume, agent CRUD, FORCE RLS on both tables, migration 0060 down/up.
- HTTP end-to-end: version required (400), enrol (201), auth required (401), addressed claim returns only the caller's order, progress->result idempotent (200 on repeat), mis-addressed order 404, revoked agent 403.
- `go build ./...`, `go vet ./...`, `go test ./...` (no DSN), gofmt all clean.

Refs #405

### DIFF
--- a/cmd/synapse-api/main.go
+++ b/cmd/synapse-api/main.go
@@ -89,6 +89,7 @@ import (
 	"github.com/KKloudTarus/synapse-ce/internal/platform/idgen"
 	"github.com/KKloudTarus/synapse-ce/internal/platform/jobs"
 	"github.com/KKloudTarus/synapse-ce/internal/platform/logging"
+	"github.com/KKloudTarus/synapse-ce/internal/platform/worksign"
 	"github.com/KKloudTarus/synapse-ce/internal/usecase/agenttools"
 	analysisuc "github.com/KKloudTarus/synapse-ce/internal/usecase/analysis"
 	"github.com/KKloudTarus/synapse-ce/internal/usecase/approval"
@@ -108,6 +109,8 @@ import (
 	exploitationuc "github.com/KKloudTarus/synapse-ce/internal/usecase/exploitation"
 	exportuc "github.com/KKloudTarus/synapse-ce/internal/usecase/export"
 	findingsuc "github.com/KKloudTarus/synapse-ce/internal/usecase/findings"
+	"github.com/KKloudTarus/synapse-ce/internal/usecase/fleetagentuc"
+	"github.com/KKloudTarus/synapse-ce/internal/usecase/fleetwork"
 	"github.com/KKloudTarus/synapse-ce/internal/usecase/fptriage"
 	"github.com/KKloudTarus/synapse-ce/internal/usecase/llmverifier"
 	"github.com/KKloudTarus/synapse-ce/internal/usecase/orchestrator"
@@ -168,6 +171,8 @@ func main() {
 	var repo ports.EngagementRepository
 	var projectRepo ports.ProjectRepository
 	var assetStore ports.AssetRepository
+	var workOrderStore ports.WorkOrderStore
+	var fleetAgentStore ports.FleetAgentStore
 	var findingRepo ports.FindingRepository
 	var judgmentStore analysisuc.Store // postgres or memory; satisfies both the narrow Store + ports.JudgmentStore
 	var commentRepo ports.CommentRepository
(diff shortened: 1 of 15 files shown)


nghiadaulau:
## Review round applied (commit c2cb65f)

Security-auditor: the credential design and tenant isolation are sound (secret is 32 bytes from crypto/rand, only the SHA-256 hash stored, constant-time compare, forged prefix cannot authenticate, auth-plane separation correct, single-use enrol atomic with no TOCTOU, every store method through WithTenant/RLS, fail-closed on weak signer key and non-RLS role). go-arch-reviewer: dependency rule PASS. Three findings addressed:

- **Security MEDIUM (fixed).** `listFleetAgents` serialized `*fleetagent.Agent` directly, which (no JSON tags) emitted `TokenHash` — the credential-verifier material — over the operator API. It now maps to an `agentView` DTO that omits `TokenHash`; no handler serializes the domain agent anymore.
- **Security LOW (fixed).** Rate limiting was post-authentication only, so `enrol` and failed-auth attempts drove unbounded DB lookups on the internet-facing plane. A per-client-IP limiter now runs in a new `entry()` wrapper BEFORE any DB work on every fleet route (including enrol), keyed on `RemoteAddr` (a throttling key, not an authorization input). The per-agent limiter still applies post-auth.
- **Security LOW / arch minor (fixed).** The limiter map is now bounded: `keyedLimiter` prunes expired windows past a cap, so the untrusted IP key space cannot grow it without bound. (Per-process limiter under horizontal scale is noted for P5.)
- **Arch minors (fixed).** `mintEnrolToken` uses `fleetBodyCap`; `Handler()` no longer double-wraps `normalizePath`.

Informational, left as noted: the in-memory stores ignore `ctx` (established repo pattern for dev/test stores); the per-process limiter becomes a shared-limiter concern at horizontal scale (P5).

Verification after fixes: HTTP end-to-end (version, enrol, auth-required, addressed claim, progress->result idempotent, mis-addressed 404, revoke->403) and Postgres integration (single-use consume, agent CRUD, FORCE RLS, migration 0060 down/up) pass; `go build`/`go vet`/`go test ./...` (no DSN)/gofmt clean.

Issue #406, opened 2026-08-07
feat(platform): leader election to replace the single process scheduler lock


nghiadaulau:
Parent: #405. Phase 0, Track A, critical path.

## Objective

Replace the single process `AcquireSingletonLock` with leader election so more than one `synapse-api` or `synapse-worker` instance can run concurrently without duplicating scheduled work.

## Why this blocks the fleet

A fleet of agents needs a control plane that survives an instance restart and scales horizontally. Today the singleton lock permits exactly one scheduler, so a rolling deploy either pauses scheduling or risks two schedulers dispatching the same work. Once agents claim work orders, a duplicate dispatch means the same target is scanned twice under two run identities, which corrupts coverage and freshness accounting.

## Current state

- `internal/usecase/worker` is already lease based: per claim visibility, a heartbeat that extends the lease while a handler runs, and backoff on failure.
- Job state is behind `ports.ScanJobStore` and run history behind `ports.ScanRunStore`.
- The singleton lock is the remaining single process assumption.

## Requirements

1. Introduce a leader election port in `internal/usecase/ports` with a Postgres implementation. Advisory locks or a leases table are both acceptable; the implementation must be documented.
2. Exactly one instance holds leadership at a time. Leadership has a bounded term and must be renewed. A holder that fails to renew loses leadership without operator action.
3. Only the leader runs periodic or scheduled dispatch. Followers continue to serve HTTP and continue to claim already queued jobs, because job claiming is already safe through leases.
4. Leadership transitions are recorded in the append only audit log with the instance identity and the reason for the transition.
5. Loss of leadership mid dispatch must not leave a job in a state where it can be dispatched twice. Dispatch is idempotent by job identity.
6. A clean shutdown releases leadership immediately rather than waiting for the term to expire.
7. Configuration through `SYNAPSE_LEADER_*` environment variables with defaults that are safe for a single instance deployment, so existing single instance operators see no behaviour change.

## Out of scope

- Multi region or cross cluster coordination.
- Replacing the lease based job claiming, which already works.
- Any change to the agent facing API, which does not exist yet.

## Specification sketch

```go
// internal/usecase/ports
type LeaderElector interface {
    // Campaign blocks until leadership is acquired or ctx is done.
    Campaign(ctx context.Context, instanceID string) error
    // Renew extends the current term. Returns ErrLeadershipLost if the term expired.
    Renew(ctx context.Context, instanceID string) error
    // Resign releases leadership immediately.
[...]

PR #437, opened 2026-08-07
feat(platform): fenced-lease leader election (#406)

nghiadaulau:
Implements #406 (epic #405): fenced-lease leader election, the foundation for running more than one control-plane instance with exactly one scheduler leader.

## What this delivers
- **Port** `ports.LeaderStore`: `Acquire` (atomic take / renew / takeover) + `Resign`, keyed by resource and holder, returning `held` plus a monotonically increasing `fence` token.
- **Migration 0061** `leader_leases`: a single upsert whose `CASE` arms take the lease only when it is free (expired) or already held (renewal, fence unchanged), bump the fence on a takeover, and leave a live foreign lease untouched. **Deliberately NOT tenant-scoped and NOT under RLS** — a leadership lease is global control-plane state, a documented exception to the tenant-table convention (like advisories in 0038).
- **Stores**: Postgres (single atomic `ON CONFLICT` upsert) + memory (mutex mirror, identical semantics).
- **Usecase** `leaderuc.Elector`: campaigns immediately, renews on a timer, caches `IsLeader()` for non-blocking callers, exposes the fence, resigns on shutdown so a follower takes over without waiting for the term, and audits every transition. Fail-safe: any `Acquire` error drops leadership rather than risk two leaders. `NewElector` enforces `0 < renew < term/2`.
- **Config** `SYNAPSE_LEADER_ENABLED` (default off) + `RESOURCE`/`TERM`/`RENEW`, wired into the API against the server lifecycle context (Postgres only; a single in-memory process is trivially the leader).

## Scope
This ships the **election mechanism**, tested end to end. Removing the single-instance lock (`AcquireSingletonLock`) and gating periodic dispatch on `IsLeader` — true horizontal scale — is the tracked follow-on: it changes the whole-server concurrency model and deserves its own careful change. The elector runs and audits leadership today but does not yet gate dispatch, so behaviour is unchanged for existing single-instance deployments.

## Verification
- Elector unit tests: single leader, renewal keeps the fence, expired takeover bumps the fence, resign frees the lease immediately (a follower acquires before the term expires), fail-safe on store error, and `NewElector` validation.
- Postgres integration: two contenders -> exactly one leader, renewal, takeover with fence bump, resign, migration 0061 down/up, and asserting `leader_leases` is NOT RLS-forced.
- `go build ./...`, `go vet ./...`, `go test ./...` (no DSN), gofmt clean.

Refs #405

### DIFF
--- a/cmd/synapse-api/main.go
+++ b/cmd/synapse-api/main.go
@@ -112,6 +112,7 @@ import (
 	"github.com/KKloudTarus/synapse-ce/internal/usecase/fleetagentuc"
 	"github.com/KKloudTarus/synapse-ce/internal/usecase/fleetwork"
 	"github.com/KKloudTarus/synapse-ce/internal/usecase/fptriage"
+	"github.com/KKloudTarus/synapse-ce/internal/usecase/leaderuc"
 	"github.com/KKloudTarus/synapse-ce/internal/usecase/llmverifier"
 	"github.com/KKloudTarus/synapse-ce/internal/usecase/orchestrator"
 	"github.com/KKloudTarus/synapse-ce/internal/usecase/ports"
@@ -173,6 +174,7 @@ func main() {
 	var assetStore ports.AssetRepository
 	var workOrderStore ports.WorkOrderStore
 	var fleetAgentStore ports.FleetAgentStore
+	var leaderStore ports.LeaderStore // postgres only; nil in memory mode (single process)
 	var findingRepo ports.FindingRepository
 	var judgmentStore analysisuc.Store // postgres or memory; satisfies both the narrow Store + ports.JudgmentStore
 	var commentRepo ports.CommentRepository
@@ -295,6 +297,7 @@ func main() {
 		assetStore = postgres.NewAssetRepository(pool)
 		workOrderStore = postgres.NewWorkOrderRepository(pool)
 		fleetAgentStore = postgres.NewFleetAgentRepository(pool)
+		leaderStore = postgres.NewLeaderStore(pool)
 		// SECURITY (#431 req 6, #432, #409): the fleet_* tables are RLS-protected, but RLS is a
 		// silent no-op if the runtime DB role is SUPERUSER or holds BYPASSRLS. When any fleet
 		// feature is enabled we refuse to serve unless the role can actually enforce isolation.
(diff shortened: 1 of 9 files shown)


nghiadaulau:
## Review round applied (commit 5bf1612)

Security-auditor confirmed the fenced lease's core is correct: the single-statement `INSERT ... ON CONFLICT DO UPDATE` with Postgres conflict re-evaluation never yields two DB-level leaders, on any of the four CASE arms or the `<=` expiry boundary; the memory store mirrors it; fail-safe, shutdown-resign and fail-closed wiring are all clean; no golden-rule issues. go-arch-reviewer: dependency rule PASS, `-race` clean. Findings addressed:

- **Security HIGH (fixed).** `Resign` deleted the lease row, so the next acquire re-INSERTed with `fence=1`, resetting the fence and breaking the monotonicity the mechanism promises (latent only because no gate consumes the fence yet). `Resign` now EXPIRES the lease (clears the holder, sets the term to now) and keeps the row, so a graceful handover bumps the fence via the normal takeover path and never resets it. Both stores updated; the Postgres test now asserts the fence does not decrease across resign+reacquire.
- **Security MEDIUM (documented).** The lease term is compared against each instance's wall clock. This is safe against two simultaneous leaders regardless of skew (the DB serializes the takeover), but a fast-clocked challenger can take over early. Recorded the NTP-bounded-skew operating assumption in the migration, and that moving the arithmetic onto the database clock is the hardening to do when a downstream first gates on leadership. (Kept the injected clock so the takeover/expiry tests stay deterministic.)
- **Security LOW (noted).** `IsLeader`/`Fence` are independent atomics; a combined accessor will be added when a gate first reads the pair (not consumed yet).
- **Arch (fixed).** Postgres store errors wrapped with `%w`; the elector test uses an in-file fake `ports.LeaderStore` instead of importing the memory package (no usecase->infra test edge); `main.go` logs a warning when leader election is enabled in memory mode.

Verification after fixes: `go test -race` for the elector, Postgres integration (single leader, renewal keeps fence, takeover bumps fence, resign, fence-monotonic across resign+reacquire, migration 0061 down/up, `leader_leases` not RLS-forced), `go build`/`go vet`/`go test ./...` (no DSN)/gofmt clean.

PR #438, opened 2026-08-07
feat(fleet): agent certificate identity and revocation (#408)

nghiadaulau:
Implements #408 (epic #405): cryptographic agent identity + revocation, on top of the #409 token-based identity.

## Scope note
The server runs plaintext behind a TLS-terminating proxy today (no in-process TLS), so full in-process mutual TLS is a P5 transport-topology decision. This PR delivers the **certificate identity mechanism** and binds it to the transport via the standard trusted-proxy pattern (the proxy verifies mTLS and passes the client certificate in a header). All of it is unit/fixture/Postgres tested.

## What this delivers
- **CA** `internal/infrastructure/fleetca` (behind `ports.CertificateIssuer`): verifies a CSR (signature + key-strength floor: RSA >= 2048 or ECDSA) and issues a short-lived client certificate whose subject encodes agent id (CN) and tenant (OU); returns the PEM + SHA-256 fingerprint. The agent's private key never leaves the host. `GenerateCA` bootstraps a dev/test CA.
- **Domain**: `Fingerprint`, a `stale` state, and revocation attribution (`RevokedAt/By/Reason`) with a `Revoke` method.
- **Migration 0062**: fingerprint + revoke columns, widened state CHECK, partial fingerprint index (RLS-scoped cert lookup).
- **Usecase**: `Enrol` accepts an optional CSR and (with a CA) issues a certificate + records the fingerprint; `AuthenticateCertificate` verifies `(tenant, agent id, fingerprint)` with a constant-time compare and refuses a revoked agent; `Revoke` takes operator + reason and cancels the agent's in-flight work orders (new `WorkOrderStore.CancelForAgent`).
- **Transport**: the agent plane authenticates by client certificate when a trusted proxy passes it in `SYNAPSE_FLEET_CLIENT_CERT_HEADER` — it reads tenant+agent from the subject and verifies the fingerprint, falling back to the bearer token otherwise. **SECURITY / trust boundary**: this header is trusted only because the operator asserts a trusted proxy verifies mTLS and strips any client-supplied value, and the app is not directly reachable. Documented in code + here.
- **Config** `SYNAPSE_FLEET_CA_CERT/KEY`, `SYNAPSE_FLEET_CERT_TTL`, `SYNAPSE_FLEET_CLIENT_CERT_HEADER`; the CA fails startup closed if configured but invalid.

## Deferred (referenced)
In-process mTLS listener / CSR-over-mTLS bootstrap and certificate rotation are P5 transport-topology work; the fingerprint identity is built so a cert can replace the token without lifecycle changes.

## Verification
- CA unit tests: issue + verify subject/fingerprint/client-auth-EKU/verify-against-CA; reject non-CSR, missing identity, weak RSA (1024), and a non-CA signer.
- Usecase tests: CSR->cert at enrol, cert auth success, wrong-fingerprint and revoked rejection, no-CA path, revoke-cancels-in-flight-orders.
- HTTP test: the client-cert header path (issued cert authenticates; malformed header unauthenticated, not 500; revoked cert 403) — this caught and fixed a raw-PEM URL-unescape corruption bug.
- Postgres integration: fingerprint + revoke persistence, single-use consume, migration 0062 down/up.
- `go build`/`go vet`/`go test ./...` (no DSN)/gofmt clean.

Refs #405

### DIFF
--- a/cmd/synapse-api/main.go
+++ b/cmd/synapse-api/main.go
@@ -29,6 +29,7 @@ import (
 	"github.com/KKloudTarus/synapse-ce/internal/infrastructure/cache/sbomcache"
 	"github.com/KKloudTarus/synapse-ce/internal/infrastructure/ebpf"
 	egressinfra "github.com/KKloudTarus/synapse-ce/internal/infrastructure/egress"
+	"github.com/KKloudTarus/synapse-ce/internal/infrastructure/fleetca"
 	"github.com/KKloudTarus/synapse-ce/internal/infrastructure/llm/openai"
 	"github.com/KKloudTarus/synapse-ce/internal/infrastructure/logstream"
 	"github.com/KKloudTarus/synapse-ce/internal/infrastructure/persistence/file"
@@ -1068,7 +1069,20 @@ func main() {
 			log.Error("fleet work service init failed", "err", werr)
 			os.Exit(1)
 		}
-		router.SetFleet(agentSvc, workSvc, clock.Now)
+		// Revoking an agent cancels its in-flight work orders (#408).
+		agentSvc.SetWorkOrders(workOrderStore)
+		// Optional certificate identity (#408): when a control-plane CA is configured, enrolment
+		// with a CSR issues a client certificate. Fail closed on a misconfigured CA.
+		if cfg.FleetCACertPEM != "" && cfg.FleetCAKeyPEM != "" {
+			ca, cerr := fleetca.New([]byte(cfg.FleetCACertPEM), []byte(cfg.FleetCAKeyPEM), cfg.FleetCertTTL)
+			if cerr != nil {
+				log.Error("fleet CA configured but invalid – check SYNAPSE_FLEET_CA_CERT/KEY", "err", cerr)
+				os.Exit(1)
+			}
+			agentSvc.SetCA(ca)
+			log.Info("fleet agent certificate identity ENABLED (CSR enrolment issues client certs)")
+		}
+		router.SetFleet(agentSvc, workSvc, clock.Now, cfg.FleetClientCertHeader)
 		router.SetFleetAdmin(agentSvc)
 		log.Info("fleet agent transport ENABLED (agent-auth plane; operator agent-admin routes)")
 	}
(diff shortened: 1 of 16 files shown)


nghiadaulau:
## Review round applied (commit cf23742)

Security-auditor confirmed the cert-identity scheme is sound and the forgery reasoning holds given the documented proxy trust boundary (the CA sets the subject itself so a CSR cannot forge identity; the fingerprint compare + CN-bound lookup make cross-agent auth impossible; the possession proof is the proxy's mTLS, and the header trust is gated on config + documented). go-arch-reviewer: dependency rule PASS. Findings addressed:

- **Security MEDIUM (fixed).** The app now enforces the certificate validity window: `authByClientCert` rejects a certificate whose `NotBefore`/`NotAfter` does not contain now, so a stored fingerprint no longer outlives its certificate. Added an expired-certificate rejection test.
- **Security MEDIUM (fixed).** Enrol is side-effect-safe on a bad CSR: the certificate is issued (CSR verified) BEFORE the single-use token is consumed and the agent created, and a CSR failure is wrapped as `shared.ErrValidation` (HTTP 400, not 500). No burned token, no orphan agent. The fingerprint is persisted in the single `CreateAgent` insert. Added a "bad CSR does not burn the enrol token" test.
- **Security/arch LOW (fixed).** `checkKeyStrength` requires an ECDSA curve >= 256 bits (matching the RSA-2048 floor); `fleetca.New` also requires the signing cert to carry `KeyUsageCertSign`.
- **Arch (fixed).** The fingerprint (SHA-256 of DER, hex) now has ONE definition — `fleetagent.CertFingerprint` in the domain — referenced by both the CA and the auth path, so they cannot drift.
- **Deferred (noted).** A strict certificate-required mode (refuse the bearer fallback when the cert header is configured) is a follow-up; the current fallback is a documented migration posture. `CertificateIssuer.Issue` omitting `context` is acceptable (pure in-process crypto).

Auditor's stated caveat is intentional and documented: the fingerprint proves "Synapse issued this certificate," not private-key possession; possession is proven by the proxy's mTLS handshake, and the header is trusted only when configured behind a proxy that strips client-supplied values.

Verification after fixes: all CA/usecase/HTTP/Postgres tests pass (including the two new ones); `go build`/`go vet`/`go test ./...` (no DSN)/gofmt clean.

Issue #388, opened 2026-08-05
feat(ai-triage): evaluation harness and shadow mode (P1.1)


nghiadaulau:
Parent: #387 (AI-assisted false-positive triage).

## Goal

Prove AI triage quality on representative, labelled data before any automatic gate exemption is allowed. This is the gate that P1.2 to P1.5 and the canary rollout depend on.

## Scope

- Build a golden dataset of human-reviewed findings with version, provenance, and reviewer recorded.
- Cover true positives, false positives, uncertain cases, and adversarial prompt-injection cases.
- Segment results by language, finding kind, CWE, severity, and framework.
- Record precision, recall, false-negative escape rate, disagreement rate, and coverage.
- Add a shadow mode: run proposer and verifier, persist the decision, but always keep the finding gating.
- Compare prompt and model versions per evaluation run.
- Emit a machine-readable report usable in CI and for model promotion.

## Non-goals

- No automatic gate exemption from this issue; shadow mode must never produce `gate_exempt`.
- No dependency on production data for the CI evaluation.

## Acceptance criteria

- [ ] Dataset has version, provenance, and reviewer.
- [ ] Evaluation runs reproducibly in CI without production data.
- [ ] Shadow mode cannot produce `gate_exempt`.
- [ ] A report exists per model/prompt/policy version.
- [ ] PM and Security approve the quality threshold before canary.

## References

- Depends on a deterministic verifier: #384 (temperature 0 dropped) must be fixed, otherwise disagreement/precision metrics are non-reproducible.
- Relates to the Judgment primitive (ADR-0021) and `verdict.EvidenceThreshold` (75).

H1eu232:
## Implementation complete ? PM/Security threshold sign-off remains

The engineering scope of P1.1 is on `main` (original work #451, CI mirror/final merge #477), and the downstream controls are now also merged:

- shadow evaluation and versioned golden dataset/report
- deterministic baseline comparison (#554)
- drift evidence (#555)
- versioned PM/Security promotion and rollback ledger (#559)

The first four acceptance criteria in this issue are therefore implemented. The remaining checkbox is intentionally a human governance decision, not another code change:

- [ ] PM and Security approve the quality threshold before canary

The release boundary currently enforces a floor no weaker than `DefaultAIEvaluationPromotionPolicy`:

- minimum precision: **95%**
- maximum false-negative escape rate: **0%**
- maximum precision, recall, coverage, and verifier-coverage drop: **0 bp**
[...]

PR #451, opened 2026-08-08
feat(ai-triage): add P1.1 shadow evaluation harness

H1eu232:
Refs #388
Parent epic: #387

## Summary

- add a fail-closed `shadow|enforce` rollout mode; `shadow` is the default and records `would_gate_exempt` while forcing `gate_exempt=false`
- seal shadow decisions with scan evidence and revalidate the observation set against the full P0 consensus/floor policy
- add a strict, versioned golden-dataset schema with provenance, reviewer, TP/FP/uncertain labels, and adversarial prompt-injection fixtures
- add `synapse-fptriage-eval`, which runs the production triager and production policy in forced shadow mode and emits deterministic JSON
- report precision, recall, false-negative escape rate, disagreement, coverage, plus language/kind/CWE/severity/framework breakdowns
- bind reports to dataset/model/prompt/policy identity with `dataset_sha256` and deterministic `run_id`
- document configuration, evaluation workflow, report contract, and Make target

## Safety properties

- missing, empty, or unknown rollout mode fails closed to `shadow`
- shadow policy clears forged `GateExempt`, never enters `AIGateExemptKeys`, and rejects an evaluation report if any gate exemption appears
- `WouldGateExempt` is observability-only and is revalidated from the full DTO plus finding eligibility and human-review floors
- evaluation requires distinct canonical model IDs and exact model/prompt/policy metadata
- the CI fixture is synthetic and contains no production findings or source

## #388 acceptance mapping

- [x] dataset has version, provenance, reviewer, strict schema, and unique cases
- [x] deterministic CI evaluation uses a repository fixture without production data
- [x] shadow mode cannot produce `gate_exempt`
- [x] machine-readable report identifies dataset/model/prompt/policy versions
- [ ] PM/Security approve the quality threshold before canary (external approval; enforcement remains shadow by default)

## Verification

- `go test ./internal/platform/config ./cmd/synapse-fptriage-eval`
- `go vet ./internal/usecase/fptriage ./internal/usecase/sca ./internal/platform/config ./cmd/synapse-fptriage-eval`
- `git diff --check`

The local Windows Defender installation blocks reading the pre-existing `internal/infrastructure/tools/sast/php_lex_test.go` signature, so local aggregate `cmd/synapse-api` / `cmd/synapse-cli` package tests could not run; the file is restored and unchanged. CI should run those packages on a clean Linux runner.


nghiadaulau (changes requested):
## Changes requested — the branch does not compile

`go build ./...` fails on this branch as submitted, in two of the packages the description says were skipped locally:

```
cmd/synapse-cli/main.go:1326:27: sca.AITriageMode undefined (type *sca.Service has no field or method AITriageMode)
cmd/synapse-api/main.go:731:30:  undefined: sca
```

Both commands import the use case as **`scauc`**, and in the CLI `sca` is the local `*scauc.Service` (`cmd/synapse-cli/main.go:1167`). So:

```go
// cmd/synapse-cli/main.go:1326
-sca.SetFPTriageMode(sca.AITriageMode(cfg.FPTriageMode))
+sca.SetFPTriageMode(scauc.AITriageMode(cfg.FPTriageMode))

// cmd/synapse-api/main.go:731
-scaService.SetFPTriageMode(sca.AITriageMode(cfg.FPTriageMode))
+scaService.SetFPTriageMode(scauc.AITriageMode(cfg.FPTriageMode))
```

Two identifiers. But I want to be direct about why this matters more than a typo: **those two lines are the entire rollout-mode wiring**, so `SYNAPSE_FP_TRIAGE_MODE` has never reached a running binary. The `enforce` path is unreachable and untested end-to-end, and the shadow default currently holds only because `AITriageMode`'s zero value happens to be shadow — by luck of the composition never building, not by the mechanism this PR adds. The unchecked `make build vet test typecheck` box in the checklist is exactly where this hid; a targeted `go vet` list that omits `./cmd/...` cannot catch a composition-root error.

I applied those two edits locally and re-ran everything: `go build ./...` clean, `go vet ./...` clean, full suite **166 packages, 0 failures**. So this is the only thing standing between the PR and green.

**Also needs a rebase** — it now conflicts with `main` in `internal/usecase/sca/service.go` and `internal/platform/config/config.go`. Both conflicts are gofmt column alignment plus one new field on each side; trivial to resolve.

---

## What I checked, and what holds

I went after the one property that matters: **can a shadow decision reach a CI gate?** It cannot, and the separation is defended at more than one layer rather than by a single flag.

**Fail-closed in three independent places.** `normalizeFPTriageMode` returns `shadow` for anything that is not exactly `enforce`; `SetFPTriageMode` does the same again at the service; and `applyAIGatePolicy` tests `mode != AITriageModeEnforce`, so the zero value is shadow too. A config typo, a missing env var and an unset field all land in the same safe state.

**The two key sets are genuinely disjoint.** `AIGateExemptKeys` now requires `!c.Shadow && c.GateExempt`; `AIWouldGateExemptKeys` requires `c.Shadow && c.WouldGateExempt && !c.GateExempt`, and both re-derive the full policy — consensus, `humanReviewFloor`, `isFPTriageEligible` — from the finding rather than trusting the DTO. A forged `GateExempt` on an inbound critique is cleared at the top of every loop iteration in `applyAIGatePolicy`.

**Nothing consumes the shadow set as authority.** I traced every caller: `AIWouldGateExemptKeys` appears only in the stderr summary line (`cmd/synapse-cli/main.go:1386`) and in the scan log counters. The gate at `cmd/synapse-cli/main.go:1465` reads `fpGateExempt` only.

**The evaluation harness cannot grant authority even by accident.** It forces `AITriageModeShadow` (`fptriage_evaluation.go:226`) and then hard-rejects the whole report if any case comes back with `GateExempt` (`:263`) — a tripwire rather than an assumption. The bundled dataset is synthetic: 8 cases, fabricated snippets, 4 TP / 2 FP / 2 uncertain, no production finding or source.

**Policy version bumped `fp-gate-v2` → `fp-gate-v3`**, which is the right call — the authorization contract changed, and an audit replay needs to distinguish them.

---

## Two smaller things, not blocking

1. **`AITriagePolicyVersion` and `PromptVersion` were exported** so the eval binary can read them. That widens the API of the policy package for a tooling need. A narrow accessor (`sca.EvaluationIdentity()`) would keep the constants unexported; your call, but worth a sentence in the PR if you keep it as is.
2. **No composition-root test.** Nothing asserts that `SYNAPSE_FP_TRIAGE_MODE=enforce` actually reaches `applyAIGatePolicy` — the policy-level tests cover both modes well, but the wire from env to policy is the part that just broke. A small test that builds the service from a config and asserts the effective mode would close it permanently.

Push the two-identifier fix and a rebase, and I will re-run the full verification and merge.

H1eu232:
Addressed the requested changes in `673359c` after rebasing onto current `main` (including merged #452 and all changes through `2e7b25e`). The PR is mergeable again.

### Blocking compile issue

Fixed both composition roots. I removed the fragile package-type casts entirely: `Service.SetFPTriageMode` now accepts the normalized config string and performs its own fail-closed normalization, so both binaries wire it as:

```go
scaService.SetFPTriageMode(cfg.FPTriageMode)
sca.SetFPTriageMode(cfg.FPTriageMode)
```

This preserves the three-layer fail-closed behavior while avoiding the `sca` local/package alias class of error.

### Non-blocking follow-ups also addressed

1. Exported policy constants were made private again. The eval command now uses narrow immutable accessors: `fptriage.EvaluationPromptVersion()` and `sca.EvaluationPolicyVersion()`.
2. Added `TestFPTriageModeConfigReachesService`, which exercises `SYNAPSE_FP_TRIAGE_MODE -> config.Load() -> Service.SetFPTriageMode -> effective internal mode` for default, enforce, and invalid values. This specifically covers the composition wire that was broken.
3. Preserved #452's CLI canonical-alias warning while resolving the rebase.

### Verification on the rebased tree

- `go build ./...` ? PASS (clean snapshot; both composition roots compile)
- `go vet ./...` ? PASS
- `go test ./cmd/synapse-api ./cmd/synapse-cli` ? PASS as part of the full run
- `go test -race ./internal/usecase/fptriage ./internal/usecase/sca` ? PASS
- targeted `fptriage` / `sca` / `config` / eval-command tests ? PASS
- `pnpm run typecheck` ? PASS
- `git diff --check` ? PASS

The Windows full-suite run reached all packages but retained three platform/environment failures unrelated to this PR: two pre-existing POSIX `0600` assertions on Windows (`synapse-agent`, `fleetclient`) and one PHP analyzer case because Windows Defender blocks the pre-existing `php_lex_test.go`. The relevant packages and both command composition roots are green; Linux CI/re-review should exercise the complete suite normally.

Please re-review when convenient.

nghiadaulau (approved):
## Approve — the two properties this rests on are real, and I broke them to prove it

Reviewed strictly, checked out, and **mutation-tested the safety claims** rather than reading them. An AI component that can mark a finding exempt from a quality gate is the highest-consequence thing in this repo, so "the tests pass" is not evidence on its own.

### Property 1: shadow mode cannot produce `gate_exempt`

I removed the `continue` from the shadow branch so it falls through to `GateExempt = true`:

```
--- FAIL: TestApplyAIGatePolicyShadowNeverExempts
--- FAIL: TestScanEvidenceContentSealsShadowDecision
--- FAIL: TestEvaluateFPTriageGoldenDataset
```

Three independent tests, at three different levels — the policy, the sealed evidence, and the evaluation report. That is the right shape for a property this load-bearing: no single test deletion can quietly remove it.

### Property 2: an unknown mode fails closed

I made `normalizeFPTriageMode` pass its input through unchanged:

```
--- FAIL: TestFPTriageModeDefaultsToShadow
```

The normalizer is the right design: **only the exact spelling `enforce` enables enforcement**, everything else — empty, typo, `Enforce `, `true`, `1` — becomes shadow. A misconfiguration cannot accidentally grant gate authority, and the `aiTriageMode` zero value is shadow too, so a caller that forgets to pass a mode gets the safe one rather than the fast one.

### What is genuinely well judged

**`hasVerifiedConsensus` re-derives instead of trusting `Verified`.** The comment names the threat exactly — a buggy or alternate `FPTriager` setting one boolean must not grant itself gate authority. It rechecks both verdicts, both confidences against `verdict.EvidenceThreshold`, and proposer ≠ verifier through `agent.CanonicalModelID`, which is the same canonicalization #471 is adding in `llmverifier`. Consistent, and consistent for the right reason.

**`GateExempt = false` is set unconditionally at the top of every iteration**, before any branch. That is what makes "clears forged `GateExempt`" true rather than aspirational: a DTO arriving with the field pre-set is overwritten, not inspected.

**The shadow branch sits after every other check, not before them.** So `would_gate_exempt` means "this would have been exempted **and** it cleared consensus, the finding lookup, the human-review floor, eligibility, and evidence" — not the looser "the model thought so". Placed earlier, the whole evaluation would measure the wrong thing and read as more capable than it is.

**A report can only be produced against the current policy** (`run.PolicyVersion != EvaluationPolicyVersion()` → error), and `dataset_sha256` plus a deterministic `run_id` bind it to the exact dataset. So a stale report cannot be presented as current, which is the failure mode that makes evaluation harnesses dangerous rather than useless.

**Bumping `fp-gate-v2` → `fp-gate-v3`** — the policy identity moves when the behaviour moves. Easy to forget; it was not forgotten.

**`GateExempt` is carried in the report as a tripwire** that must always be false, rather than omitted because it "cannot happen". Asserting the invariant where it would be observed is better than trusting the code path.

### One thing I checked and was wrong about, worth recording

I suspected an empty-model bypass: `agent.SameModel` returns `false` when either id canonicalizes to empty, so `SameModel("", "")` is false and would *pass* a distinctness check. `EvaluateFPTriage` refuses empty proposer/verifier/prompt/policy **before** reaching that call, so the hole is closed. Worth stating because the same shape is a real trap elsewhere — `SameModel` alone is not a distinctness check, only `non-empty && !SameModel` is.

### Minor

- `cmd/synapse-fptriage-eval` has no tests, which is fine for an 84-line composition root — the logic under it is covered. Worth keeping it that thin.
- The last acceptance box (PM/Security approve the threshold before canary) is correctly left unticked; enforcement staying shadow by default is what makes that safe to defer rather than a gap.

### Before merging

CI has never run on this PR — only GitGuardian, because Actions was still disabled when it was opened, so no workflow run object was ever created for it. I am refreshing the branch against current `main` to get the pipeline to run, and will merge on green, not before.

nghiadaulau:
### Why this PR cannot be merged directly, and where it went (#477)

This PR never got a GitHub Actions check suite. It was opened while Actions was **disabled** on this repository, and GitHub does not create one retroactively — a synchronizing push does not fix it either. Its head commit lists GitGuardian alone, so its pipeline can never run, and I am not merging anything that has not been through CI.

I could not re-open a fresh PR from your fork branch (`Fork collab can't be granted by someone without permission`), so the branch is mirrored into this repository and the pipeline runs there. **Identical commits, your authorship preserved**, and it will be merged with a merge commit rather than a squash so each commit keeps you as author.

My full review — including the mutation-test evidence — stays on this PR; nothing about the code changed.

**One correction worth recording, because I got there the slow way.** I first blamed the fork-PR contributor-approval policy, then a missing check suite. Both were wrong. The real cause was that repository Actions had been switched to `enabled: false` partway through: no run object is created for *any* event, so `?status=action_required`, `waiting` and `queued` all report zero and it reads as a permissions puzzle rather than an off switch. The decisive probe was merging an unrelated PR to `main` and seeing that even **that** produced no run. Actions is back on now.

Mirror: #477

nghiadaulau:
### CI found one real issue, and it is not a style point

With the pipeline finally running on the mirror (#477), `Lint (Go)` failed:

```
cmd/synapse-fptriage-eval/main.go:75:18: Error return value of `out.Close` is not checked (errcheck)
    defer out.Close()
```

Worth being specific about why this one matters rather than silencing it. **`Close` is where a buffered write actually fails.** A dropped `Close` error means a **truncated evaluation report written with exit status 0** — and this report is precisely the artefact someone reads to decide whether AI triage is good enough to move from shadow to enforce. A partial report presented as a success is the one output this command must never produce, so errcheck is pointing at the exact failure mode the harness exists to prevent.

Fixed in d7e2d72 on the mirror: `Close` is reported on the success path, still called on the encode-error path, and `os.Stdout` is left alone.

Two notes, neither a criticism of the change:

- The PR body lists `golangci-lint` under verification. It is worth checking which version and scope ran locally — CI uses **v2.12.2** across the whole tree, and it caught this in `cmd/`, which is easy to miss when running the linter over the packages you touched.
- This is the only issue in the whole run. Everything else — Backend (Go), Windows, Frontend, deb ×2, rpm, msi, the libc-floor refusal — passed. For a 971-line change touching the gate policy, that is a good result.

The review and the mutation-test evidence for both safety properties are unchanged and still stand.

PR #452, opened 2026-08-08
fix(ai-triage): finish P0 review follow-ups

H1eu232:
Follow-up to #398
Parent epic: #387

## Summary

Completes the three non-blocking follow-ups from the final safety review on #398:

1. **CLI/API operator-warning parity**
   - after `WithVerifier`, the CLI now checks `coord.VerifierModel()` exactly like the API
   - canonical aliases remain advisory-only and emit a clear warning

2. **Zero-padded CWE canonicalization**
   - numeric identifiers such as `CWE-0079` normalize to `CWE-79`
   - exact-token matching is preserved: `CWE-0890` becomes `CWE-890` and does not collide with protected `CWE-89`
   - zero-padded credential IDs such as `CWE-0798` remain excluded from LLM candidate selection
   - malformed/non-CWE tokens are left unchanged rather than guessed

3. **Explicit FPTriager trust boundary**
   - documents that injected implementations are trusted in-process components
   - deterministic policy revalidation contains buggy/forged DTO fields, but is not a sandbox for malicious code that already has process authority

## Verification

- `go vet ./internal/usecase/fptriage ./internal/usecase/sca`
- `git diff --check`

Regression tests cover protected zero-padded CWE matching, exact non-collision, and credential-CWE candidate exclusion.

The local Windows Defender installation blocks reading the pre-existing `internal/infrastructure/tools/sast/php_lex_test.go` signature, so the aggregate CLI package could not be compiled locally; that file is restored and unchanged, and CI can compile it on Linux.

### DIFF
--- a/cmd/synapse-cli/main.go
+++ b/cmd/synapse-cli/main.go
@@ -1330,6 +1330,9 @@ func run(path string, failOn shared.Severity, mode, priority string, ignoreUnfix
 			if cfg.VerifierModel != "" && cfg.VerifierModel != cfg.FPTriageModel {
 				if vllm, verr := openai.New(cfg.LLMBaseURL, cfg.LLMAPIKey, cfg.VerifierModel, cfg.LLMTimeout); verr == nil {
 					coord.WithVerifier(vllm, cfg.VerifierModel)
+					if coord.VerifierModel() == "" {
+						fmt.Fprintf(os.Stderr, "synapse-cli: verifier model %q aliases the proposer after canonicalization; AI triage remains advisory-only\n", cfg.VerifierModel)
+					}
 				} else {
 					fmt.Fprintf(os.Stderr, "synapse-cli: verifier model %q unavailable; AI triage remains advisory-only: %v\n", cfg.VerifierModel, verr)
 				}
--- a/internal/usecase/ports/fptriage.go
+++ b/internal/usecase/ports/fptriage.go
@@ -44,7 +44,11 @@ type AICritique struct {
 // FPTriager runs an LLM false-positive critique over candidate findings from a workspace and returns a
 // per-candidate advisory verdict. It is best-effort and PROPOSE-ONLY: it never mutates or deletes a
 // finding. The caller applies the deterministic gate policy; only a verified critique that clears the
-// human-review floor can be retain-and-mark gate-exempt. Injected optionally into the scan pipeline;
+// human-review floor can be retain-and-mark gate-exempt.
+//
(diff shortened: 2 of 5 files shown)


nghiadaulau (approved):
## Review: APPROVE (with a hardening commit applied)

Reviewed the three P0 follow-ups. All sound. Verified locally (Actions is disabled here, so GitGuardian is the only automated check — everything below was run by hand):

- ✅ **Full build incl. `synapse-cli`** on macOS — the Windows-Defender caveat in the description is environment-specific; `php_lex_test.go` compiles fine and the aggregate CLI package builds.
- ✅ `go vet`, `go test -race ./internal/usecase/sca/... ./internal/usecase/fptriage/...`, `golangci-lint` — all green, 0 issues.

### 1. CWE zero-padded canonicalization — correct
The protected sets (`humanReviewCWEs`, `hasCredentialCWE`) are all canonical (no leading zeros) and matched **exactly** (map/switch, not substring), so canonicalizing incoming tokens *closes* the `CWE-0079`→`CWE-79` bypass without de-protecting anything or colliding (`CWE-0890`→`CWE-890`, never `CWE-89`). Fail-closed on empty-token SAST/Misconfig preserved.

### 2. LLM propose-only boundary — sound
`hasVerifiedConsensus` still revalidates the whole DTO (distinct proposer≠verifier, both refuted, both ≥ evidence threshold); `applyAIGatePolicy` resets `GateExempt` each pass. A single model or a forged field cannot grant exemption.

### 3. CLI/API operator-warning parity — sound
`WithVerifier` genuinely no-ops on alias (`agent.SameModel`), so the CLI's `VerifierModel()==""` warning accurately tells the operator exemption is impossible — mirroring the API path (`cmd/synapse-api/main.go:739-743`).

### 4. Trust-boundary docs — accurate, documentation-only.

### Hardening applied on top (`b6e3fb1`)
An independent security pass found one **LOW** defense-in-depth gap: the tokenizer's separator set was narrower than the canonicalizer's tolerance, so a dangerous CWE spelled with an embedded separator/punctuation (`CWE-79: XSS`, NBSP/`\f`, `CWE-79A`, bare `CWE-`) produced a token that neither canonicalized, matched the protected set, nor tripped the empty-token fail-closed path — letting a dangerous class slip the human-review floor on a malformed spelling. Not presently exploitable (eligible findings come from first-party curated rules emitting canonical IDs), but a real fail-closed asymmetry. Fixed by widening the tokenizer to all unicode whitespace + `: . ( )` and treating any unparseable `CWE-`-prefixed token as fail-closed to human review, with a regression test.

Merging.

PR #455, opened 2026-08-08
fix(llm): enforce output token limits

XUanhoa04:
## Summary

`ChatRequest.MaxTokens` was populated by the agent, false-positive triage, and judgment verifier, but the OpenAI-compatible adapter never serialized it. Provider defaults could therefore exceed the output budget the callers requested.

This change makes the limit effective while preserving compatibility across model generations. It follows the current Chat Completions field, `max_completion_tokens`; the deprecated `max_tokens` spelling is used only after a legacy gateway explicitly rejects the current field.

Refs #392.

## Changes

- send positive `ChatRequest.MaxTokens` as `max_completion_tokens`
- leave non-positive limits unset so provider-default behavior is unchanged
- on an explicit field-unsupported 400, retry once with legacy `max_tokens`
- cache the negotiated dialect per model ID, since one gateway may route both current and legacy model families
- keep unrelated 400/auth errors terminal and preserve existing bounded transient retries
- return `ErrOutputTruncated` on `finish_reason=length`, so a partial structured verdict cannot reach triage or a judgment gate
- add wire-level regression coverage and an Unreleased changelog entry

Current field reference: https://platform.openai.com/docs/api-reference/chat

## Safety

- no prompt, source context, request body, or API key is added to logs
- fallback occurs only when a 400 response both names `max_completion_tokens` and identifies it as unsupported
- a token-truncated reply fails closed instead of being parsed or acted upon
- proposer/verifier semantics and the deterministic report path are unchanged

## Verification

- `go test ./internal/infrastructure/llm/openai ./internal/usecase/fptriage ./internal/usecase/llmverifier ./internal/usecase/orchestrator ./internal/usecase/agenttools`
- `go vet ./internal/infrastructure/llm/openai ./internal/usecase/ports ./internal/usecase/fptriage ./internal/usecase/llmverifier ./internal/usecase/orchestrator ./internal/usecase/agenttools`
- `git diff upstream/main...HEAD --check`

The race test could not run locally because this Windows Go environment has CGO disabled. Full repository CI is left to the Linux workflow; local Windows Defender quarantines the pre-existing `internal/infrastructure/tools/sast/php_lex_test.go` fixture, which is not part of this commit.

## Checklist

- [ ] `make build vet test typecheck` passes (targeted Go test/vet commands above pass; full Linux CI pending)
- [x] `cd web && pnpm build` is not applicable (no dashboard changes)
- [x] Documentation updated via `CHANGELOG.md`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -47,6 +47,12 @@ capabilities below are already shipped on `main`.
 
 ### Fixed
 
+- **Bounded LLM output.** `ChatRequest.MaxTokens` was set by the agent, false-positive triage, and
+  judgment-verifier call sites but never reached the OpenAI-compatible wire request, so provider defaults
+  could exceed the intended output budget. The adapter now sends the current `max_completion_tokens`
+  field, negotiates and caches the deprecated `max_tokens` spelling only when a legacy gateway explicitly
+  rejects the current field, and fails closed on `finish_reason=length` rather than parsing a potentially
+  truncated structured verdict.
 - **Deterministic AI adjudication.** An explicit `temperature: 0` never reached the provider: the
   OpenAI-compatible adapter only sent the field when it was greater than zero, so the value a
   deterministic caller asks for was indistinguishable from "unset" and was dropped from the request. The
--- a/internal/infrastructure/llm/openai/client.go
+++ b/internal/infrastructure/llm/openai/client.go
@@ -15,6 +15,7 @@ import (
 	"io"
 	"net/http"
 	"strings"
+	"sync"
 	"time"
 
 	"github.com/KKloudTarus/synapse-ce/internal/domain/agent"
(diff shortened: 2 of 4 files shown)


nghiadaulau (approved):
## Approved

I checked this against the thing that actually matters here — **can a truncated structured verdict reach a judgment gate** — and it cannot. Verified at all four call sites, not just the two the description names:

| Caller | On error |
|---|---|
| `fptriage/coordinator.go:189` (critique) | `res.Err = ...; return` — never reaches `parseCritique` |
| `fptriage/coordinator.go:217` (verifier) | same |
| `llmverifier/assess.go:84` | returns before the verdict is sealed |
| `orchestrator.go:305` | `o.fail(...)` |

No caller ignores the error, and none loops on a non-transient one, so `ErrOutputTruncated` really is terminal everywhere. `finish_reason=length` has no benign meaning, so there is no case where this drops a complete response.

The negotiation is correctly conservative: the fallback fires only on a **400** that both names `max_completion_tokens` *and* carries an unsupported-marker, so an unrelated bad request, an auth failure and a truncated completion all stay terminal — `TestChatDoesNotFallbackOnUnrelatedBadRequest` pins that. Caching per **model** rather than per client is the right call and the test proves a current model routed through the same gateway does not inherit the legacy spelling. Wrapping 429/5xx in `providerStatusError` preserves `errors.Is(err, ErrUnavailable)`, so the existing bounded transient retry is unchanged.

### Verification I ran (the part the description left to CI)

The branch was 10 commits behind `main`, so I merged `main` in locally first — it merges clean:

- `go build ./...` · `go vet ./...` — clean
- full suite: **177 packages, 0 failures**
- `go test -race -count=3 ./internal/infrastructure/llm/openai` — clean (the `bodies` slice appended from the httptest handler is the kind of thing `-race` usually catches; it does not fire here)
- `go test -race` on `fptriage`, `llmverifier`, `orchestrator`, `agenttools` — clean
- `golangci-lint` on the touched packages — 0 issues

Golden rules: no prompt, request body or key reaches a log; the provider snippet in `providerStatusError.Error()` is the same content the previous `fmt.Errorf` already carried, so no new exposure. Proposer/verifier semantics and the report path are untouched.

### Three things to know, none blocking

1. **The agent path changes behaviour**, not just triage. `orchestrator.go:305` now fails a session when the model hits the cap, where it previously recorded a truncated reply as `StatusSucceeded`. That is the right direction — a cut-off "final answer" sealed as success is worse — but it is a live behaviour change for anyone running with a non-zero `MaxTokens`.
2. **The truncated call's tokens are not accounted.** `sess.TokensUsed += resp.Usage.TotalTokens` sits *after* the error return, so the spend on the truncated call never lands in the budget. The session fails anyway, so this is cosmetic today; it would matter if truncation ever became recoverable.
3. **The dialect cache is never invalidated.** A gateway upgraded from legacy to current keeps receiving `max_tokens` for that model until the process restarts. Harmless (current gateways still accept it), worth remembering.

Merging.

PR #457, opened 2026-08-08
feat(reachability): add Yarn lockfile resolution support

VietSory:
## Summary

Adds deterministic Yarn Berry lockfile resolution to the existing JS/TS dependency reachability pipeline.

This PR has been rebased and narrowed following maintainer review. The overlapping npm, pnpm, SBOM correlation, and resolver work from the original R2C implementation has been removed; this PR now contains only the Yarn-specific functionality that is not already present on `main`.

* resolve Yarn Berry dependency descriptors to exact lockfile locators
* verify locator package identity and version before accepting resolution evidence
* support Berry lockfile metadata versions used by current Yarn releases
* support scoped package names and exact descriptor-to-version correlation
* handle identity-preserving Yarn `npm:<version>` selectors without changing npm/pnpm alias semantics
* keep same-name workspace dependencies ambiguous rather than incorrectly treating them as third-party packages
* preserve valid positive resolutions while marking coverage incomplete when unrelated malformed Yarn locator evidence is observed
* fail closed on duplicate descriptors, malformed locators, unsupported metadata, and other ambiguous lockfile evidence
* ignore stale npm lockfile evidence when the root project explicitly selects Yarn
* retain explicit coverage for unsupported Yarn Classic/v1 behavior

All regression tests are behavior-focused and the implementation remains deterministic, bounded, and offline. It does not execute Yarn, Node, project scripts, registries, or other external tooling.

## Validation

The rebased implementation was validated against current upstream `main`:

* `go build ./...`
* `go vet ./...`
* full `go test ./...` on Linux
* `CGO_ENABLED=0 go build ./...`
* `golangci-lint`
* Windows build and vet
* full `internal/infrastructure/tools/jsresolve` tests on Windows

The remaining Windows full-suite failures are existing POSIX file-permission assertions outside `jsresolve`. The frontend suite also retains the existing jsdom/undici worker-startup failure and is unrelated to this Go-only change.

The PR is intentionally split from the remaining lockfile-hardening work, which is tracked separately in #470.


nghiadaulau:
Heads-up on an overlap, and an apology for it: R2C of #400 has just landed on `main` from a parallel effort — #459 (exact-PURL correlation) and #460 (lockfile importer selection). I did not check the open-PR list before starting, which I should have; that's on me.

So this branch will now conflict substantially. Rather than close it, I want to flag what **this PR has that `main` does not**, because it is real and I deliberately deferred it:

- **Yarn.** `main` currently *reports* yarn as a limitation (`CoverageUnsupportedPackageManager`: "yarn lockfiles do not record per-importer resolutions"), because recovering a workspace's own resolution needs yarn's descriptor matching re-implemented. This PR actually does it, and goes further with **Berry locator identity verification** before emitting an exact PURL — which is a correctness property `main` does not have at all.
- **Shrinkwrap**, pnpm peer-importer edges, and the JSON-depth/duplicate-metadata hostile cases, which have no equivalent on `main`.

What landed on `main`, for comparison, so you can see the delta precisely:
- `resolver_sbom.go` — npm-only component index, exact-PURL correlation, case-exact matching (`JSONStream` ≠ `jsonstream`), `maxComponents` bound, sorted-name determinism.
- `resolver_depspec.go` — the declared dependency spec decides identity: an `npm:` alias correlates to the **alias target**, and `file:`/`link:`/`portal:`/`workspace:`/`patch:`/`git+`/`github:`/`https:`/`owner/repo` **refuse** correlation. (A security review found that correlating an aliased `"lodash": "npm:lodash-es@^4"` to a same-named component attaches the wrong identity while looking deterministic — worth checking this branch handles it too.)
- `resolver_lockfile.go` — pnpm `importers:` and npm nearest-wins hoisting.

Suggestion, if you're willing: rebase onto `main` and scope this down to the **yarn + shrinkwrap delta** plus any hostile regressions `main` lacks. That keeps the part of this work that is genuinely additive and avoids re-litigating the parts that are already in. I'm happy to review that rebase promptly.

If you'd rather I do the rebase myself, say so and I will.

nghiadaulau (changes requested):
## Changes requested — needs a rebase, but most of this is still wanted

First, the part that is on us and not on you: **R2C landed on `main` from another branch while this PR was open** (#459, #460), and #400 has since been closed. That is why this now shows `CONFLICTING`. I am not going to ask you to re-justify the work — I went and diffed your branch against what actually shipped, so the ask below is specific rather than "please rebase and see what happens".

### What `main` has today

- pnpm `importers:` parsing, so a workspace's own resolution selects the version
- npm lockfile hoisting with nearest-wins install-path selection
- alias handling (`"lodash": "npm:lodash-es@^4"` resolves to the target identity, not the alias name)
- exact-then-folded SBOM component matching, with a folded match reported rather than sealed
- resolution-as-filter: anything unresolved, ambiguous or unsupported forces `Complete = false`

### What `main` does **not** have, and this PR does

I verified each of these against `main` rather than taking the description on trust:

1. **Yarn is not supported at all.** `resolver_lockfile.go:101-108` records a coverage issue — *"yarn lockfiles do not record per-importer resolutions, so version selection stays ambiguous"* — and stops there. Because any coverage issue makes resolution incomplete, and both Tier-1 and Tier-2 reachability refuse on incomplete resolution, **every Yarn repository currently gets no JS reachability at all.** Your Berry descriptor→locator work with identity verification and `npm:` protocol selectors is the single most valuable thing in this branch, and nothing on `main` overlaps it.
2. **pnpm `snapshots:` consistency and peer-qualified keys.** `main` reads `importers:` only; there is no `snapshots` handling anywhere (`grep -l snapshots` on the production files returns nothing), so a dangling importer version is not cross-checked, and canonical keys like `react-dom@19.2.0(react@19.2.0)` are not decomposed.
3. **`npm-shrinkwrap.json`** — no occurrence on `main`.
4. **Duplicate-JSON-key detection with a nesting budget** — no equivalent on `main`; nothing guards `encoding/json`'s last-value-wins. Worth keeping, though I would like the PR to state the threat model explicitly: JS `JSON.parse` is also last-value-wins, so the divergence you are defending against is between *parsers* rather than between us and npm. Refusing a manifest we cannot read unambiguously is defensible on its own terms — just say that, rather than implying npm would read it differently.

So this is complementary, not superseded, and I would like it in.

### What I am asking for

Rebase onto `main` and reduce the branch to the delta above. Concretely, the parts that will collide and should give way to what is already on `main`:

- npm hoisting / install-path selection and pnpm `importers:` — `main`'s versions ship and are tested; please drop yours rather than reconciling two implementations
- alias resolution and SBOM component correlation — same
- the `resolver_types.go` / `resolver.go` surface — take `main`'s shape and add to it

What should survive: the Yarn path, pnpm `snapshots`/peer-qualified instances, shrinkwrap, and the strict-JSON duplicate-key guard, plus their tests.

### On the shape of the branch

19 production files and 19 test files is a lot of surface for one PR, and after the rebase it will be smaller but still broad. I would find it much easier to review — and you would get it merged faster — as two: **(a) Yarn support**, and **(b) pnpm snapshots + shrinkwrap + strict JSON**. They are independent, and (a) alone unlocks reachability for a whole package manager.

Also worth trimming: eleven files named `resolver_r2c_maintainer_*_test.go` is naming that describes the *review episode* rather than the behaviour under test. `resolver_yarn_test.go`, `resolver_pnpm_snapshots_test.go` and so on will age better.

I could not run the suite on this branch because it does not merge cleanly. Push the rebase and I will do the full verification — build, vet, full suite, `-race`, lint — and merge if it is green.

nghiadaulau (approved):
## Approve — and the guards here are actually tested, which I verified by breaking them

Reviewed strictly: checked out, built, ran the suite, then **mutation-tested the load-bearing checks**. Green tests alone say nothing; a guard that has never been seen failing is not yet a guard.

### The checks are real

Disabling the locator identity/version comparison — keeping only `!ok` so it still compiles:

```
--- FAIL: TestYarnBerryLocatorIdentityMismatchFailsClosed
--- FAIL: TestYarnBerryLocatorVersionMismatchFailsClosed
--- FAIL: TestYarnBerryUnrelatedBadLocatorPreservesPositiveButIncomplete
```

Accepting any `__metadata` version instead of the `{8, 10}` allow-list:

```
--- FAIL: TestYarnBerryUnknownMetadataVersionFailsClosed
```

Three independent tests for the identity check and one for the version gate. That is what coverage of a security-relevant parser looks like.

### The important design calls, and why they are right

**A descriptor whose locator names a different package or version is refused, not repaired.** This is the whole value of the PR: a lockfile is the one artefact that says which exact version a workspace resolves, so if its own cross-references disagree, the honest answer is to stop rather than pick the more plausible half. A mismatch here would attribute a version of one package to another — and every downstream CVE lookup would then be against the wrong component.

**An unknown `__metadata` version is refused with a coverage issue.** Exactly the right direction, and worth noting explicitly because it is the opposite of what #470 does with `lockfileVersion` (which fails *open* on an unrecognised value — see my review there). This PR shows the pattern the other one should adopt.

**A Yarn-selected version absent from the SBOM becomes `StatusUnresolved` + `CoverageMissingSBOMComponent`.** The comment says it best — *do not let SBOM-only correlation manufacture certainty*. Two pieces of evidence disagreeing must not silently collapse into whichever one is convenient.

**Hostile-input handling is careful.** `scanner.Err()` **is** checked and maps to `yarnLockInvalid`, so a single over-long line cannot truncate the parse and leave a partial read looking complete — that is the classic `bufio.Scanner` trap and it is not present here. Descriptor and entry budgets are enforced before allocation, tab indentation and a repeated `__metadata` section are refusals rather than best-effort guesses.

### One question, not a blocker

**The two open PRs answer "what if part of the lockfile is malformed?" differently.** Here, a malformed locator for one package preserves valid resolutions for others and marks coverage incomplete. In #470, one dangling pnpm snapshot discards the whole file's importer evidence. Both are honest — coverage is marked either way — but they are opposite policies and a reader comparing them will see arbitrariness.

I think both are actually defensible and the distinction is real: a Yarn lockfile is a flat descriptor→locator map whose entries are genuinely independent, whereas pnpm v9's `packages`/`snapshots` disagreement is a **structural** inconsistency in the file's own cross-references, which implicates the file rather than one entry. Worth writing that reasoning into a comment in one of the two, so the next person does not read it as an accident.

### Two minor observations

- `yarnNPMProtocolKeepsIdentity` returns `true` when `yarnAliasTargetName` yields `""`. I traced the cases: `""` means "a bare version range" (correct — `npm:^1.2.3` does preserve identity) or "a target that is not a parseable package name". The second is a narrow fail-open, but the resolution line is separately identity-verified, so nothing reaches a component on that path alone. Fine as it stands; just not obvious from reading the function alone.
- Yarn Berry 2/3 (`__metadata` version 6) gets no importer selection. That is honest incompleteness rather than a silent gap, but it is worth a line in the docs so the coverage report is not surprising.

### Before merging

The branch is **behind `main`** — GitHub reports `mergeable: UNKNOWN`, and CI has never run here (only GitGuardian, because the workflows were still disabled when the PR was opened). I checked with `git merge-tree` that it merges onto current `main` cleanly, so this is mechanical, but the code has not actually been *tested* against current `main`.

I am triggering the pipeline and will merge on green — not before. Note for sequencing: this and #470 both touch `resolver_lockfile.go`, so whichever lands second needs a rebase. Since #470 has a change requested, this one goes first.

nghiadaulau:
### Why this PR cannot be merged directly, and where it went (#476)

This PR never got a GitHub Actions check suite. It was opened while Actions was **disabled** on this repository, and GitHub does not create one retroactively — a synchronizing push does not fix it either. Its head commit lists GitGuardian alone, so its pipeline can never run, and I am not merging anything that has not been through CI.

I could not re-open a fresh PR from your fork branch (`Fork collab can't be granted by someone without permission`), so the branch is mirrored into this repository and the pipeline runs there. **Identical commits, your authorship preserved**, and it will be merged with a merge commit rather than a squash so each commit keeps you as author.

My full review — including the mutation-test evidence — stays on this PR; nothing about the code changed.

**One correction worth recording, because I got there the slow way.** I first blamed the fork-PR contributor-approval policy, then a missing check suite. Both were wrong. The real cause was that repository Actions had been switched to `enabled: false` partway through: no run object is created for *any* event, so `?status=action_required`, `waiting` and `queued` all report zero and it reads as a permissions puzzle rather than an off switch. The decisive probe was merging an unrelated PR to `main` and seeing that even **that** produced no run. Actions is back on now.

Mirror: #476

nghiadaulau:
Merged to `main` as e0c3293 via the CI mirror #476, with a **merge commit rather than a squash** so your commit `fbf6fde` keeps you as its author — it is on `main` unchanged.

Closing this one because the code has landed; it could not be merged from here since it never received an Actions check suite. The full review and the mutation-test evidence stay on this PR. Thanks — the locator identity/version verification and the "do not let SBOM-only correlation manufacture certainty" branch are the parts I'd point anyone else at as the model to copy.

nghiadaulau:
Merged to `main` as e0c3293 via the CI mirror #476, with a **merge commit rather than a squash** so your commit fbf6fde keeps you as its author — it is on `main` unchanged.

Closing this one because the code has landed; it could not be merged from here since it never received an Actions check suite. The full review and the mutation-test evidence stay on this PR.

Thanks — the locator identity/version verification and the "do not let SBOM-only correlation manufacture certainty" branch are the parts I would point anyone else at as the pattern to copy.

PR #463, opened 2026-08-09
feat(sca): Tier-1 reachability for Rust, PHP and Ruby (#414)

### LINKED ISSUE #414: feat(sca): reachability for Rust, PHP and Ruby
Parent: #405. Phase 2, Track B. Independent of the fleet work, so it can start now.

## Objective

Extend reachability analysis to Rust, PHP and Ruby, following the tier model already used for Go, the JVM and Python.

## Why this outranks new rule packs

Reachability is what converts a large vulnerability list into a queue someone will actually work. The market position is explicit about this. Endor Labs advertises call graph reachability across forty or more languages. Snyk supports Java and JavaScript only. Semgrep offers pattern based hints rather than a call graph. This project has a deterministic proof model that supersedes any model opinion, which is architecturally stronger than a hint, and it currently covers three language families. Every language added multiplies the value of detection breadth that already exists.

## Current state

- Tier 2, call path proof: Go through `synapse-callgraph` using `go/ssa`, and the JVM.
- Tier 1, import or module level: Python. JavaScript and TypeScript are in progress under #378 with #379, #400 and #401.
- The coordinator in `reachproof` already ranks a deterministic proof above any model produced opinion.

## Requirements

1. Ship Tier 1 for all three languages first. Tier 1 answers whether the vulnerable package is imported or required at all, on a real path, from first party code. That alone removes a large share of the queue.
2. Per language entry point discovery, because the wrong entry point set makes reachability meaningless:
   - Rust: binary and library targets from the manifest, plus test and bench targets excluded by default and included only on request.
   - PHP: front controller files, the autoload map, and framework route registration where it can be read statically.
   - Ruby: the load path, `require` and `require_relative`, plus Rails initialisers and route files where present.
3. Dynamic loading is reported, not guessed. Rust macro generated calls, PHP variable function names and includes with computed paths, Ruby `send`, `const_get` and metaprogramming all produce an explicit unknown region rather than a silent gap.
4. An unknown region degrades the verdict to unknown for anything that depends on it. Unknown is never rendered as unreachable. This is the non negotiable of the whole feature: an unreachable verdict suppresses work, so a false unreachable is worse than no verdict.
5. Confidence and tier are recorded on the result, with the evidence that produced it: the file, the symbol and the path. A verdict with no evidence is not stored.
6. Heavy analysis of untrusted source stays out of the server process. Any language toolchain invoked for analysis runs inside the sandbox as an argv binary, following the `synapse-callgraph` precedent. No language runtime is linked into the API server.
7. No network and no dependency resolution during analysis. If resolution is required to answer accurately, the answer is unknown with a coverage reason, following the resolution as filter pattern already used by the Go module graph and the Gradle resolvers.
8. Tier 2 for these languages is scoped as a follow up per language, with the decision recorded in the issue when Tier 1 lands. Do not promise Tier 2 for three languages in one issue.

## Out of scope

- C and C++ reachability. Separate issue, materially harder, and the build system diversity makes it a different problem.
- Any change to the reachability tier model itself.

## Specification sketch

```go
// per language, behind the existing port
type Resolver interface {
    EntryPoints(ctx context.Context, root string) ([]Symbol, []CoverageIssue, error)
    Imports(ctx context.Context, root string) (ImportGraph, []CoverageIssue, error)
}
```

Verdict rules:

| Condition | Verdict |
|---|---|
| Vulnerable package not imported on any path from an entry point | `unreachable` |
| Imported and the vulnerable symbol is referenced | `reachable`, Tier 1 |
| Imported but the symbol reference cannot be determined | `unknown` |
| Any unknown region on every candidate path | `unknown` |
| Entry point discovery incomplete | `unknown`, with the coverage reason |

## Acceptance criteria

- [ ] Tier 1 verdicts for Rust, PHP and Ruby on a fixture corpus with hand verified expected results, including at least one true reachable, one true unreachable and one genuine unknown per language.
- [ ] Every dynamic loading construct listed above produces an unknown region, verified per construct.
- [ ] No input produces `unreachable` when an unknown region exists on a candidate path. Asserted by a dedicated test for each language.
- [ ] Incomplete entry point discovery yields `unknown` plus a coverage reason, never a clean unreachable.
- [ ] All analysis executes argv only inside the sandbox, asserted by an execution path test.
- [ ] No network access during analysis, asserted with a blocked egress test.
- [ ] Verdicts appear in the existing finding and prioritisation surfaces with tier and evidence, with no new parallel display path.
- [ ] Measured on a real repository per language: the reduction in actionable findings is reported in the pull request, because the claim of this feature is queue reduction and it should be evidenced.

## Test requirements

1. Fixture corpora per language with expected verdicts committed alongside.
2. One negative test per dynamic construct asserting unknown.
3. A sandbox and egress assertion test.
4. A performance bound test on a large fixture, since a reachability pass that doubles scan time will be disabled by users.

### Existing surface to reuse (do not reimplement)
- `internal/domain/callgraph`, `internal/domain/reachability`, `internal/domain/reachproof`, `internal/domain/taint` are the domain primitives.
- `internal/usecase/reachability`, `internal/usecase/reachproof` and `internal/usecase/taintscan` are the coordinators; `reachproof.Coordinator` already ranks a deterministic proof above any model opinion.
- Migration `0042_finding_class_reachability.sql` already carries the reachability finding class. No new migration is expected for Tier 1; verdicts attach to the existing finding.

### New packages (one resolver per language)
- `internal/infrastructure/reach/rustresolve`, `.../phpresolve`, `.../rubyresolve`. Any language toolchain invoked for analysis runs as an argv binary inside the sandbox, following the `synapse-callgraph` precedent; nothing links a language runtime into the API server.
- A shared port `internal/usecase/ports/reachresolver.go`:
```go
type ReachResolver interface {
    EntryPoints(ctx context.Context, root string) ([]reachability.Symbol, []reachability.CoverageIssue, error)
    Imports(ctx context.Context, root string) (reachability.ImportGraph, []reachability.CoverageIssue, error)
    Lang() string
}
```

### Entry points per language
- Rust: `[[bin]]` and `[lib]` targets from `Cargo.toml`; test/bench excluded unless requested.
- PHP: front controller, `composer.json` autoload map, statically-readable framework route registration.
- Ruby: load path, `require` / `require_relative`, Rails initialisers and route files where present.

### Verdict table (normative, already in this issue)
`unreachable` is only permitted when no unknown region exists on any candidate path. The dedicated per-language test that asserts "unknown never resolves to unreachable" is the acceptance gate.

### Config
| Key | Default |
|---|---|
| `SYNAPSE_REACH_RUST` / `_PHP` / `_RUBY` | `false` (gated, reversible) |
| `SYNAPSE_REACH_TIMEOUT` | `120s` per analysis |

No network, no dependency resolution during analysis; if resolution is required to answer, the verdict is `unknown` with a coverage reason (resolution-as-filter, matching the Go module graph and Gradle resolvers).

nghiadaulau:
Closes #414 (Phase 2, Track B).

## What

Deterministic Tier-1 import-reachability for **Rust, PHP and Ruby**. A declared dependency that first-party source never references becomes a `not_reachable` judgment and, through the existing export path, an OpenVEX `not_affected` justification.

All three share **one** foundation rather than three parallel stacks:
- `ports.SourceImportScanner` — the observation contract
- `internal/infrastructure/tools/srcimports` — the per-language scanners
- `internal/usecase/srcreach` — a single Tier-1 analyzer parameterised by a per-language candidate namer

## Source-only, no sandbox needed

None of the scanners runs cargo, composer, bundler or a language runtime; none resolves a dependency over the network; nothing is compiled or executed. That is the same reasoning that keeps the Python and JavaScript scanners in-process, and it is why no sandbox is required here (contrast `synapse-callgraph`, which compiles untrusted source).

## Unknown never becomes unreachable

The non-negotiable of the issue. Each language's dynamic constructs produce an explicit **unknown region**, and one unknown region refuses the whole analysis rather than yielding a negative:

| Language | Constructs that degrade coverage |
|---|---|
| Rust | `macro_rules!`, `include!`, derive + procedural macros, FFI |
| PHP | `eval`, `call_user_func`, variable class name, variable variables, reflection, custom autoloader, computed include path |
| Ruby | `const_get`, `send`, `method_missing`, `define_method`, `instance_eval`, `class_eval`, `autoload`, computed/interpolated `require` |

Tested one case per construct, per language, plus a dedicated per-language assertion that an unknown region can never produce an unreachable verdict.

## Generous matching, on purpose

A dependency is referenced under several plausible names: a Rust crate written with hyphens is *used* with underscores, a Composer package name is not its namespace, and a gem's require path often differs from its gem name. Over-matching biases toward **reachable**, which is safe; under-matching would suppress a real finding.

## Entry points

From the manifests a reader can check a proof against: Cargo `bin`/`lib` targets (test and bench **excluded** — a test-only dependency is not evidence that shipped code reaches it), composer autoload roots, and the conventional Ruby load points.

## Attribution

Tier-1 proof identities are per-language (`system:rustimport-scan`/`-engine`, etc.), so a Rust proof is never attributed to the Python or JavaScript import engine in the audit log or the sealed rationale.

## Config

`SYNAPSE_REACH_RUST` / `SYNAPSE_REACH_PHP` / `SYNAPSE_REACH_RUBY`, each off by default and each also requiring `SYNAPSE_JUDGMENTS_ENABLED`.

## Verification

`go build ./...` · `go vet` · **golangci-lint 0 issues** · full suite **174 packages, 0 failures** · `-race` green.

## Not yet done from the issue's acceptance list

- The **measured reduction in actionable findings on a real repository per language** is not included. That is an empirical claim the issue rightly asks for, and I would rather leave it open than assert a number I have not measured.
- Tier-2 for these languages is explicitly scoped by the issue as a per-language follow-up.

### DIFF
--- a/cmd/synapse-api/main.go
+++ b/cmd/synapse-api/main.go
@@ -81,6 +81,7 @@ import (
 	"github.com/KKloudTarus/synapse-ce/internal/infrastructure/tools/risk"
 	"github.com/KKloudTarus/synapse-ce/internal/infrastructure/tools/sast"
 	"github.com/KKloudTarus/synapse-ce/internal/infrastructure/tools/secretscan"
+	"github.com/KKloudTarus/synapse-ce/internal/infrastructure/tools/srcimports"
 	"github.com/KKloudTarus/synapse-ce/internal/infrastructure/tools/syft"
 	"github.com/KKloudTarus/synapse-ce/internal/infrastructure/tools/taintcallgraph"
 	"github.com/KKloudTarus/synapse-ce/internal/infrastructure/tools/vexfile"
@@ -135,6 +136,7 @@ import (
 	"github.com/KKloudTarus/synapse-ce/internal/usecase/safety"
 	"github.com/KKloudTarus/synapse-ce/internal/usecase/sbomcrosscheckjudge"
 	scauc "github.com/KKloudTarus/synapse-ce/internal/usecase/sca"
+	"github.com/KKloudTarus/synapse-ce/internal/usecase/srcreach"
 	"github.com/KKloudTarus/synapse-ce/internal/usecase/taintscan"
 	threatmodeluc "github.com/KKloudTarus/synapse-ce/internal/usecase/threatmodeluc"
 	transferuc "github.com/KKloudTarus/synapse-ce/internal/usecase/transfer"
@@ -1215,6 +1217,41 @@ func main() {
 		}
 		scaService.SetJSReachability(jsRecorder)
 		log.Info("Tier-1 JavaScript import-reachability ENABLED (source-only, direct dependencies only → OpenVEX not_affected; best-effort)")
+	}
+
+	// Deterministic Tier-1 import-reachability for Rust, PHP and Ruby. Each scanner is SOURCE-ONLY: it
+	// lexes text and never runs cargo, composer, bundler or a language runtime, so no sandbox is needed
+	// and no dependency is resolved over the network. Each refuses a verdict whenever a dynamic
+	// construct (a Rust macro, a PHP variable class name, Ruby metaprogramming) could hide a reference.
+	// Composition-root only.
+	for _, lang := range []struct {
+		enabled  bool
+		env      string
+		label    string
+		purlType string
+		scanner  ports.SourceImportScanner
+		named    srcreach.CandidateNamer
+		language reachproof.Language
+	}{
+		{cfg.RustReachabilityEnabled, "SYNAPSE_REACH_RUST", "rust reachability", "cargo", srcimports.NewRustScanner(), srcimports.RustCandidates, reachproof.LanguageRust},
+		{cfg.PHPReachabilityEnabled, "SYNAPSE_REACH_PHP", "php reachability", "composer", srcimports.NewPHPScanner(), srcimports.PHPCandidates, reachproof.LanguagePHP},
+		{cfg.RubyReachabilityEnabled, "SYNAPSE_REACH_RUBY", "ruby reachability", "gem", srcimports.NewRubyScanner(), srcimports.RubyCandidates, reachproof.LanguageRuby},
+	} {
+		if !lang.enabled || !requireJudgmentsOrSkip(log, judgmentSvc != nil, lang.env, lang.label) {
+			continue
+		}
+		analyzer, aerr := srcreach.New(lang.scanner, lang.named)
+		if aerr != nil {
+			log.Error(lang.label+" analyzer init failed", "err", aerr)
+			os.Exit(1)
+		}
+		coord, cerr := reachproof.NewCoordinatorForLanguage(analyzer, judgmentSvc, auditLog, clock, judgment.Tier1, lang.language)
+		if cerr != nil {
+			log.Error(lang.label+" coordinator init failed", "err", cerr)
+			os.Exit(1)
+		}
+		scaService.SetSourceReachability(lang.purlType, coord)
+		log.Info("Tier-1 " + lang.label + " ENABLED (source-only dead-dependency detection → OpenVEX not_affected; best-effort)")
 	}
 
 	// Deterministic taint-analysis CapSAST proposals, opt-in. Builds the workspace call
--- a/cmd/synapse-api/main.go
+++ b/cmd/synapse-api/main.go
@@ -1240,7 +1240,11 @@ func main() {
 		if !lang.enabled || !requireJudgmentsOrSkip(log, judgmentSvc != nil, lang.env, lang.label) {
 			continue
 		}
-		analyzer, aerr := srcreach.New(lang.scanner, lang.named)
+		purlType := lang.purlType
+		reader := func(ctx context.Context, dir string) (map[string]bool, bool) {
+			return srcimports.DirectDependencies(ctx, dir, purlType)
+		}
+		analyzer, aerr := srcreach.New(lang.scanner, lang.named, reader)
 		if aerr != nil {
 			log.Error(lang.label+" analyzer init failed", "err", aerr)
 			os.Exit(1)
--- a/internal/infrastructure/tools/srcimports/php.go
+++ b/internal/infrastructure/tools/srcimports/php.go
@@ -1,0 +1,155 @@
+package srcimports
+
+import (
+	"context"
+	"encoding/json"
+	"strings"
+
+	"github.com/KKloudTarus/synapse-ce/internal/usecase/ports"
+)
+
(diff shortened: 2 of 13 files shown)


nghiadaulau:
Security audit run on this branch. It found the refusal architecture sound, but the **scanners under-detected what counts as an unknown region**, and the subject set included packages the technique cannot answer for. Every one of these would have suppressed real findings. All are now fixed with regression tests (`e42927e`):

| Finding | Impact if shipped |
|---|---|
| `reachproof.Language.Valid()` never listed rust/php/ruby | Enabling **any** of the three flags made the API `exit(1)` at startup — the feature had never executed |
| No direct-dependency guard | A lockfile-derived SBOM is fully resolved, so most subjects are **transitive**; first-party source never imports those → every one sealed `not_reachable` |
| `Bundler.require` / Rails loaders not unknown regions | A Rails app never writes `require "nokogiri"` → **every gem** reported unreferenced |
| Rust inline paths (`serde_json::from_str`) and attribute macros | Rust 2018 needs no `use` → crates the code demonstrably calls reported unused |
| PHP inline FQCNs (`new \Monolog\Logger`) and PHP 8 attributes | PHP needs no `use` either; `#[` was being stripped **as a comment**, deleting real code |
| Comment stripping ran *before* marker detection | Its quote-parity heuristic could **delete** a coverage marker, flipping a file from unknown to "fully observed" |
| Empty observation accepted as authoritative | A parser that understood nothing answered every subject "not referenced" |
| `activesupport` vs `active_support` | Missed exactly the highest-advisory Rails gems |

Markers are now detected on the **raw** body (a marker inside a comment is a false unknown — the safe direction), an empty observation refuses, and a subject that no first-party manifest declares is refused rather than answered.

Still outstanding from the audit, and I'd rather flag than quietly skip:
- **MEDIUM-1** file-type gaps (`.erb`/`.haml`/`.slim`, extension-less `Rakefile`/`config.ru`, `.phtml`/`.inc`) and no coverage reason when a skipped directory contained source.
- **MEDIUM-2** punctuation-exact dynamic markers miss paren-less forms (`obj.send :run`, `eval code`, `class_exists ($c)`).
- **MEDIUM-3** `phpHasComputedInclude` substring-matches `$includePath` and `'required|email'`, so PHP refuses almost always — safe, but the sealed reason is inaccurate.
- **MEDIUM-4** Rust `use ::serde_json::X`, multi-line root groups, and two `use` statements on one line.
- The issue's **measured reduction on a real repository per language** is not included.

Verified: build, vet, golangci-lint 0 issues, full suite 174 packages green, under `-race`.

PR #470, opened 2026-08-09
feat(reachability): harden npm and pnpm lockfile evidence

VietSory:
## Summary

Hardens the existing JS/TS lockfile resolver with the remaining non-overlapping cases identified during review of #457.

* validate pnpm v9 importer resolutions against both `packages` and `snapshots`
* support peer-qualified pnpm snapshot keys while preserving the canonical package identity
* fail closed on dangling, malformed, or version-inconsistent pnpm snapshot evidence
* support `npm-shrinkwrap.json` with the expected precedence over `package-lock.json`
* reject duplicate JSON object keys before `encoding/json` can apply last-value-wins semantics
* enforce a bounded nesting depth while validating hostile npm lockfile JSON
* add focused regression coverage for each of the above cases

This intentionally builds on the resolver implementation already merged to `main`; it does not reintroduce the overlapping npm importer, SBOM correlation, alias, or resolver-surface work from the original R2C branch.

## Validation

The resulting tree was validated against current upstream `main`:

* `go build ./...`
* `go vet ./...`
* full `go test ./...` on Linux
* `CGO_ENABLED=0 go build ./...`
* `golangci-lint`
* Windows build and vet
* `internal/infrastructure/tools/jsresolve` passes on Windows

The remaining Windows full-suite failures are existing filesystem-permission tests requiring Unix-style `0600` semantics, outside this change. The frontend failure is likewise the existing Node 20 / jsdom / undici startup failure and is unrelated to this Go-only diff.

Follow-up to #457.


nghiadaulau (changes requested):
## Requesting one change — the hardening switches itself off on a version bump

The work itself is good and the direction is right. I checked it out, ran the suite, and then **mutation-tested each guard** rather than accepting green tests as proof. Two findings, both reproduced with numbers. The first is why I am not approving yet.

### 1. Blocking: the pnpm cross-check is gated on an exact version string, and fails OPEN when it does not match

`isPNPMV9LockfileVersion` matches exactly `"9"` or `"9.0"`. Everything the PR adds — the `packages` × `snapshots` identity cross-check, `invalidV9Evidence` — hangs off that boolean. I ran **identical dangling evidence** through three lockfiles differing only in `lockfileVersion`:

```
lockfileVersion '9.0'   -> status=ambiguous  picked=
lockfileVersion '9.1'   -> status=component  picked=4.17.21
lockfileVersion '10.0'  -> status=component  picked=4.17.21
```

In all three the importer claims `lodash@4.17.21` while `packages:` and `snapshots:` contain only `4.17.20`. At `'9.0'` that is correctly refused. At `'9.1'` the resolver **picks a version that appears in neither section** and reports it as a resolved component — the exact evidence this PR exists to reject, accepted with no coverage issue and no signal.

Today nothing hits it: pnpm 10 still writes `'9.0'`. But the day pnpm bumps the format, this hardening disappears silently and the resolver gets *less* honest with no code change. A dormant fail-open is worse than a known gap, because nobody will be looking.

**Suggestion:** make the trigger structural rather than a version literal — if `snapshots:` is present, the cross-check applies. That is robust to any future numbering and needs no table of known versions. If you prefer to keep it version-based, parse the major and treat **>= 9** as v9-shaped, so drift errs toward strict.

### 2. Non-blocking: the strongest guard in the PR is the one with no test

`invalidV9Evidence` is never exercised. Both cases in `TestPNPMV9RequiresMatchingPackageAndSnapshot` declare a **single** dependency, so when the identity check `continue`s, `selected` is empty and the pre-existing `len(selected) == 0` return already handles it. I removed only the `invalidV9Evidence` term from the guard — **the whole suite stayed green.**

The behaviour it protects is real, and it is the right behaviour. With one valid dependency alongside one dangling one:

```
guard present  -> ambiguous   (whole-lockfile evidence discarded)
term removed   -> component   (the "good" dep from a proven-inconsistent lockfile is trusted)
```

Discarding everything is correct — an inconsistent lockfile cannot be trusted piecewise, because you cannot tell which half is lying. It just needs a two-dependency case so a future refactor cannot delete it in silence. (My first attempt at this probe was wrong, worth mentioning so the pattern is clear: I put one lodash version in the SBOM, so it resolved without the lockfile and told me nothing. The existing tests get this right — two versions, so the SBOM is ambiguous and only the lockfile can decide.)

### What is right, and worth saying

- **Shrinkwrap precedence, and refusing to fall back.** Observing `npm-shrinkwrap.json` and then silently reading `package-lock.json` would resolve against a file the package manager itself ignores. Reporting `CoverageUnsupportedPackageManager` instead of quietly using the weaker file is the honest branch, and `Lstat` keeps a symlink from being followed.
- **Duplicate JSON keys.** `encoding/json` last-value-wins means an attacker appends a second `"version"` and the resolver believes the last one while a human reading the file believes the first. For metadata that establishes *exact package identity* that is a real forgery primitive, and rejecting the document is the right call, not picking a winner.
- **The nesting bound is bounded by a byte budget too** — `readBoundedLockfile` caps input before the walk, so the per-object `seen` map cannot be blown up independently of file size. I checked that specifically.
- `link:`/`file:` specifiers are skipped without being marked invalid, which is correct: a workspace link is not dangling evidence.

**Minor:** `validateNoDuplicateJSONKeys` compares `err != io.EOF` directly; `errors.Is` is the safer habit even though `json.Decoder` returns the sentinel unwrapped today.

Happy to approve as soon as finding 1 is addressed — the rest of this is solid, and the two guards I could reach behave exactly as documented.

nghiadaulau (approved):
## Approving — I pushed the fix rather than leaving it blocked

Both findings from my review are addressed in `2f03bb0` on this branch, and the merge conflict with #457 is resolved.

**The fail-open is closed.** The `packages` × `snapshots` cross-check no longer hangs off a version literal. The trigger is **structural** — a `snapshots:` section is present — with the version consulted only as a fallback so a *declared* v9 lockfile with no snapshots section stays strict instead of being downgraded to a pre-v9 layout. Same dangling evidence as before:

```
before:  '9.0' -> ambiguous   '9.1' -> component 4.17.21   '10.0' -> component 4.17.21
after:   '9.0' -> ambiguous   '9.1' -> ambiguous           '10.0' -> ambiguous
```

**`invalidV9Evidence` now has a test.** `TestPNPMV9InconsistentLockfileDiscardsAllImporterEvidence` pairs a valid dependency with a dangling one, which is what the single-dependency cases could not reach. It asserts the whole lockfile's importer evidence is discarded — an inconsistent lockfile cannot be trusted piecewise, because you cannot tell which half is lying.

Both guards were re-verified by mutation: removing only the `invalidV9Evidence` term fails the new test, and restoring the version-literal gate fails the drift test on `'9.1'`, `'10.0'` and `'11'`. A guard nobody has seen fail is not yet a guard.

**Merge resolution.** This branch and #457 both edited `readImporterResolutions` and conflicted textually, not semantically. Both precedence rules are kept and both now apply: a root `packageManager` selecting Yarn excludes npm/pnpm lockfiles, and an observed `npm-shrinkwrap.json` excludes `package-lock.json`. The comment says so, so the next person does not read one of them as a leftover.

Everything I praised originally stands — the shrinkwrap precedence refusing to fall back, duplicate-JSON-key rejection as a forgery defence rather than a last-value-wins guess, and the nesting bound sitting behind a byte budget.

CI green (5/5). Merging.

PR #471, opened 2026-08-09
fix(ai): fail closed on malformed model output

XUanhoa04:
## Summary

Fail closed when AI providers return malformed decision fields or untrustworthy usage metadata.

Previously, out-of-range false-positive confidence and verifier scores were clamped into valid values, including converting oversized values to 100. Model aliases could also bypass the configured proposer/verifier separation check, and under-reported token usage could let an over-budget response execute a tool before the next budget check.

This PR also repairs two baseline CI portability issues exposed by the new checks: POSIX mode assertions now remain strict on Unix without misreading Windows ACL-backed files, and the frontend job uses the minimum Node version required by its locked `undici` dependency.

## Changes

- Reject missing or out-of-range false-positive confidence and judgment-verifier scores instead of coercing them.
- Canonicalize provider prefixes, case, and dated model aliases before enabling automated verification.
- Conservatively account for prompt, completion, and total token usage; reject malformed usage and stop before an over-budget tool call executes.
- Cap each budgeted model request to the remaining session token allowance.
- Prefer validated structured source locations during false-positive triage, with legacy title parsing only as a fallback.
- Bound and token-check model-proposed SAST claims and risk-narrative driver lists before persistence.
- Keep secret-file mode assertions at `0600` on POSIX systems while using platform-appropriate assertions on Windows.
- Run frontend CI on Node 22.19.0, the minimum version supported by the locked `undici@8.10.0` dependency.
- Add regression coverage for each fail-closed path and document the AI fix in the changelog.

## Validation

- `go test ./internal/domain/judgment ./internal/usecase/agenttools ./internal/usecase/findings ./internal/usecase/fptriage ./internal/usecase/llmverifier ./internal/usecase/orchestrator -count=1`
- `go test ./internal/usecase/taintscan ./internal/usecase/llmverifier ./cmd/synapse-api -count=1`
- `go test ./cmd/synapse-agent ./internal/infrastructure/fleetclient -count=1` on Windows
- `make build`
- `make typecheck`
- `pnpm dlx node@22.19.0 node_modules/vitest/vitest.mjs run` (31 files, 276 tests)
- `cd web && pnpm build`
- `git diff --check`

A local `go test ./... -count=1` run on Windows is still affected by Microsoft Defender quarantining a PHP scanner fixture containing malicious-code samples. The affected package passes in Linux CI, and every changed package passes independently.

## Checklist

- [x] `cd web && pnpm build` passes (dashboard unchanged)


nghiadaulau (approved):
## Approve — this closes a real hole, and I verified it by reintroducing it

Reviewed strictly, checked out locally, and **mutation-tested the guards** rather than trusting that green tests mean a real check.

### The clamp was a privilege escalation, not a rounding detail

`parseVerdict` clamped an out-of-range score into `0..100`. Restoring that clamp and re-running the suite:

```
--- FAIL: TestAutoVerifyRejectsOutOfRangeScore/101
    out-of-range score must fail closed, got {Attempted:1 Confirmed:1 ...}
```

**A model returning `101` became a confirmed judgment.** Clamping mapped malformed output onto the exact maximum, sailing past `verdict.EvidenceThreshold` (75) — so the one number that decides whether an AI claim gets promoted was the number most tolerant of garbage. Same shape in `parseCritique`: I made a missing confidence default to 100 and `TestParseCritiqueFailClosed` caught it. Both guards fail against the pre-fix code, which is the only evidence that they are guards.

The split between what stays strict and what may be normalized is drawn in the right place and stated in the comments: **verdict and confidence participate in the gate, the driver is an audit label.** Normalizing a label grants no authority; normalizing a score grants all of it.

### The alias bypass was the subtler one

`VerifierModel != LLMModel` is string inequality, so `openai/gpt-5` vs `gpt-5` — the same weights — read as two independent parties and satisfied ADR-0021's *distinct verifier* requirement. Routing it through `agent.SameModel` (provider prefix, case, dated aliases) fixes the check that the whole judgment gate rests on. `ConfiguredModelsDistinct` requiring both identities to be present is right: independence cannot be established from absent metadata.

### Token accounting

Treating provider usage as untrusted is correct, and the re-check placement is the point — **before** a proposed tool call is interpreted, not on the following turn, so the response that crosses the budget cannot still cause an action. I checked the clamp for an off-by-one: `overBudget` runs first and returns on `TokensUsed >= TokenBudgetMax`, so `remaining >= 1` and `MaxTokens` can never be sent as `0` (unbounded) or negative. Overflow checks on non-negative addition are correct.

`go build ./...`, `go vet`, and all seven changed packages pass locally.

---

## Two follow-ups — neither blocking

**1. The disable is silent, and this PR makes it fire more often.** A config of proposer `openai/gpt-5` + verifier `gpt-5` previously *enabled* auto-verification and now correctly does not — but `main.go` logs nothing on that path, so an operator sees automated verification simply stop. Gated judgments then wait for a human `Accept`, which is fail-closed but indistinguishable from a broken verifier. This repo's standard elsewhere is that a refusal explains itself; a `log.Warn` naming both models when they canonicalize to the same id would finish the change. Worth a small follow-up.

**2. `CanonicalModelID` only strips after the last `/`.** Bedrock-style identifiers (`anthropic.claude-…-v1:0`, `us.anthropic.…`) use `.` and `:`, so `anthropic.claude-opus-5` vs `claude-opus-5` still read as distinct and would let one model verify its own proposal. Pre-existing, and the port is documented as OpenAI-compatible, so out of scope here — but it is the same class of bypass this PR set out to close, so it belongs on the list.

**3. Housekeeping:** the description claims POSIX secret-mode assertions and the Node version bump, and neither is in this diff — both landed in #472. Not a defect, but a reviewer diffing the body against the changes will stumble on it; worth trimming so the next reader isn't hunting for changes that aren't there.

Merging once the pipeline is green.

PR #474, opened 2026-08-09
feat(asset): add asset-centric security management

nnatuan03:
## Summary

- add a business-level Asset entity above Engagement with lifecycle, metadata, ownership, and stable tenant-scoped identity
- add Asset CRUD, Project/technical-asset memberships, Engagement assignment, findings, coverage, posture, and assessment history APIs
- add Asset list/detail UI using the existing component library, tokens, routing, and interaction patterns
- retrofit the Engagement assessment graph and durable jobs for tenant-aware RLS enforcement
- preserve existing Engagements as unassigned for backward compatibility and retain imported-finding provenance and `unknown` posture semantics

## Database and compatibility

- adds migration `0066_business_asset_management.sql` after fleet rollout migration `0065`
- preserves legacy IDs, evidence/audit custody hashes, and existing assessment records
- removes the superseded Business Service write path and `member_of` relationship model

## Validation

- `go build ./...`
- `go test ./...`
- `go vet ./...`
- `golangci-lint v2.12.2`: 0 issues
- focused `go test -race` for Asset, HTTP, and memory persistence packages
- PostgreSQL migration `0065 -> 0066`, RLS isolation, legacy graph, and Asset membership tests on a clean database
- frontend typecheck
- frontend tests: 33 files, 280 tests passed
- frontend production build
- local browser flow: create Asset and open Asset detail successfully

Related to #405.


nghiadaulau (changes requested):
## Requesting changes — seven new broken Postgres tests, and CI structurally cannot see them

CI is green on this PR. **That green does not cover the riskiest thing in it.** `.github/workflows/ci.yml` has no Postgres service and never sets `SYNAPSE_TEST_DB_DSN`, so all **33** Postgres integration test files skip — including every test that touches migration 0066, tenant RLS, and the append-only evidence guard. So I ran them against real Postgres 17 myself, on **fresh** databases, and compared against `origin/main`.

### Baseline vs this branch, both on a clean database

`origin/main` — 2 failures, **both pre-existing** and unrelated to this PR (I confirmed these separately earlier):

```
--- FAIL: TestMigration0044
--- FAIL: TestMigration0056BackfillsOnlyCanonicalRelativePaths
```

This branch — the same 2, plus **7 new**:

```
--- FAIL: TestAgentDecisionStore_AppendIdempotentSeq
      null value in column "tenant_id" of relation "agent_sessions" violates not-null constraint (23502)
--- FAIL: TestAgentPlanStore_CreateThenCAS                      (same cause)
--- FAIL: TestPostgresEvidenceForkGuard                          tenant context is required
--- FAIL: TestPostgresEvidenceVerifyRoundTrip                    tenant context is required
--- FAIL: TestFindingRepository                                  tenant context is required
--- FAIL: TestAppendOnlyEnforcement                              tenant context is required
--- FAIL: TestMigration0066PreservesLegacyAssessmentGraph
      duplicate key value violates unique constraint "audit_chain_link_uniq" (23505)
```

Six are fixtures that were not updated for the new contract — `agent_sessions.tenant_id` is now `NOT NULL`, and the repositories now require a tenant bound in `ctx`. **The production paths are fine**: I checked that `httpapi/auth.go:89` and `cmd/synapse-worker/main.go:326` both call `shared.WithTenant`, so this is a test-fixture gap, not a runtime break. The seventh is different — `TestMigration0066PreservesLegacyAssessmentGraph` is **this PR's own new test**, and it fails on a fresh database, so it is not stale-state either.

### Why `TestAppendOnlyEnforcement` specifically makes this blocking

Migration 0066 is, as far as I can tell, the only migration in this repository's history that **deliberately switches off the append-only triggers**:

```sql
ALTER TABLE evidence DISABLE TRIGGER evidence_append_only;
ALTER TABLE audit_log DISABLE TRIGGER audit_log_append_only;
```

`TestAppendOnlyEnforcement` is the test that proves those triggers are back on — it asserts `UPDATE evidence` and `UPDATE audit_log` are refused by the database. **This PR breaks the one test that would notice if the triggers were left disabled**, and it breaks it in a suite CI does not run. Merged as-is, `main` shows green while golden rule 6 has no working proof, at exactly the commit that suspends it. That combination is what makes this a blocker rather than a cleanup task.

For the record, I did verify the two claims the migration rests on, rather than taking them on trust:

- **"tenant_id is not part of their custody hashes" — true.** `evidence.ComputeHash` binds `PreviousHash, Kind, FindingID, StorageRef, CreatedBy, CreatedAt, Content`, and the domain `Evidence` struct has no `TenantID` field at all. The blanket `UPDATE … SET tenant_id='default'` cannot invalidate a chain.
- **The DISABLE/ENABLE pair is transaction-safe.** There is no `-- +goose NO TRANSACTION`, so goose wraps the migration; a failure rolls the DISABLE back rather than leaving the guard off. Good, and worth the comment it already has.

### What is genuinely strong here

**Cross-tenant links are impossible at the schema level**, not merely checked in Go:

```sql
FOREIGN KEY (tenant_id, business_asset_id) REFERENCES fleet_business_services(tenant_id, id)
FOREIGN KEY (tenant_id, project_id)        REFERENCES projects(tenant_id, id)
```

and `TestBusinessAssetMigrationIsolation` asserts the FK violation (`23503`) on real Postgres. A composite FK is the strong form of tenant isolation: an application bug cannot forge a cross-tenant membership, because the database will not store one. This is the right way to do it, and it passed when I ran it.

**All 14 new routes go through the single `rt.authz(...)` chokepoint**, with `PermView` on reads and `PermOperate` on writes, and every handler passes `requestTenant(r)` explicitly down into the service rather than letting the store infer it. `requestTenant` funnels through `TenantOrDefault`, so an absent tenant becomes `default` rather than an empty string that would match a wildcard — the safe direction.

**`CALL synapse_enable_tenant_rls(...)` on the new tables**, and the postgres repository uses `WithTenant` throughout (16 call sites).

### Also worth fixing while you are here

`TestBusinessAssetMigrationIsolation` passed on a fresh database but failed on my **second** run against the same one (`duplicate key … fleet_business_services_pkey`). Its `t.Cleanup` does not fully undo its seed, which matters because these tests are meant to be runnable against a persistent dev database.

### The larger point, offered separately

The reason seven broken tests could reach a green PR is that **the Postgres suite does not run in CI at all**. That is not this PR's fault, but this PR is the one that demonstrates the cost. A `services: postgres:17` block in `ci.yml` with `SYNAPSE_TEST_DB_DSN` set would turn 33 skipped files into real coverage — and would have caught all seven of these before review. Worth its own issue; I am happy to open it.

Re-review as soon as the fixtures and the new 0066 test are green — the design underneath this is sound and I would like to merge it.

nghiadaulau:
Opened #478 for the underlying CI gap so it does not ride on this PR — that one is not your responsibility, but it is why these seven failures reached a green review.

nghiadaulau (approved):
## Approving — I pushed the fixes and re-measured against real Postgres

`d459d66` (rebased onto your `45f201b`) brings this branch to **parity with `main`** on the Postgres suite. Measured, on fresh containers:

| | run 1 | run 2 (same DB) |
|---|---|---|
| `origin/main` | 2 failures | 4 |
| this branch, before | **9** | 10 |
| this branch, now | 2 | 4 |

The remaining failures are identical to `main`'s and are not yours: `TestMigration0044` and `TestMigration0056…` on the first run, plus `TestAdvisoryRepository` and `TestScanJobStorePersistsDebugEvents` which only fail on a re-run. Those belong to #478.

### The fixture fixes

Six failures were fixtures that had not caught up with the new contract: repositories require a tenant bound in `ctx`, and `agent_sessions.tenant_id` is now `NOT NULL`. Seeds now state the tenant they act as — the same thing `httpapi/auth.go` and `cmd/synapse-worker` already do — and move off the empty partition that 0066 retires, including the `findings` row whose composite FK to `(tenant_id, engagement_id)` no longer matched its engagement.

That restores **`TestAppendOnlyEnforcement`**, which was the reason I blocked: it is the only test proving the append-only triggers are back on, and 0066 is the one migration that deliberately switches them off.

`audit_chain_link_uniq` turned out to be `UNIQUE(previous_hash)` **globally** — it is the fork guard, so the database has exactly one `previous_hash=''` slot. Your new 0066 test claimed it and collided with `integrity_test`; its seed now uses a per-run value, since `tenant_id` is what that test is about.

### Two real defects behind the fixtures, both fixed

**1. Migration 0066's `Down` could not run.** It restored `UNIQUE (tenant_id, name)` — which the `Up` drops *on purpose*, so two assets may share a display name and be keyed by `key` instead. Once anyone uses that permission the constraint no longer describes the data, and rollback failed with a bare `23505` from the index build that named neither the rows nor the reason. It now refuses with a message listing the colliding names.

A migration that suspends the append-only guard on `evidence` and `audit_log` needs a rollback that explains itself. Leaving the column silently non-unique instead would hand back a schema laxer than the one being rolled back to, which is its own surprise — so it refuses rather than quietly succeeding.

**2. The root cause of the whole cascade — and it was invisible by construction.** `business_asset_migration_test` used `defer pool.Close()`. A deferred close runs when the function **returns**, which is *before* any `t.Cleanup` callback, so every cleanup `DELETE` ran against a closed pool — and because the cleanup discarded its errors (`_, _ = pool.Exec`), it failed in total silence. The seeded pair of same-name assets survived, 0066 could then no longer roll back, and every later `goose.DownTo` failed with an error pointing nowhere near the cause. That is how one test-teardown bug became ten failures.

Cleanup now reports its failures, and the pool closes via `t.Cleanup` so LIFO ordering runs the deletes first. Making the teardown speak is what turned an inscrutable cascade into a one-line diagnosis — worth copying to the other integration tests.

### What I said before and still mean

The composite foreign keys are the strong form of tenant isolation: `(tenant_id, business_asset_id) REFERENCES fleet_business_services(tenant_id, id)` means an application bug **cannot** forge a cross-tenant membership, because the database will not store one. `TestBusinessAssetMigrationIsolation` asserts the `23503`, and it passes. All 14 routes go through the single `rt.authz` chokepoint, and I verified the two claims the migration rests on rather than trusting them: `tenant_id` is genuinely absent from `evidence.ComputeHash`, and the DISABLE/ENABLE pair is inside goose's transaction.

Merging once the pipeline finishes.

nnatuan03:
Dashboard/UI continuation is available in #480. PR #474 had already been merged and its remote branch deleted, so the follow-up reuses the same branch name on top of the latest main. It includes the Security Operations dashboard, grouped sidebar navigation, real Asset/Engagement/Fleet metrics, mobile theme fix, and 288 passing frontend tests.

Issue #417, opened 2026-08-07
feat(dast): crawler and first-party check corpus with proof per finding


nghiadaulau:
Parent: #405. Phase 3, Track C. Depends on authenticated scanning for full value but can be developed in parallel.

## Objective

Build a crawler and a first party check corpus so DAST covers a real application surface rather than a single probe, executed argv only inside the sandbox under the existing admission model.

## Current state

The DAST path today is one governed safe HTTP probe admitted through `internal/usecase/safety/gate.go`, producing a sealed proof and promoting through the judgment gate. The governance is right and the coverage is one request.

### Crawler

1. Discover the application surface from multiple sources rather than links alone: HTML links and forms, sitemap and robots hints, an OpenAPI or Swagger document when supplied, a GraphQL introspection response when permitted, and JavaScript sourced endpoints extracted statically.
2. Bounded by construction: maximum depth, maximum pages, maximum requests, maximum wall clock, and a per host request rate. Every bound has a conservative default and hitting one is reported as a coverage limit, not as completion.
3. Deduplicate URLs by a normalised form that collapses session identifiers and pagination parameters, so a calendar or an infinite pagination does not consume the whole budget.
4. Scope enforced per request, server side. A discovered link outside scope is recorded as out of scope and never requested.
5. Respect the deny list from the authenticated scanning issue, including logout and destructive verbs. Non idempotent methods are only sent when the operator explicitly enables them, and never during discovery.
6. Coverage is reported as a surface map: what was discovered, what was requested, what was skipped and why. A skipped area is never absent from the report.

### Check corpus

7. First party checks, clean room, organised by class: injection, broken access control, server side request forgery, insecure deserialisation, security misconfiguration, sensitive data exposure and authentication weakness. Start with the classes where a safe deterministic signal exists.
8. Every check declares its blast radius and whether it is safe against production. A check that could modify state is separately gated and off by default. This mirrors the production safe positioning that is now standard in this category.
9. Every check produces a proof class, not an opinion: the request, the redacted response, and the specific observation that supports the claim. A check that cannot produce a proof does not produce a finding.
10. Checks are versioned and catalogued exactly like the SAST rule catalogue, with the same catalogue tests, so drift between the catalogue and the implementation is caught in CI.
11. Every finding promotes only through the judgment gate on a distinct verifier's sealed verdict.
[...]

PR #475, opened 2026-08-09
feat(dast): add governed authenticated scanning and crawling (#416, #417)

### LINKED ISSUE #416: feat(dast): authenticated scanning with sessions from the credential vault
Parent: #405. Phase 3, Track C. Independent of the fleet work.

## Objective

Add authenticated scanning to the DAST path, with sessions acquired through the existing credential vault, so the majority of an application that sits behind a login can be assessed at all.

## Why this first in the DAST track

An unauthenticated scan reaches the login page and stops. Every commercial competitor treats authenticated scanning as the baseline. This project already has the parts that are normally the hard ones: a credential vault with server side placeholder substitution, an argv only sandbox with egress scoping, an admission model in `internal/usecase/safety/gate.go`, sealed proof evidence, and a judgment gate. What is missing is session handling.

## Requirements

1. Credentials come from the vault by reference. A credential value never appears in a work order, a log line, a model context, an evidence artifact or an error message. Substitution happens server side at execution time, following the pattern already in use.
2. Support the authentication schemes that cover most real applications: form login with a documented success signal, HTTP basic, a bearer token from a token endpoint, and a header or cookie supplied by the operator.
3. Session validity is verified, not assumed. Before any probe, a session check request confirms the session is live. A scan that proceeds with a dead session produces false negatives, so this is a correctness requirement rather than an optimisation.
4. Re authentication on session loss, with a bounded retry count. Exceeding it aborts the scan and reports incomplete coverage with the reason, never a clean result.
5. Logout and destructive path avoidance. A configurable deny list of paths and a default deny for anything matching logout, sign out, delete, purge and similar patterns. Hitting a logout link mid scan invalidates everything after it, which again is a correctness issue.
6. Scope enforcement is unchanged. Host, port and path scope plus the authorization window are enforced server side before any request is issued. An authenticated session does not widen scope.
7. Every request and response pair kept as evidence is redacted for credential material, session tokens and cookies before it is sealed. The redaction is tested with a real captured session.
8. Rate limiting and concurrency ceilings per target, configurable, with a conservative default. An authenticated scanner that overloads a customer application is a production incident.
9. Every claim still promotes only through the judgment gate on a distinct verifier's sealed verdict, exactly as the current safe probe path does.

## Out of scope

- Single sign on flows with an external identity provider, including multi factor. Documented as unsupported rather than partially attempted.
- Browser driven crawling, which belongs to the crawler issue.
- Any credential storage outside the vault.

## Specification sketch

```go
// internal/usecase/dast/session
type Scheme string // form, basic, bearer, header

type Config struct {
    Scheme        Scheme
    CredentialRef string        // vault reference, never a value
    LoginRequest  RequestSpec
    SuccessSignal Signal        // status, body match, header presence, cookie set
    CheckRequest  RequestSpec   // session liveness probe
    DenyPaths     []string
    MaxReauth     int
}
```

`SuccessSignal` is required. A login step with no verifiable success signal is a configuration error, because otherwise a failed login is indistinguishable from a successful one and the entire scan is silently meaningless.

## Acceptance criteria

- [ ] A form login scan against a fixture application reaches authenticated paths that the unauthenticated scan cannot.
- [ ] A login with wrong credentials is detected as failed and aborts the scan with a typed reason, rather than continuing unauthenticated.
- [ ] Session loss mid scan triggers re authentication, and exceeding the retry limit aborts with incomplete coverage.
- [ ] A logout path is never requested, verified with a request log assertion.
- [ ] No credential, cookie or session token appears in any log line, stored evidence artifact or error message. Verified by a test that greps the full captured output of a complete authenticated scan.
- [ ] An out of scope host is never contacted even when the authenticated session would permit it.
- [ ] A closed authorization window aborts before the first request.
- [ ] Rate limit and concurrency ceilings are honoured, verified with a counting fixture server.
- [ ] Findings promote only through the judgment gate with a distinct verifier.

## Test requirements

1. A fixture application in the test tree with a form login, a bearer endpoint, a basic realm and a logout path.
2. A redaction test over a real captured session, asserting zero credential material in evidence.
3. Negative tests for wrong credentials, dead session, exceeded re authentication and closed window.
4. A scope test asserting no out of scope request is issued.

### Existing surface to extend (do not bypass)
- `internal/usecase/dastrunner/service.go` already executes a bounded, argv-only probe via `Service.Execute(ctx, admitted safety.AdmittedAction, probe Probe)`, seals the proof through `evidence.Service`, and promotes through the judgment gate.
- `internal/usecase/dastverifier` and `internal/usecase/dastworkflow` are the verify and workflow layers.
- Credentials come from the existing vault via `ports.Credentials` with server-side placeholder substitution; a value never reaches a probe struct, a log, or an evidence artifact.

### New packages
- `internal/usecase/dast/session/service.go`: acquires and validates a session, then hands a live session context to `dastrunner`.
- `internal/domain/dastsession/config.go`: the typed `Config` in this issue (Scheme, CredentialRef, LoginRequest, SuccessSignal, CheckRequest, DenyPaths, MaxReauth). `SuccessSignal` is required; a config without it returns `shared.ErrValidation`.

### Persistence
Optional migration `00NN_dast_sessions.sql` for a per-engagement session config (referencing a vault credential by id, never a value). If the config is passed per-run instead, no migration is needed; decide in the PR and state it.

### Session liveness
Before every probe batch, the `CheckRequest` confirms the session is live. A dead session triggers re-auth up to `MaxReauth`; exceeding it aborts with incomplete coverage and a typed reason, never a clean result. A logout/destructive path in `DenyPaths` (default-deny for logout/signout/delete/purge patterns) is never requested.

### Redaction (a test over a real captured session)
Every stored request/response pair is redacted for credentials, session tokens and cookies before sealing, reusing the `internal/platform/redact` helper. Acceptance greps the full captured output of a complete authenticated scan and asserts zero credential material.

### Config
| Key | Default |
|---|---|
| `SYNAPSE_DAST_MAX_REAUTH` | `2` |
| `SYNAPSE_DAST_RATE` | `5 req/s/target` |
| `SYNAPSE_DAST_CONCURRENCY` | `4/target` |

Scope, authorization window and egress stay enforced server-side through `safety.Gate.Admit`; an authenticated session never widens scope.

pho-veteran:
## Summary

Adds governed authenticated DAST scanning with credential-vault sessions, bounded crawling, first-party checks, and sealed proof promotion. This combines and closes #416 and #417 while preserving server-side authorization and secret-handling invariants.

## Changes

- Add form, basic, bearer, header, and cookie session configuration backed by credential-vault references.
- Add the isolated `synapse-dast-helper`, per-request authorization pipes, pinned egress, redaction, shared session state, rate limits, and bounded reauthentication.
- Add deterministic bounded crawling with depth, page, request, and wall-clock limits plus coverage accounting.
- Add first-party DAST checks, sealed proof custody, distinct proof verification, and finding promotion.
- Add durable atomic approval consumption for memory and PostgreSQL stores.
- Wire authenticated DAST into the API, analysis pipeline, release packaging, and container image.
- Add domain, use-case, infrastructure, HTTP, persistence, and platform regression coverage.
- Make Unix permission assertions platform-aware and process-group tests container-correct.

## Checklist

- [x] `cd web && pnpm build` passes (dashboard unchanged; build verified)
- [x] Documentation updated if needed (no documentation changes required)

### DIFF
--- a/.goreleaser.yaml
+++ b/.goreleaser.yaml
@@ -52,6 +52,16 @@ builds:
     goarch: *build_goarch
     ignore: *build_ignore
 
+  - id: synapse-dast-helper
+    main: ./cmd/synapse-dast-helper
+    binary: synapse-dast-helper
+    env: *build_env
+    flags: *build_flags
+    ldflags: *build_ldflags
+    goos: *build_goos
+    goarch: *build_goarch
+    ignore: *build_ignore
+
   - id: synapse-worker
     main: ./cmd/synapse-worker
     binary: synapse-worker
@@ -88,6 +98,7 @@ archives:
     ids:
       - synapse-cli
       - synapse-api
+      - synapse-dast-helper
       - synapse-worker
       - synapse-callgraph
       - synapse-mcp
--- a/cmd/synapse-api/main.go
+++ b/cmd/synapse-api/main.go
@@ -27,6 +27,8 @@ import (
 	"github.com/KKloudTarus/synapse-ce/internal/infrastructure/acquire"
 	"github.com/KKloudTarus/synapse-ce/internal/infrastructure/blob"
 	"github.com/KKloudTarus/synapse-ce/internal/infrastructure/cache/sbomcache"
+	"github.com/KKloudTarus/synapse-ce/internal/infrastructure/dastchecks"
+	"github.com/KKloudTarus/synapse-ce/internal/infrastructure/dastengine"
 	"github.com/KKloudTarus/synapse-ce/internal/infrastructure/ebpf"
 	egressinfra "github.com/KKloudTarus/synapse-ce/internal/infrastructure/egress"
 	"github.com/KKloudTarus/synapse-ce/internal/infrastructure/fleetca"
@@ -104,6 +106,7 @@ import (
 	credentialsuc "github.com/KKloudTarus/synapse-ce/internal/usecase/credentials"
 	"github.com/KKloudTarus/synapse-ce/internal/usecase/crosscheckjudge"
 	dastrunneruc "github.com/KKloudTarus/synapse-ce/internal/usecase/dastrunner"
+	dastsessionuc "github.com/KKloudTarus/synapse-ce/internal/usecase/dastsession"
 	dastverifieruc "github.com/KKloudTarus/synapse-ce/internal/usecase/dastverifier"
 	dastworkflowuc "github.com/KKloudTarus/synapse-ce/internal/usecase/dastworkflow"
 	egresspolicy "github.com/KKloudTarus/synapse-ce/internal/usecase/egress"
@@ -1038,13 +1041,31 @@ func main() {
 					log.Error("DAST safe verifier runner init failed", "err", derr)
 					os.Exit(1)
 				}
-				dastWorkflowSvc, werr := dastworkflowuc.NewService(safetyGate, approvalSvc, approvalStore, dastRunnerSvc, clock, ids)
+				dastWorkflowSvc, werr := dastworkflowuc.NewService(safetyGate, approvalSvc, approvalStore, dastRunnerSvc, evidenceService, clock, ids)
 				if werr != nil {
 					log.Error("DAST verifier workflow init failed", "err", werr)
 					os.Exit(1)
 				}
 				router.SetDASTWorkflow(dastWorkflowSvc)
-				log.Info("DAST verifier workflow ENABLED (sandbox egress-enforced)")
+				engine, eerr := dastengine.New(reconRunner, cfg.DASTHelperBin, cfg.DASTMaxWallClock, cfg.ReconMaxOutput)
+				if eerr != nil {
+					log.Error("authenticated DAST engine init failed", "err", eerr)
+					os.Exit(1)
+				}
+				sessionSvc, serr := dastsessionuc.NewService(engine, reconGuard, evidenceService)
+				if serr != nil {
+					log.Error("authenticated DAST session init failed", "err", serr)
+					os.Exit(1)
+				}
+				ceilings := dastworkflowuc.DefaultScanCeilings()
+				ceilings.MaxReauth, ceilings.RatePerSec, ceilings.Concurrency = cfg.DASTMaxReauth, cfg.DASTRatePerSec, cfg.DASTConcurrency
+				ceilings.Limits.Depth, ceilings.Limits.Pages, ceilings.Limits.Requests, ceilings.Limits.WallClock = cfg.DASTMaxDepth, cfg.DASTMaxPages, cfg.DASTMaxRequests, cfg.DASTMaxWallClock
+				if err := dastWorkflowSvc.SetScan(sessionSvc, cfg.DASTHelperBin, evidenceService, dastchecks.NewEvaluator(), dastchecks.NewEvaluator(), judgmentSvc, runtimeVerifierSvc, ceilings); err != nil {
+					log.Error("authenticated DAST scan workflow init failed", "err", err)
+					os.Exit(1)
+				}
+				router.SetDASTScan(dastWorkflowSvc)
+				log.Info("DAST verifier and authenticated scan workflows ENABLED (sandbox egress-enforced)")
 			} else {
 				log.Warn("DAST verifier workflow DISABLED: sandbox kernel egress enforcement is unavailable")
 			}
--- a/cmd/synapse-dast-helper/main.go
+++ b/cmd/synapse-dast-helper/main.go
@@ -1,0 +1,23 @@
+// synapse-dast-helper owns HTTP clients and authenticated session state outside the API process.
+package main
+
+import (
+	"context"
+	"fmt"
+	"os"
+	"strings"
+
+	"github.com/KKloudTarus/synapse-ce/internal/infrastructure/dastengine"
+)
+
+func main() {
+	if len(os.Args) != 3 || os.Args[1] != "run" || !strings.HasPrefix(os.Args[2], "--config-sha256=") || len(strings.TrimPrefix(os.Args[2], "--config-sha256=")) != 64 {
+		fmt.Fprintln(os.Stderr, "usage: synapse-dast-helper run --config-sha256=<hex>")
+		os.Exit(2)
+	}
+	if err := dastengine.RunHelper(context.Background(), os.Stdin, os.Stdout); err != nil {
+		// RunHelper errors deliberately omit request headers, response bodies, and secrets.
+		fmt.Fprintln(os.Stderr, "synapse-dast-helper:", err)
+		os.Exit(1)
+	}
+}
--- a/deploy/Dockerfile
+++ b/deploy/Dockerfile
@@ -19,7 +19,8 @@ WORKDIR /src
 COPY go.mod go.sum ./
 RUN go mod download
 COPY . .
-RUN CGO_ENABLED=0 go build -trimpath -o /out/synapse-api ./cmd/synapse-api
+RUN CGO_ENABLED=0 go build -trimpath -o /out/synapse-api ./cmd/synapse-api \
+    && CGO_ENABLED=0 go build -trimpath -o /out/synapse-dast-helper ./cmd/synapse-dast-helper
 # synapse-cli is bundled too so the image doubles as a cross-platform scanner
 # (`docker run … synapse-cli scan /scan`) – the SCA path runs identically on any host
 # that runs Linux containers (Windows/macOS via Docker Desktop, Linux natively).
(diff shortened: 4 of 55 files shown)


nghiadaulau (changes requested):
## Requesting changes — one of the two headline paths cannot execute, and the test fake is why nobody saw it

First, the architecture here is genuinely good, and I want to be specific about that before the finding, because most of this PR is right.

### What is well built

**The out-of-process authorization broker is the right shape.** The helper owns the HTTP client, the API process owns the decision, and they talk over two pipes. The helper must ask before every request and the server answers through `execution.Guard` — so scope and the authorization window stay where golden rule 2 requires them, per request, and the scanner cannot self-authorize. Passing `EgressPolicy` to the runner as well means the kernel is the backstop if the helper ever misbehaves rather than the helper's own good manners.

**Every branch of the helper's authorizer fails closed.** Missing FDs, an encode failure, a decode failure, and `!decision.Allowed` all deny. Fail-closed on the *channel* being broken is the part that is usually missed.

**Redirects are not followed.** `CheckRedirect: http.ErrUseLastResponse` is what makes the `Location` authorization check meaningful — I checked this specifically, because if the client followed redirects internally the check would run *after* the fetch and a `302` to an internal host would be a scope escape with an audit trail that says it was authorized.

**`proofURL` redacts every query value rather than a denylist of names.** That is the right direction: a denylist misses `?tok=`. Userinfo and fragment are dropped too, so credentials cannot reach the server, the audit trail, or the outcome.

**A failed login is `Incomplete: true, Reason: "authentication_failed"`, not an empty result.** An unauthenticated scan of an authenticated surface reporting no findings would render as clean, which is the worst outcome available. Getting this right matters more than the checks themselves.

### Blocking: `Service.Execute` cannot succeed against the real engine

`ConfigDigest` is assigned in exactly **one** place in the whole PR:

```
internal/usecase/dastsession/service.go:194:  HelperBin: helperBin, ConfigDigest: configDigest,   <- crawl path
internal/usecase/dastsession/service.go:75:   (session/replay path — sets neither)
```

And the engine refuses any plan without it:

```go
if len(plan.ConfigDigest) != 64 {
    return ports.DASTOutcome{}, fmt.Errorf("%w: approved DAST configuration digest is required", shared.ErrValidation)
}
```

So the authenticated request-replay path — `Service.Execute`, one of the two features #416/#417 are about — fails with *"approved DAST configuration digest is required"* every time it is wired to `dastengine.Engine`. Only the crawl path can run. (`HelperBin` is missing on the same path but is benign; the engine tolerates empty and only refuses a mismatch.)

**Why the suite does not notice, which is the more useful half of this finding:**

```go
func (e *authorizingEngine) Run(ctx context.Context, _ ports.DASTPlan, _ []string, authorize func(...) error)
```

The fake **discards the plan**. It exercises the authorization callback thoroughly — which is why the scope tests are trustworthy — but it can never disagree with the real engine about what a valid plan is. Every test on this path passes, and `go test ./internal/usecase/dastsession/ ./internal/infrastructure/dastengine/ ...` is green on my machine, which is exactly why green was not enough here.

**Two things to fix, and the second is the one that lasts:**
1. Set `ConfigDigest` (and `HelperBin`) on the session path, from the same approved-configuration source the crawl path uses.
2. Give the fake at least the plan validation the real engine performs — or assert in one test that the plan `Service.execute` builds is accepted by `dastengine.Engine`. A fake that ignores the argument under test converts a contract into a decoration.

### Non-blocking

**Authorization is enforced in `send`, not in a `RoundTripper`.** Today `r.client.Do` appears exactly once and the check immediately precedes it, so the invariant holds. But it holds *by convention*: a second `client.Do` added anywhere in this package silently bypasses the guard, and this is the guard that keeps a pentest tool inside its authorized scope. Wrapping the client's `Transport` would make it unbypassable by construction, which is what an invariant this important deserves.

### Scope of this review, stated plainly

55 files and ~4400 lines. I reviewed the golden-rule surfaces in depth — process boundary and argv exec, per-request scope authorization, credential handling and redaction, fail-closed behaviour, and the session/crawl plan contract — and I read the domain validation for `dastsession`, `dastsurface`, and `dastcheck`. I did **not** review the crawler's HTML discovery, the check catalog's individual detections, the goreleaser/Dockerfile changes, or the judgment/finding wiring in comparable depth. Worth a second reviewer on the check catalog specifically, since that is where false positives would come from.

Happy to re-review promptly once the digest path is fixed.

nghiadaulau:
### Correction to my review: I over-stated the severity of the `ConfigDigest` finding

I said the replay path was "one of the two headline paths" and that a shipped feature could not execute. That was wrong, and the difference matters for how you prioritise it.

I traced the callers properly this time:

- `dastworkflow/service.go:134`'s `s.runner.Execute` is **`dastrunner`** — the safe-HTTP-probe path bound to `synapse-dast-probe` — not `dastsession.Service.Execute`. I conflated the two.
- **`dastsession.Service.Execute` has no production caller at all.** Only `service_test.go` calls it. Nothing routes to it, so no user-facing feature is broken today.

So the accurate finding is narrower but still real: `Service.Execute` is an **exported entry point that cannot work with the real engine**, and the reason nobody can tell is that its only tests run against a fake whose signature is `Run(ctx, _ ports.DASTPlan, ...)` — it discards the very argument that would fail. Dead-on-arrival API surface, not a broken feature.

**What is genuinely good here, and it is the reason the crawl path works:** the digest is not decoration. `dastworkflow/scan.go` canonicalises the approved scan configuration, hashes it, and then requires the stored approval's `Argv[2]` to equal `--config-sha256=<digest>` under `subtle.ConstantTimeCompare`. That binds a human approval to one exact configuration — approve a rate-limited crawl of one host and you cannot have it execute a different one. That is the right primitive, and it is why the engine insisting on a 64-char digest is correct rather than pedantic.

Which is also why I am not fixing this one for you, unlike the mechanical fixes I pushed to #470 and #474. There are two defensible resolutions and the choice is yours:

1. **Delete `Execute`/`execute`** if the crawl path is the intended way in. An exported method that cannot reach the engine is a trap for the next caller.
2. **Give it the same approval binding** — thread `helperBin` + `configDigest` through from an approval whose argv commits to the request set, exactly as `CrawlWithRate` does. Inventing that binding myself would be inventing security semantics for a feature you designed.

Either way, the durable fix is the second half of my original point: **make the test fake validate the plan the real engine validates**, or assert once that the plan a service builds is accepted by `dastengine.Engine`. A fake that ignores the argument under test turns a contract into a decoration — and that, not the missing digest, is what let this sit unnoticed.

The `RoundTripper` observation stands unchanged, and everything I praised in the first review I re-checked and still stand behind: the pipe-based authorization broker, fail-closed on a broken channel, `ErrUseLastResponse` so redirects cannot outrun the scope check, and `proofURL` redacting every query value rather than a denylist.

nghiadaulau (approved):
## Resolved in `a8e87b0` — and I was wrong to say this needed your decision

I said fixing the digest binding would mean inventing security semantics for your feature, so I would leave it to you. Looking again, **you had already decided it**:

```go
func (s *Service) Crawl(...) (dastcrawl.Result, error) {
	return dastcrawl.Result{}, fmt.Errorf("%w: DAST crawl requires an approval-bound helper and configuration digest", shared.ErrValidation)
}
```

`Crawl` already refuses rather than constructing an unbound plan. There was nothing for me to invent — only your own decision to apply to the sibling method it had not reached. `Execute` now refuses with the same wording, and `ExecuteWithBinding` carries `helperBin` + `configDigest` into the plan exactly as `CrawlWithRate` does. Refusing at the boundary beats failing deep inside the engine with a message about a digest the caller was never asked to supply.

### The durable half

`authorizingEngine.Run(ctx, _ ports.DASTPlan, ...)` discarded the one argument that would have failed, which is precisely why a plan the real engine rejects passed every test. It now mirrors the engine's plan validation. Verified by mutation: **removing `ConfigDigest` from the plan again fails two tests; before, it failed none.**

### This PR had never run CI

It was opened while repository Actions was disabled, so it had **no `github-actions` check suite at all** — only GitGuardian. Everything below was therefore invisible. Locally `golangci-lint` reported **8 issues** in this PR's code:

- five `errcheck` defers on `Close`
- `SA4032` — a `runtime.GOOS == "windows"` guard inside a file already carrying `//go:build !windows`, so the branch was unreachable
- `S1011` — a copy loop
- `G124` on the outgoing session cookie

**G124 I suppressed rather than "fixed", with the reason in the code.** It is a false positive here: `Secure`/`HttpOnly`/`SameSite` are `Set-Cookie` *response* directives, and `AddCookie` serialises only `Name=Value` — setting them would change nothing on the wire. Adding attributes to satisfy a linter, where they have no effect, would be worse than the warning.

`Lint (Go)` now passes on this PR.

### One thing I checked before deleting, and am glad I did

`discover` and `discoveredRequest` were flagged unused, and my first thought was that link-following was not implemented — which would have been a significant gap for something called a crawler. **It is implemented**, in `dastcrawl.Crawl`, which enqueues from `parseObservation` at `depth+1`. These two were a second, unused copy inside the helper. Deleting them without checking would have reached the right result for the wrong reason.

### Flagged, not changed

`TestRunHelperCrawlSharesBoundedReauthentication` failed once during a full-tree run and then passed 5/5 in isolation. It looks timing-sensitive under parallel load rather than incorrect, and it predates this commit — worth a look, but I did not want to change test semantics to chase it.

### Everything I praised in the first review still stands

I re-read each of these after making changes: the pipe-based authorization broker keeping the scope decision in the API process while the helper owns the HTTP client; fail-closed on *every* branch including a broken channel; `CheckRedirect: http.ErrUseLastResponse`, which is what makes the redirect authorization check meaningful rather than after-the-fact; `proofURL` redacting every query value instead of a denylist; and a failed login reported as `Incomplete` rather than as an empty, clean-looking result.

Merging when the pipeline finishes — the two backend jobs are still running as I write this.

PR #479, opened 2026-08-09
fix(ai-triage): keep verifier assessments blind

### LINKED ISSUE #390: feat(ai-triage): blind, provider-independent verifier (P1.3)

XUanhoa04:
## Summary

This hardens AI false-positive triage against two correlated-verifier bypasses:

- the verifier previously received the proposer verdict, driver, and confidence, which could anchor it toward a rubber-stamped consensus;
- Amazon Bedrock geographic/global inference-profile IDs could make the same model family appear distinct.

This advances #390 by completing the blind-prompt and fail-closed alias portions. It intentionally does not close #390: separate verifier endpoint/credential configuration and persisted provider metadata remain follow-up scope.

## Changes

- run the verifier on every candidate before the proposer result exists;
- send only the finding and source context to a neutral verifier prompt, then compare typed claims in Go;
- bump the auditable prompt contract to `fp-triage-v2`;
- canonicalize Bedrock geographic/global profile prefixes, provider namespaces, deployment revisions, and dated snapshots for separation-of-duties checks;
- reject missing or aliased model identities before enabling verification or granting `GateExempt`;
- warn API and CLI operators when model aliasing disables automated verification;
- add anchoring, provider-outage, missing-identity, Bedrock-alias, and forged-consensus regression coverage;
- update the changelog and AI-triage documentation.

The Bedrock normalization follows AWS documentation for [geographic inference-profile prefixes](https://docs.aws.amazon.com/bedrock/latest/userguide/geographic-cross-region-inference.html) and [global inference-profile IDs](https://docs.aws.amazon.com/bedrock/latest/userguide/global-cross-region-inference.html).

## Mutation checks

- Reintroducing the proposer signature into the verifier transcript makes `TestVerifierAssessmentIsBlindAndRunsBeforeComparison` fail and changes the simulated verifier from disagreement to rubber-stamped consensus.
- Disabling Bedrock scope normalization makes the domain alias test fail and lets the SCA policy tripwire observe `map[bedrock-alias:true]` from a forged consensus DTO.

## Validation

- `go build ./...`
- `go vet ./...`
- `go test -count=1 ./internal/domain/agent ./internal/usecase/fptriage ./internal/usecase/llmverifier ./internal/usecase/sca ./cmd/synapse-cli ./cmd/synapse-api`
- `go test ./...` passed for every available package except the local Windows SAST package: Microsoft Defender quarantines the pre-existing `php_lex_test.go`, causing `TestAnalyzerNormalizesNestedFindingPaths` to miss its PHP fixture. The quarantined deletion is not part of this commit; Linux CI is authoritative for the full suite.

## Checklist

- [x] CI equivalents of build, vet, full Go tests, and frontend typecheck pass
- [x] Dashboard unchanged; web typecheck passes
- [x] Documentation updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -47,6 +47,10 @@ capabilities below are already shipped on `main`.
 
 ### Fixed
 
+- **Blind AI verifier independence.** False-positive verification now runs on every candidate before the
+  proposer result is available and receives only the finding plus source context, preventing proposer
+  verdict anchoring. Provider/date aliases and Amazon Bedrock geographic/global inference-profile IDs
+  fail closed as the same model family, and operators are warned when aliasing disables verification.
 - **Fail-closed AI validation.** Reject missing or out-of-range false-positive confidence and verifier
   scores instead of coercing malformed model output into a valid judgment; canonicalize model aliases
   before enforcing proposer/verifier separation; account conservatively for missing, negative, or
--- a/cmd/synapse-api/main.go
+++ b/cmd/synapse-api/main.go
@@ -750,13 +750,15 @@ func main() {
 		} else {
 			coord := fptriage.New(tllm, cfg.FPTriageModel)
 			mode := "advisory-only (distinct verifier required for gate exemption)"
-			if cfg.VerifierModel != "" && cfg.VerifierModel != cfg.FPTriageModel {
+			if strings.TrimSpace(cfg.VerifierModel) != "" && agent.SameModel(cfg.FPTriageModel, cfg.VerifierModel) {
+				log.Warn("AI FP-triage verifier aliases the proposer; triage remains advisory-only",
+					"proposer_model", cfg.FPTriageModel, "verifier_model", cfg.VerifierModel,
+					"canonical_model", agent.CanonicalModelID(cfg.FPTriageModel))
+			} else if strings.TrimSpace(cfg.VerifierModel) != "" {
 				if vllm, verr := openai.New(cfg.LLMBaseURL, cfg.LLMAPIKey, cfg.VerifierModel, cfg.LLMTimeout); verr == nil {
 					coord.WithVerifier(vllm, cfg.VerifierModel)
 					if coord.VerifierModel() != "" {
 						mode = "verified by " + coord.VerifierModel()
-					} else {
-						log.Warn("AI FP-triage verifier aliases the proposer after canonicalization; triage remains advisory-only", "verifier_model", cfg.VerifierModel)
 					}
 				} else {
 					log.Warn("AI FP-triage verifier unavailable; triage remains advisory-only", "err", verr)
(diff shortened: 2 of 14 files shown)


nghiadaulau (approved):
## Approve, with one fix pushed as `4f3e236`

This closes both non-blocking items I raised on #471 — the silent disable now warns, and `CanonicalModelID` handles Bedrock-style identifiers — plus a third that I missed and you found: the verifier was reading the proposer's verdict. Thank you for including mutation checks in the description; that made this much faster to verify.

### The blind-verifier change is stronger than the description sells it as

The load-bearing guard is not the leak-detection string matching, it is this:

```go
if len(llm.reqs) != 2 || llm.reqs[0].Model != "verifier-model" || llm.reqs[1].Model != "proposer-model" {
```

**Running the verifier first makes anchoring structurally impossible** — you cannot put a value into a prompt before it has been computed. That is a far better invariant than inspecting the prompt for forbidden substrings, and it is what makes "distinct verifier" mean something rather than "second call to the same model with the answer attached". A correlated verifier that has seen the proposer's confidence is not a second opinion; it is the first one with extra steps.

### One thing worth knowing about your own leak detector

I tried to defeat it and it held on ordering, but not on content. `anchoringLLM` matches four literals — `"first reviewer's verdict"`, `"verdict=refuted"`, `"driver=not_reachable"`, `"confidence=92"`. Appending this to the verifier prompt passes every test:

```go
return userPrompt(f, snippet) + "\n\nAnother reviewer concluded: refuted (confidence 95)."
```

That is not a hole in the *proposer*-anchoring defence — ordering already makes that impossible — but it does mean **static anchoring language is untested**. A future prompt edit adding something like "findings of this kind are usually false positives" would anchor the verifier and nothing would notice. Asserting the verifier transcript contains **no verdict vocabulary at all** would be a positive invariant rather than a denylist of four strings. Same lesson as the scopes below, which is why I mention it.

### The fix I pushed: the scope allow-list failed open

`bedrockScope` covers `us/eu/apac/global`. Measured against `anthropic.claude-opus-5-v1:0`:

```
us. eu. apac. global.  -> SameModel=true
ca. jp. au. sa.        -> SameModel=false, canonical="ca.anthropic.claude-opus-5"
```

An unlisted region prefix leaves the **same model looking like an independent verifier** and passing the separation-of-duties check — precisely the bypass this canonicalization exists to close. AWS keeps adding cross-Region inference profiles, so the list will drift, and it drifts in the unsafe direction.

`4f3e236` inverts what gets enumerated: a leading segment is treated as a scope whenever **what follows it starts with a known provider namespace**. The provider set is what a Bedrock model ID is built from and it changes far more slowly than the region list, so unknown regions now collapse toward "same family" — fail-closed for a distinctness check. Verified by mutation: with the allow-list alone, the new test fails on `ca/jp/au/sa`.

The test also pins the **other** half, which matters just as much: `us.anthropic.claude-opus-5` and `us.meta.llama3-70b` must stay distinct. Collapsing everything would make every configuration look correlated and silently disable verification — a different failure, equally bad, and easy to introduce while fixing the first one.

### What else I checked

- The revision (`-vN:M`) and undelimited `YYYYMMDD` snapshot collapsing both err toward "same family", which is the right direction — two revisions of one model are not independent reviewers.
- Warning operators when aliasing disables verification is the piece I asked for on #471. A capability that turns itself off silently is indistinguishable from one that is broken.
- `go build`, `go vet`, `golangci-lint` (0 issues), and the full tree pass locally.

Merging when the pipeline is green.

PR #480, opened 2026-08-09
feat(web): add security operations dashboard

nnatuan03:
## Summary
- add a Synapse-native Security Operations dashboard using the existing UI components and design tokens
- aggregate real Asset, Engagement, and optional Fleet coverage data without hardcoded production metrics
- make `/dashboard` the default route and add `Command center → Dashboard` to the grouped sidebar
- preserve `New Engagement` before the Engagement list and fix mobile theme synchronization

## Asset Management relationship
Follow-up to #474, which was merged on August 9, 2026. The original PR and remote branch were already closed/deleted, so this PR reuses the same branch name on top of the latest `main`.

## Validation
- `pnpm run typecheck`
- `pnpm test` — 288 tests passed
- `pnpm run build`
- browser QA at 1440×900 and 390×844, including navigation, dark mode, empty states, Fleet-unavailable fallback, horizontal overflow, and console errors

### DIFF
--- a/web/src/App.rules.test.tsx
+++ b/web/src/App.rules.test.tsx
@@ -1,4 +1,4 @@
-import { render, screen, waitFor, fireEvent } from '@testing-library/react'
+import { render, screen, waitFor, fireEvent, within } from '@testing-library/react'
 import { describe, it, expect, vi, beforeEach } from 'vitest'
 import { MemoryRouter } from 'react-router-dom'
 import App from './App'
@@ -11,6 +11,7 @@ vi.mock('./lib/api', () => ({
     getRule: vi.fn(),
     listEngagements: vi.fn(),
     listBusinessAssets: vi.fn(),
+    fleetCoverageSummary: vi.fn(),
     listProjects: vi.fn(),
     getProject: vi.fn(),
     getAuditLogs: vi.fn(),
@@ -33,6 +34,7 @@ describe('App Routing - Rules', () => {
     vi.mocked(api.listRules).mockResolvedValue([])
     vi.mocked(api.listEngagements).mockResolvedValue([])
     vi.mocked(api.listBusinessAssets).mockResolvedValue({ items: [], total: 0, limit: 200, offset: 0 })
+    vi.mocked(api.fleetCoverageSummary).mockResolvedValue({ agentsByState: {}, rowsByVerdict: {}, oldestPerCapability: {}, assetsWithoutAgent: 0 })
     vi.mocked(api.listProjects).mockResolvedValue([])
     vi.mocked(api.getRule).mockResolvedValue({
       key: 'go:sql',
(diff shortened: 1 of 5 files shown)


nghiadaulau (approved):
## Approve — and this gets the hard part right, which is what a dashboard usually gets wrong

**CI is disabled on this repository, so I ran the pipeline locally instead of trusting a green tick.** All four steps of the `Frontend (web)` job, on this branch:

```
pnpm install --frozen-lockfile   ok
pnpm test                        35 files, 288 tests passed
pnpm run typecheck               clean
pnpm build                       built
```

### The thing I went looking for, and you had already done

A dashboard is the easiest place in a security product to lie by omission. If the fleet API is down and the tile reads **`0` coverage gaps**, an operator sees a healthy fleet. You render `—`:

```tsx
value={coverageGaps ?? '—'}
hint={fleetUnavailable ? 'Fleet telemetry unavailable' : 'All non-covered states'}
```

and `coverageGaps` is `null` — not `0` — when `fleet` is absent, so there is no path where an outage becomes a reassuring number. The panel then says it out loud: *"Coverage is not assumed clean."*

I mutation-tested it rather than taking the code's word. Changing `?? '—'` to `?? 0`:

```
× keeps core operations visible when Fleet telemetry fails
```

So the guard is real, and it is asserted on the **rendered label** (`Coverage gaps: —`) rather than on internal state — which is where it matters, because the operator reads the label.

**Partial failure is scoped correctly too.** Fleet telemetry is fetched separately from assets and engagements, so a fleet outage degrades one card instead of blanking the page, while still refusing to imply coverage. Failing the whole dashboard would have been the easy choice and a worse one; silently zeroing the tile would have been the easiest and the worst.

### Also right

- **All three states present**: `Spinner` while loading, `ErrorState` on failure, `QueueEmpty` with an actionable hint for an empty queue.
- `POSTURE_WEIGHT` ranks `unknown: 2` **above** `good: 1`. Unknown posture sorts as more concerning than healthy posture — the honest ordering, and easy to get backwards.
- Truncation is disclosed: *"Posture distribution reflects the first N loaded Assets."* A silently truncated distribution reads as the whole population.
- **No raw hex anywhere** — design tokens throughout. Icons are `lucide-react` only, no emoji-as-icon.
- a11y is not an afterthought: `aria-labelledby` on each section, `aria-hidden` on every decorative icon, and `aria-label` on the metric cards carrying label *and* value, which is why the test can assert `Coverage gaps: —` in the first place.
- `useEffect` cleans up with an `active` flag, so a fast unmount cannot set state on a dead component.

### Notes, neither blocking

- The queues cap at `.slice(0, 6)`, so the >50-row virtualization rule in `docs/04-ui-ux.md` does not apply here. Fine as-is — worth remembering if these lists ever grow a "show all".
- `pnpm build` warns that a chunk exceeds 500 kB. Pre-existing and not caused by this PR, but the dashboard is now the default route, so first paint is the one place it will be felt. Worth a look someday, not here.

Merging.

Issue #478, opened 2026-08-09
ci: the Postgres integration suite never runs (33 test files skip silently)


nghiadaulau:
## The Postgres integration suite never runs in CI

`.github/workflows/ci.yml` has no Postgres service and never sets `SYNAPSE_TEST_DB_DSN`, so **33 test files skip silently** on every run. A skipped test reads exactly like a passing one in the job summary — which is the failure mode this project refuses everywhere else in its own product: unknown must not render as clean.

### What is not covered today

Everything only a real database can prove:

- **Row Level Security and tenant isolation** — `synapse_enable_tenant_rls`, `WithTenant`, cross-tenant `ErrNotFound`, and the composite-FK guards that make a cross-tenant link unstorable.
- **DB-enforced append-only evidence and audit** — `TestAppendOnlyEnforcement` asserts `UPDATE evidence` and `UPDATE audit_log` are refused by triggers. Golden rule 6 has no CI proof at all.
- **Every migration**, including the fork guard, anchor/tail-truncation logic, and the backfills. Migrations are **auto-applied at startup**, so a broken one is found by a deployment rather than by a pipeline.

### Concrete evidence that this costs real defects

Reviewing #474 I ran the suite against Postgres 17 on fresh databases and compared with `origin/main`:

| | failures |
|---|---|
| `origin/main` | 2 (both pre-existing: `TestMigration0044`, `TestMigration0056…`) |
| #474 | 9 — the same 2 plus **7 new** |

The seven include four `tenant context is required` failures, two `agent_sessions.tenant_id` NOT-NULL violations, and **that PR's own new migration test**. All invisible to a green CI run. One of them, `TestAppendOnlyEnforcement`, is the only test proving the append-only triggers are re-enabled — and #474's migration is the one that deliberately disables them.

Two of the failures are older than #474 and have been sitting on `main` unnoticed, which is the same point from the other direction.

### Proposed fix

Add a Postgres service to the `Backend (Go)` job:

```yaml
services:
  postgres:
    image: postgres:17
    env:
      POSTGRES_PASSWORD: ${{ secrets.CI_PG_PASSWORD }}
      POSTGRES_DB: synapse
    ports: ["5432:5432"]
    options: >-
      --health-cmd pg_isready --health-interval 10s
      --health-timeout 5s --health-retries 5
```

[...]

PR #481, opened 2026-08-09
feat(ai-triage): add human-review API and UI (P1.2)

### LINKED ISSUE #389: feat(ai-triage): human-review API and UI (P1.2)

H1eu232:
Closes #389
Parent epic: #387

## Summary

- materialize every `review_required` AI-triage result into a durable, tenant-scoped review queue after scan evidence is sealed
- add list, claim, accept, and reject APIs with RBAC, optimistic concurrency, ownership, separation of duties, and evidence-linked audit metadata
- persist policy/prompt/model identity and P1.1 shadow-rollout fields; preserve review owner and advance version on rescans
- add the review queue UI, filters, decision rationale, AI-triage badges, and full DTO mapping without hiding findings from existing views or reports
- add migration 0067 with composite tenant constraints and forced RLS, plus migration assertions

## Safety properties

- AI cannot approve its own recommendation; machine/model actors cannot decide
- accept is persisted and audited before a finding can be marked false-positive
- reject moves the finding back to gating first
- missing evidence, audit failures, store failures, stale versions, and owner mismatches fail closed
- this PR does not change the deterministic P0 gate policy

## Integration notes

- rebased onto latest `main`, including P1.1 (#477) and P1.3 (#479)
- the current frontend does not include `@tanstack/react-query` or `docs/04-ui-ux.md`; the queue follows the repository's existing API/loading patterns and design-system tokens

## Verification

- `go build ./...`
- `go vet ./...`
- all Go package tests across `cmd` and `internal`
- focused `go test -race` for AI review, SCA, and HTTP paths
- `golangci-lint v2.12.2 run` — 0 issues
- `pnpm typecheck`
- `pnpm test` — 35 files / 285 tests
- `pnpm build`
- `git diff --check`

The PostgreSQL migration integration test runs when `SYNAPSE_TEST_DB_DSN` is available; without it, the test compiles and skips locally.

### DIFF
--- a/api/openapi.yaml
+++ b/api/openapi.yaml
@@ -1006,6 +1006,68 @@ paths:
       requestBody: {required: true, content: {application/json: {schema: {type: object, required: [asset_id], properties: {asset_id: {type: string, description: Empty string clears assignment}}}}}}
       responses: {'204': {description: Assignment updated}, '400': {$ref: '#/components/responses/BadRequest'}, '404': {$ref: '#/components/responses/NotFound'}}
 
+  /api/v1/ai-triage/reviews:
+    get:
+      tags: [ai-triage]
+      operationId: listAITriageReviews
+      summary: List the tenant's durable human-review queue
+      parameters:
+      - {name: severity, in: query, schema: {type: string, enum: [critical, high, medium, low, info, unknown]}}
+      - {name: cwe, in: query, schema: {type: string}}
+      - {name: project, in: query, schema: {type: string}}
+      - {name: state, in: query, schema: {type: string, enum: [pending, accepted, rejected]}}
+      responses:
+        '200':
+          description: Review queue
+          content:
+            application/json:
+              schema:
+                type: object
+                required: [reviews]
+                properties:
+                  reviews: {type: array, items: {$ref: '#/components/schemas/AITriageReview'}}
+        '400': {$ref: '#/components/responses/BadRequest'}
+        '401': {$ref: '#/components/responses/Unauthorized'}
+        '403': {$ref: '#/components/responses/Forbidden'}
+
+  /api/v1/ai-triage/reviews/{rid}/claim:
+    parameters: [{$ref: '#/components/parameters/AITriageReviewId'}]
+    post:
+      tags: [ai-triage]
+      operationId: claimAITriageReview
+      summary: Assign an unowned review to the authenticated reviewer
+      requestBody:
+        required: true
+        content:
+          application/json:
+            schema: {$ref: '#/components/schemas/AITriageReviewClaimRequest'}
+      responses:
+        '200': {description: Claimed review, content: {application/json: {schema: {$ref: '#/components/schemas/AITriageReview'}}}}
+        '400': {$ref: '#/components/responses/BadRequest'}
+        '401': {$ref: '#/components/responses/Unauthorized'}
+        '403': {$ref: '#/components/responses/Forbidden'}
+        '404': {$ref: '#/components/responses/NotFound'}
+        '409': {description: Version or ownership conflict, content: {application/json: {schema: {$ref: '#/components/schemas/Error'}}}}
+
+  /api/v1/ai-triage/reviews/{rid}/decision:
+    parameters: [{$ref: '#/components/parameters/AITriageReviewId'}]
+    post:
+      tags: [ai-triage]
+      operationId: decideAITriageReview
+      summary: Accept or reject an owned AI false-positive recommendation
+      requestBody:
+        required: true
+        content:
+          application/json:
+            schema: {$ref: '#/components/schemas/AITriageReviewDecisionRequest'}
+      responses:
+        '200': {description: Decided review, content: {application/json: {schema: {$ref: '#/components/schemas/AITriageReview'}}}}
+        '400': {$ref: '#/components/responses/BadRequest'}
+        '401': {$ref: '#/components/responses/Unauthorized'}
+        '403': {$ref: '#/components/responses/Forbidden'}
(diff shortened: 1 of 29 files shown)


nghiadaulau (approved):
## Approve — two fixes pushed as `6498cb6`

**CI is disabled on this repository, so I ran the pipeline myself** rather than reporting a green tick that does not exist. On this branch, against real Postgres 17:

```
go build ./...            ok        pnpm install --frozen-lockfile  ok
go vet ./...              ok        pnpm test        35 files, 285 tests
golangci-lint             0 issues  pnpm typecheck   clean
go test ./... (+Postgres) see below pnpm build       built
```

Postgres suite: only the four failures `main` already carries (`TestMigration0044`, `TestMigration0056…`, and on a re-used database `TestAdvisoryRepository`, `TestScanJobStorePersistsDebugEvents`) — all tracked in #478. **No new failures once the two items below are fixed.**

### 1. The machine-principal denylist missed `system:`, and the chain is two steps

`machineActor` covered `agent:`, `llm:` and `mcp:`. But `system:` is the **largest machine family this codebase mints** — `system:dast-engine`, `system:dast-verifier`, `system:taint-scan`, `system:cross-check`, `system:callgraph-engine`, and more.

Testing `Decide` alone hides it, because the owner check refuses first with *"must be decided by its assigned owner"*. The reachable path is claim-then-decide:

```
step 1  CLAIM by "system:dast-engine" SUCCEEDED, owner="system:dast-engine"
step 2  a machine principal accepted its own AI-triage review  (state=accepted)
```

The engine becomes the owner, and then the owner check is satisfied *by itself*. With `system:` added, step 1 is refused: *"an AI or machine principal cannot own its triage review."*

**To be fair about severity: this is not reachable over HTTP.** RBAC is the first line and holds — `user.rolePermissions` grants machine and unknown roles nothing, and the routes require `PermReview`. But this domain check exists *precisely* as the second line for in-process callers, and a second line that misses the widest class of machine identity is not a second line. Your own `TestClaimAssignsOnlyHumanReviewer` had the right instinct; it just tested `agent:bot`.

**The durable fix is inversion**, and I deliberately did not do it: a prefix denylist fails open on the next identity scheme somebody invents, but requiring an actor to be a *known human principal* needs a definition of human identity this package does not own. Added the missing prefixes plus a test covering every family, and left the inversion as a follow-up rather than guessing at it.

### 2. The TRUNCATE assertion was measuring a side effect

Migration 0067's foreign key to `evidence` changed what a plain `TRUNCATE` reports:

```
before: append-only: TRUNCATE on evidence is forbidden (custody chain, golden rule 6)
after:  cannot truncate a table referenced in a foreign key constraint
```

`TestAppendOnlyEnforcement` broke. **The property still held** — evidence is still untruncatable — so this is not a regression in your PR. But it exposed that the assertion was fragile in a way worth fixing rather than loosening: **the foreign key is incidental protection.** It works only because some other table happens to reference `evidence`, and it evaporates the day that table does not.

`TRUNCATE evidence CASCADE` is the variant an FK cannot stop — it truncates the referencing table too, leaving only the trigger:

```
TRUNCATE evidence          -> refused by the foreign key
TRUNCATE evidence CASCADE  -> refused by the trigger: "...golden rule 6"
```

The test now accepts either mechanism for the plain case and requires the **trigger's own refusal** for CASCADE. Verified by dropping `evidence_no_truncate`: CASCADE then succeeds and wipes evidence, the new assertion catches it, and the old one would have passed. So this turned a test that had quietly become a false negative into one that detects actual loss of the guarantee.

### What is strong here

- **`Decide` validates the whole envelope, not just the decision**: pending state, optimistic version, valid decision enum, non-empty actor, machine refusal, `sameIdentity` against *both* proposer and verifier model (so a human cannot borrow a model's name), owner match, rationale length **and** control-character rejection, non-zero timestamp. That is the right density for the one call that can flip a finding to false-positive.
- **Ordering is correct where it matters**: accept is persisted and audited before the finding is marked, and reject moves the finding back to gating first. A crash between the two leaves the finding gating, not exempt.
- **Composite tenant constraints and forced RLS** on migration 0067, consistent with 0066 — a cross-tenant review row is unstorable rather than merely unqueried.
- Findings are not hidden from existing views or reports while awaiting review, so the queue adds a step without creating a blind spot.

### Note

The description says the frontend has no `docs/04-ui-ux.md`. That document does exist and is referenced from `CLAUDE.md` — worth reading before the next UI change, though nothing in this PR conflicts with it: I checked for raw hex and non-lucide icons and found none.

Merging.

PR #482, opened 2026-08-09
feat(redteam): add deterministic attack path graph

### LINKED ISSUE #419: feat(redteam): attack path graph over the asset inventory with provenance per edge

pho-veteran:
## Summary

Implements #419 by correlating the tenant-scoped fleet inventory, governed findings, and confirmed reachability evidence into deterministic, explainable attack paths. Every step retains producer provenance and closed observed/inferred confidence, while incomplete evidence is surfaced as explicit uncertainty rather than promoted to confidence.

## Changes

- Adds the pure attack-path graph, deterministic ranking and stable IDs, cycle-safe traversal, AND-composable filters, and enforced path-length/path-count/wall-clock bounds.
- Adds tenant-scoped derived asset-to-finding bindings with memory and PostgreSQL adapters, RLS/FK enforcement, producer-scoped atomic replacement, and concurrency-safe projection claims.
- Wires authoritative attribution across SCA, manual findings, exploitation, verified SAST/DAST/threat findings, and SARIF imports, including canonical/imported finding identity separation and retry-safe partial writes.
- Exposes `GET /api/v1/attack-paths`, configuration limits, OpenAPI documentation, hostile-harness coverage, and a synthetic estate with 10,000 assets.
- Preserves upstream business-asset management and authenticated DAST behavior after rebasing; issue migrations occupy versions 0067–0070.

## Checklist

  - Linux `go build ./...` and `go vet ./...` pass.
  - Issue/conflict-focused Go tests and frontend `tsc --noEmit` pass.
  - Full Linux tests pass except an unchanged upstream DAST timing test that is flaky under full-suite load; its isolated run passed 5/5.
  - No dashboard code changed; the production web build also passes.

Closes #419


nghiadaulau (approved):
## Approve — with a migration renumber and five test fixes pushed as `5b61650`

**CI is disabled on this repository, so I ran the pipeline myself** rather than reporting a tick that does not exist. Against real Postgres 17, on fresh containers:

```
go build ./...      ok        golangci-lint   0 issues
go vet ./...        ok        migrations      apply cleanly to version 71
```

The branch is now at **parity with `main`** — only `main`'s own pre-existing failures remain (`TestMigration0044`, `TestMigration0056…`, the re-run pair tracked in #478, and the DAST timing test you also identified as flaky under load; my run agrees with your 5/5 isolated result).

### Blocking when I started: migration 0067 collided

#481 merged `migrations/0067_ai_triage_reviews.sql` while this branch carried `0067_fleet_asset_edge_confidence.sql`. Both branches also had a `migration_0067_test.go`. **Migrations are auto-applied at startup**, so a duplicate version is a deployment failure, not a merge inconvenience.

Renumbered 0067→0068, 0068→0069, 0069→0070, 0070→0071, renamed the affected test files, and moved every `goose` `UpTo`/`DownTo` target with them. One needed thought rather than a mechanical bump: the down-migration test descended to **66**, which was one below when this was 0067. It now descends to **67** — going lower would revert #481 as well and change what the test exercises.

### Your Postgres tests had never executed

They skip without `SYNAPSE_TEST_DB_DSN`, and CI does not set it (#478), so **five of this branch's own new tests failed** the first time they ran. Four were unbound-tenant contexts, which the repositories refuse so that no query can silently escape RLS. Two of those needed judgment rather than a uniform `"default"`:

- `seedAttackPathTargets` is a helper with no `ctx` of its own; it now derives a tenant-bound context, while its raw `pool.Exec` seeds stay on a plain context because they create the `tenants` row itself.
- `TestFindingRepositoryDerivesEngagementTenant` binds **per case**. Its whole point is that an engagement with no tenant projects into `"default"` — binding one fixed tenant for both cases would have deleted the behaviour under test.

**The fifth was the interesting one, and your code was right.** The test expected `ErrConflict` from a claim against another tenant's engagement and got `ErrNotFound`. `ErrNotFound` is the *better* answer: the insert is guarded by `WHERE EXISTS (SELECT 1 FROM engagements …)` under the other tenant's RLS scope, so the engagement is simply invisible. Reporting a conflict would confirm it exists in some other tenant — a disclosure a tenant boundary must not make. The assertion now requires not-found **and** requires that it is not a conflict, so the non-disclosure property cannot be relaxed by accident.

### And one test could only pass once

`TestAttackPathStoreReplaceBindingsSerializesProducer` failed on every run after the first:

```
ERROR: function "attack_path_pause_delete" already exists with same argument types
```

**Same root cause as the one fixed in #474**, which is why I am spelling it out: `defer pool.Close()` runs when the function *returns*, before any `t.Cleanup` callback. So the cleanup's `DROP FUNCTION` executed against a closed pool and — its errors discarded with `_, _ =` — failed in total silence, leaving the function behind. Closing via `t.Cleanup` makes LIFO run the drops first; the cleanup now reports failures; `CREATE OR REPLACE` keeps a stale object from wedging a future run. Three consecutive runs now pass and the function is gone afterwards.

This pattern is worth grepping for across the integration tests — a teardown that cannot report its own failure is how one bug becomes ten unrelated ones.

### What is genuinely well judged

**Inference never adds up to observation.** An edge is `Observed` only if at least one evidence item is `EdgeObserved`:

```go
observed = observed || e.Confidence == asset.EdgeObserved
```

Any number of inferred sources agreeing still leaves the edge inferred. That is the correct direction and the easy thing to get wrong, because "several weak signals" is exactly what tempts a scoring system into manufacturing confidence.

**Bounds are reported, not silent.** `BoundReport{MaxLength, MaxPaths, MaxDuration, Truncated, WallClockHit}` means a truncated path set *says* it is truncated. I mutation-tested it — silencing all four `Truncated = true` assignments fails `TestTraverseBoundsAndCancellation`. An attack-path view that quietly returns the first 100 paths would read as "these are the paths", which is the same lie as a scanner reporting zero findings because it timed out.

**Deterministic ranking with stable IDs, cycle-safe traversal, and provenance retained per edge** — so a path is explainable rather than asserted, which is what makes this usable as evidence rather than as a picture.

### Note

The checklist says `make build vet test typecheck` passes and that full Linux tests pass. That is true of the *unit* tests, but the Postgres suite these features depend on never ran. Worth treating a green local `go test ./...` as unverified for anything touching `persistence/postgres` until #478 lands — it is exactly what hid the five failures above.

Merging.

PR #483, opened 2026-08-09
ci: run Postgres integration tests fail closed

### LINKED ISSUE #478: ci: the Postgres integration suite never runs (33 test files skip silently)

VietSory:
## Summary

* run `Backend (Go)` against PostgreSQL 17 with `SYNAPSE_TEST_DB_DSN` configured
* fail immediately when the DSN is missing instead of silently skipping database integration tests
* fix `TestMigration0044` by seeding the historical v43 schema directly rather than using the current repository API
* correct `TestMigration0056BackfillsOnlyCanonicalRelativePaths` expectations and assert rollback by schema state without modifying the shipped migration
* make advisory and scan-job fixtures clean up before their pools close
* rerun the Postgres package with `-count=1` on the same database so persistent-state leaks cannot hide behind a fresh DB or Go's test cache

## Validation

Validated from current main (249dd80) on the exact rebased head:

* `go build ./...`
* `go vet ./...`
* `go test -count=1 ./...` against PostgreSQL 17
* `go test -count=1 ./internal/infrastructure/persistence/postgres` immediately afterward on the same database
* `CGO_ENABLED=0 go build ./cmd/...`
* `golangci-lint`
* Windows build / vet / test
* frontend test / typecheck / build

All four CI jobs are green on the validated tree.

## Negative proof

The fail-closed paths were deliberately broken and run in Actions:

1. Removed only the DSN configuration step: `Backend (Go)` failed at `Require Postgres integration database` before tests could silently skip.
2. Removed only the Postgres service while keeping the DSN configured: build and vet passed, then the full test step failed when integration tests could not connect.
3. Restored the old advisory cleanup ordering so its fixture leaked: the first full suite passed, then the explicit same-database Postgres rerun failed `TestAdvisoryRepository` on the leaked row.

The live database run also exercises the append-only evidence/audit and RLS/tenant-isolation failure paths rather than skipping them.

Closes #478.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -30,6 +30,24 @@ jobs:
   backend:
     name: Backend (Go)
     runs-on: ubuntu-latest
+    env:
+      POSTGRES_USER: postgres
+      POSTGRES_PASSWORD: ${{ github.run_id }}
+      POSTGRES_DB: synapse
+    services:
+      postgres:
+        image: postgres:17
+        env:
+          POSTGRES_USER: postgres
+          POSTGRES_PASSWORD: ${{ github.run_id }}
+          POSTGRES_DB: synapse
+        ports:
+          - 5432:5432
+        options: >-
+          --health-cmd "pg_isready -U postgres -d synapse"
+          --health-interval 10s
+          --health-timeout 5s
+          --health-retries 5
     steps:
       - uses: actions/checkout@v7
 
@@ -39,6 +57,16 @@ jobs:
           go-version-file: go.mod
           cache: true
 
+      - name: Configure Postgres integration database
+        run: echo "SYNAPSE_TEST_DB_DSN=postgres://${POSTGRES_USER}:${POSTGRES_PASSWORD}@localhost:5432/${POSTGRES_DB}?sslmode=disable" >> "$GITHUB_ENV"
+
+      - name: Require Postgres integration database
+        run: |
+          if [ -z "${SYNAPSE_TEST_DB_DSN:-}" ]; then
+            echo "SYNAPSE_TEST_DB_DSN is required in CI; refusing to silently skip Postgres integration tests"
+            exit 1
+          fi
+
       - name: Build
         run: go build ./...
 
@@ -46,7 +74,10 @@ jobs:
         run: go vet ./...
 
       - name: Test
-        run: go test ./...
+        run: go test -count=1 ./...
+
+      - name: Test Postgres suite again on the same database
+        run: go test -count=1 ./internal/infrastructure/persistence/postgres
 
       - name: Build (CGO disabled – distroless parity)
         run: CGO_ENABLED=0 go build ./cmd/...
--- a/internal/infrastructure/persistence/postgres/advisory_repo_test.go
+++ b/internal/infrastructure/persistence/postgres/advisory_repo_test.go
@@ -24,11 +24,16 @@ func TestAdvisoryRepository(t *testing.T) {
 	if err != nil {
 		t.Fatalf("connect: %v", err)
 	}
-	defer pool.Close()
+	// t.Cleanup is LIFO. Register Close first so fixture deletion runs while the pool is still usable.
+	t.Cleanup(pool.Close)
 
 	repo := NewAdvisoryRepository(pool)
 	id := "GHSA-" + randHex(t)
-	t.Cleanup(func() { _, _ = pool.Exec(ctx, "DELETE FROM advisories WHERE id=$1", id) })
+	t.Cleanup(func() {
+		if _, err := pool.Exec(context.Background(), "DELETE FROM advisories WHERE id=$1", id); err != nil {
+			t.Errorf("cleanup advisory %s: %v", id, err)
+		}
+	})
 
 	a := advisory.Advisory{
 		ID: id, Aliases: []string{"CVE-2024-9"}, Summary: "rce", CVSSScore: 9.8,
--- a/internal/infrastructure/persistence/postgres/migration_0044_test.go
+++ b/internal/infrastructure/persistence/postgres/migration_0044_test.go
@@ -7,11 +7,8 @@ import (
 	"time"
 
 	"github.com/google/uuid"
-	"github.com/jackc/pgx/v5/pgconn"
 	"github.com/pressly/goose/v3"
 
-	"github.com/KKloudTarus/synapse-ce/internal/domain/engagement"
-	"github.com/KKloudTarus/synapse-ce/internal/domain/shared"
 	"github.com/KKloudTarus/synapse-ce/migrations"
 )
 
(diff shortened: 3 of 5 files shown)


nghiadaulau (approved):
## Approve — this closes #478 completely, and the two "fixes" are strengthenings rather than the usual weakenings

I filed #478, so I checked this harder than usual: the easy way to make two failing migration tests pass is to relax what they assert. **That is not what happened here.** Verified against real Postgres 17, on a fresh container:

```
go build ./...   ok     golangci-lint  0 issues
go vet ./...     ok
postgres package, three consecutive runs on the SAME database:  ok / ok / ok
whole tree, run 1: green
```

**The Postgres suite is fully green for the first time, and green on re-runs.** Every other PR I reviewed today left at least two failures behind.

### `TestMigration0056`: the test was wrong, not the migration

The old table used one `want` for both tables. The new one splits `issueWant` from `hotspotWant`, and the reasoning is written down:

> `project_issues` has both file and location and therefore refuses a mismatch. Hotspots only had location before v56, so the same canonical location is sufficient evidence to backfill it.

So for the stale case (`file=src/stale.go`, `location=src/current.go:7`) the new expectation asserts **more**: hotspots must be backfilled to `src/current.go`, where the old test wrongly demanded `nil`. That distinction is exactly right and is the interesting part of the migration: **two sources that disagree are not evidence, one source that is canonical is** — refuse to guess for issues, backfill for hotspots.

The rollback assertion also got better rather than looser. It went from matching a pgx error type (`*pgconn.PgError`, code `42703`) to querying `information_schema.columns` — and now checks **both** tables instead of one. The contract is "Down removed the columns", so asserting the schema beats asserting which wrapper pgx happened to return that day. That brittleness is what made the original fail in the first place.

### `TestMigration0044`: the right root cause

Dropping `engagement.New` plus the repository API in favour of raw historical SQL is correct for a reason worth stating: **a migration test must not depend on today's domain code**, because today's repository writes today's columns — which do not exist at v43. Seeding the historical schema directly is the only version of this test that stays true as the schema moves.

### The teardown pattern, arrived at independently

`t.Cleanup(pool.Close)` instead of `defer`, cleanups that **report** their errors instead of `_, _ =`, and a cleanup that restores HEAD schema after a historical `DownTo` **even when an assertion fails midway**. That last one is the fix for a real cascade: a test that dies after `DownTo` leaves a shared database pinned to an old version, and every later test fails somewhere far from the cause. I hit exactly this three times today in other PRs. Good to see it solved at the source.

### The CI change

- Postgres 17 with a health check, DSN written to `$GITHUB_ENV`.
- **`POSTGRES_PASSWORD: ${{ github.run_id }}`** — no hardcoded credential, which also keeps the repo's own secret scanner happy. Worth noting for readers that `run_id` is a public value, not a secret; it is fine here precisely because the database is ephemeral and job-local.
- **The fail-closed step is the important one.** `SYNAPSE_TEST_DB_DSN` is set unconditionally by the previous step, so the guard cannot fire today — its value is that if someone later removes the service or the configure step, the job **fails loudly** instead of silently skipping 33 files. That is the requirement from #478, and it is the difference between coverage and the appearance of coverage.
- **Re-running the Postgres package on the same database** is a better idea than I asked for. It catches persistent-state leaks that a fresh database hides, which is the class of bug behind three of today's failures.

### Two things I could not verify, stated plainly

**Actions is currently disabled on this repository, so I cannot confirm the "all four CI jobs are green" or the negative-proof runs.** Those must predate the switch being turned off. My approval rests on the local verification above, not on those claims.

And a consequence worth naming without pushing: with Actions off, this change is **dormant** — the job it fixes will not run until Actions is re-enabled. Landing it now is still right (the two test fixes are valuable on their own, and `main` should be correct when the switch flips), but the coverage gap in #478 stays open in practice until then. That is your call to make, not mine.

### Unrelated observation

`TestRunHelperRedactsResponseSecretsAndDerivedCookies` failed once in my whole-tree run and passed 3/3 in isolation. That is the **third** distinct `dastengine` crawl test I have seen fail only under full-suite load today, alongside `SharesBoundedReauthentication` and `ReportsDeniedPathsAsSkipped`. Three variants of the same timing sensitivity is a pattern rather than bad luck, and it is now the only thing standing between this repo and a reliably green suite — worth its own issue.

Merging.

PR #484, opened 2026-08-10
fix(web): remove duplicate engagement navigation

nnatuan03:
## Summary
- remove the duplicate `Engagements` sidebar entry
- keep `New Engagement` as the single entry for the Engagement workspace
- keep the item active across `/engagements` list, create, and detail routes

Follow-up to #480.

## Validation
- `pnpm run typecheck`
- `pnpm test` — 291 tests passed
- `pnpm run build`

### DIFF
--- a/web/src/App.rules.test.tsx
+++ b/web/src/App.rules.test.tsx
@@ -149,7 +149,7 @@ describe('App Routing - Rules', () => {
     expect(document.documentElement.dataset.theme).toBe('light')
   })
 
-  it('places New Engagement before the Engagement list', () => {
+  it('shows only New Engagement for the Engagement workspace', () => {
     render(
       <MemoryRouter initialEntries={['/assets']}>
         <Sidebar />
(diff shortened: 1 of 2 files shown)


nghiadaulau (approved):
## Approve — with one correction to the framing, because the record should be accurate

**CI is disabled, so I ran the pipeline locally:**

```
pnpm test        36 files, 291 tests passed
pnpm typecheck   clean
pnpm build       built
```

### The `active` logic is genuinely better

Collapsing the `match: 'new-engagement' | 'engagements'` special-casing back to plain `prefix` removes a three-branch ternary and a `creatingEngagement` derivation from the render path. One rule — prefix match, else exact — is easier to keep true as routes are added. That part I would take on its own.

### But these were not duplicates, and the consequence is worth stating

The title says "remove duplicate engagement navigation". They pointed at **different destinations**: `Engagements` → `/engagements` (browse the list), `New Engagement` → `/engagements?create=1` (open the create form). The body's own wording — "keep New Engagement as the single entry" — is the accurate description.

I traced what remains, because "single entry" has a real cost here. `Engagements.tsx` initialises `creating` from the query param:

```tsx
const [creating, setCreating] = useState(() => searchParams.get('create') === '1')
```

and `closeCreate()` is the only thing that clears it. So **the only sidebar route into the engagement workspace now always opens a create form first**, and a reviewer who just wants to browse must dismiss it on every visit. The nav also no longer contains the word "Engagements" anywhere, which is what someone scanning for the list will look for.

It is mitigated — the dashboard offers two paths (`View queue` and `All Engagements` both to `/engagements`) — and an action-first sidebar is a legitimate design. You authored #480's navigation as well, so this is your call in your own area, and I am not going to overturn a deliberate UX decision on a small, reversible change.

What I do want on the record: the test now **asserts the list entry is absent**, so this is locked in rather than incidental. If browsing turns out to be the common case, that assertion is the thing to revisit, not a bug to hunt.

Merging.

PR #485, opened 2026-08-10
Dashboard: add security operations analytics

nnatuan03:
## Summary
- add a tenant-scoped `GET /api/v1/dashboard/security-operations` read endpoint with bounded `7d`, `30d`, and `90d` ranges
- add four responsive Dashboard visualizations using native SVG and existing design tokens: Asset posture, Findings over time, Active finding risk mix, and Assets by criticality
- preserve first-class `unknown`, exclude non-actionable historical advisories, deduplicate imported/canonical findings, and keep existing KPI/queue surfaces available when analytics fail
- document the endpoint in OpenAPI and add aggregation, RBAC, range, tenant-isolation, frontend loading/error, and interaction tests

## Issue
Tracks the Dashboard telemetry increment documented in #405. This PR does not close the platform Epic.

## Validation
- `go test ./...`
- `go vet ./...`
- `cd web && pnpm test` — 36 files, 293 tests
- `cd web && pnpm run build`
- `git diff --check`
- localhost browser QA at 1440×900 and 390×844
- verified light/dark modes, four chart surfaces, 7d/30d/90d switching, empty states, no document horizontal overflow, and no new console errors

### DIFF
--- a/api/openapi.yaml
+++ b/api/openapi.yaml
@@ -36,6 +36,8 @@ tags:
   description: Built-in code-quality and security rule catalog
 - name: assets
   description: Business-level Asset Management and security posture
+- name: dashboard
+  description: Tenant-scoped security operations summaries
 
 paths:
   /healthz:
@@ -959,6 +961,26 @@ paths:
         '500':
           $ref: '#/components/responses/InternalServerError'
 
+  /api/v1/dashboard/security-operations:
+    get:
+      tags: [dashboard]
+      operationId: getDashboardSecurityOperations
+      summary: Get tenant-scoped Asset and Finding analytics for the Dashboard
+      parameters:
+      - name: range
+        in: query
+        schema: {type: string, enum: [7d, 30d, 90d], default: 30d}
+      responses:
+        '200':
+          description: Security operations summary
+          content:
+            application/json:
+              schema: {$ref: '#/components/schemas/DashboardSecurityOperations'}
+        '400': {$ref: '#/components/responses/BadRequest'}
+        '401': {$ref: '#/components/responses/Unauthorized'}
+        '403': {$ref: '#/components/responses/Forbidden'}
+        '500': {$ref: '#/components/responses/InternalServerError'}
+
   /api/v1/appsec/assets:
     get:
       tags: [assets]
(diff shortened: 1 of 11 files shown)


nghiadaulau (approved):
## Approve — with two missing tests added as `04130a6`

**CI is disabled, so I ran the pipeline myself.** Against real Postgres 17:

```
go build ./...   ok      pnpm test       37 files, 297 tests
go vet ./...     ok      pnpm typecheck  clean
golangci-lint    0 issues  pnpm build    built
go test ./... (+Postgres)  fully green
```

### This is the honesty work done properly, and I checked it rather than trusting it

A dashboard is where a security product most easily lies by omission, so I went looking for the places a number could be quietly wrong. You had already closed them:

- **`findings_without_timestamp` is counted and reported**, not dropped. A trend chart that silently omits undated rows understates the total; this says *"N findings excluded from the trend because no creation timestamp is available."*
- **`external_findings_included` is a first-class field**, surfaced as *"External finding storage is unavailable; third-party findings are not included."*
- **`unknown` is a real category**, in both posture and severity, and an empty value is *mapped to* `unknown` rather than folded into `good` or `info`.
- **A read error on imported findings fails the request** (`writeError` + return) rather than returning a smaller count as fact.
- **Dedup handles the canonical/imported pair correctly**: `seen[item.FindingID] || seen[item.ID]` skips an import whose canonical twin is already counted, so promotion does not double-count.
- **Partial failure is scoped**: `analyticsError` renders `ErrorState` inside the analytics section only, leaving the KPI tiles and queues intact — same shape as #480.

The endpoint itself is clean: `rt.authz(userdom.PermView, …)` through the single chokepoint, tenant from `requestTenant` bound into ctx *and* passed explicitly to every service call, and `range` is a **closed allow-list** (`7d`/`30d`/`90d`, empty→30d, 400 otherwise) rather than a caller-supplied day count that could become an unbounded query.

The charts are native SVG with zero raw hex, and the a11y is better than most chart code I read: `role="img"` plus `<title>` on each chart, **and a per-segment `<title>` carrying label, value and percentage**, so the data is reachable without sight or a hover. The wide trend chart scrolls inside its own `overflow-x-auto` container rather than the page.

### What I fixed: one of three guards was real

I mutation-tested each disclosure instead of assuming a green suite covered them.

| mutation | caught? |
|---|---|
| delete the undated-findings notice | ✅ yes |
| delete the "third-party findings are not included" notice | ❌ no |
| flip `external_findings_included ?? false` to `?? true` | ❌ no |

The third is the one that matters most. That default decides what happens when a server response omits the field — an older deployment, a partial response — and `?? true` would present an **incomplete risk mix as the whole picture**, which is precisely what these flags exist to prevent. It was unguarded.

Added a Dashboard test for the notice and a separate `api.ts` test for the mapping defaults. The second had to live at the api layer: a component test that passes `externalFindingsIncluded: false` explicitly never exercises the `??` branch at all, which is why the existing test could not have caught it. Both verified by mutation — deleting the notice fails the first, flipping the default fails the second.

### One note, not blocking

The description lists localhost QA at two viewports, light/dark, and no horizontal overflow. I could not reproduce that part — I verified the code-level equivalents instead (tokens, `overflow-x-auto`, responsive grid classes) and the automated suite. Worth keeping that manual pass in the description; it is checking things the tests do not.

Merging.

PR #486, opened 2026-08-10
feat(ai-triage): annotate gate exemptions in SARIF

### LINKED ISSUE #430: feat(ai-triage): annotate gate exemptions in SARIF
Parent: #387
Found during the security review of #398.

## Goal

Make AI gate-exemption state visible in SARIF so code-scanning consumers can see that a finding was retained but excluded from the CI gate, together with the deterministic policy reason.

## Scope

- Carry AI gate-exemption metadata into the SARIF export path.
- Emit a standards-compliant SARIF suppression or equivalent result property.
- Include the versioned policy reason as the justification.
- Keep the result present; this must remain retain-and-mark, never delete or omit it.
- Cover CLI scan SARIF and the stored engagement export consistently.
- Add deterministic serialization and regression tests.

## Acceptance criteria

- A gate-exempt AI-triaged finding remains in SARIF results.
- Its SARIF entry identifies the exemption and policy reason.
- A suspected/review-required finding is not marked as gate-exempt.
- High/Critical, secret, protected-CWE, no-ledger, and ineligible findings never receive the SARIF exemption marker.
- Output remains valid SARIF 2.1.0 and is accepted by GitHub code scanning.

## Non-goals

- Does not grant gate authority.
- Does not change P0 policy.
- Does not hide or remove findings.

XUanhoa04:
## Summary

Expose policy-authorized AI false-positive gate exemptions in SARIF without removing the finding. This closes the visibility gap where CI could pass while the exported result did not explain that AI policy excluded it from the gate.

Closes #430.

## Changes

- Add a standards-based SARIF `external` / `accepted` suppression with a deterministic justification containing the AI policy version and reason.
- Keep exempted findings in `runs[].results` and add namespaced properties for consumers that do not render suppressions.
- Revalidate persisted decisions against the exact finding view being exported, so severity escalation, protected CWE, secret, ineligible, shadow, review-required, stale-policy, and no-evidence decisions remain unmarked.
- Apply the same projection to CLI SARIF and stored engagement SARIF exports.
- Document the output and add an Unreleased changelog entry.

## Safety

- The exporter never derives authority from advisory model output; it consumes a minimal server-owned projection after the full deterministic policy check.
- Findings are retained, never deleted.
- A missing legacy scan result produces an unannotated export; unreadable exemption metadata fails the export instead of silently misrepresenting gate state.
- Reports remain deterministic and no LLM runs in the export path.

## Validation

- `go test ./...`
- `go vet ./...`
- `go build ./...`
- `cd web && pnpm test` (36 files, 291 tests)
- `cd web && pnpm build`

## Checklist

- [x] `make build vet test typecheck` equivalent commands pass
- [x] `cd web && pnpm build` passes
- [x] Documentation updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ capabilities below are already shipped on `main`.
 
 ### Added
 
+- **Visible AI gate exemptions in SARIF.** Retained findings exempted by the verified AI false-positive policy now carry a standards-based external suppression plus deterministic policy version/reason metadata in both CLI and stored engagement SARIF exports; advisory, review-required, stale, and high-risk decisions remain unmarked.
 - **Asset-centric security management.** Promotes the shipped `fleet_business_services` rows in place to stable-keyed Business Assets above Engagements, with tenant/RLS-safe Project and technical-asset memberships, Engagement assignment, Findings/Coverage/Posture/History roll-ups, and UI built from the existing application components. Third-party findings retain provenance and no-promotion authority, while reachability retains proof tier and first-class `unknown`.
 
 - **Immutable Project Code workspace.** Capture bounded analysis-time source and Git comparison artifacts so historical source, unified/split diffs, and finding locations remain inspectable without reconstructing mutable workspaces; large source and diff views use bounded windows or virtualization.
--- a/cmd/synapse-api/main.go
+++ b/cmd/synapse-api/main.go
@@ -817,6 +817,7 @@ func main() {
 	}
 	aupService := aupuc.NewService(aupStore, auditLog, clock, cfg.AUPVersion)
 	exportService := exportuc.NewService(findingRepo, clock, buildinfo.App())
+	exportService.SetAIGateExemptions(scaService)
 	findingsService := findingsuc.NewService(findingRepo, commentRepo, retestRepo, auditLog, clock, ids)
 	// Both additions are independent; the tenant resolver is wired first because the review service
 	// takes findingsService as a dependency.
--- a/cmd/synapse-cli/main.go
+++ b/cmd/synapse-cli/main.go
@@ -1395,6 +1395,7 @@ func run(path string, failOn shared.Severity, mode, priority string, ignoreUnfix
 	// Report advisory opinions separately from the smaller policy-authorized gate-exempt set.
 	fpSuspect := res.SuspectedFPKeys()
 	fpGateExempt := res.AIGateExemptKeys()
+	fpGateExemptions := res.AIGateExemptions()
 	fpWouldExempt := res.AIWouldGateExemptKeys()
 	fpReview := res.AIReviewRequiredKeys()
 	if len(res.AITriage) > 0 {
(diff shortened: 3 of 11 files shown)


nghiadaulau (approved):
## Approve — with one missing guard added as `ca190e5`

**CI is disabled, so I ran the pipeline myself.** Against real Postgres 17: `go build`, `go vet`, `golangci-lint` (0 issues), and the **full suite green** including the Postgres package.

### The design decision here is the right one

Annotating rather than removing is what makes this safe. A SARIF export that *deleted* exempted findings would be indistinguishable from a scan that never found them — and a reader would have no way to tell an AI policy decision from an absence of risk. Keeping the result in `runs[].results` with an `external / accepted` suppression plus a justification naming the **policy version and reason** means a consumer can see the finding, see that it was excluded from the gate, and see on what authority. That is the difference between a suppression and a disappearance.

Keeping namespaced properties alongside the standard suppression is also right: tools that do not render suppressions still see the annotation rather than silently showing an unexplained result.

### Golden rule 5 holds, and I checked rather than assumed

No LLM anywhere in `internal/usecase/export` — the only mentions are comments stating the rule. The exporter consumes `AIGateExemptions()`, a server-owned projection produced **after** the deterministic policy check, and `AIGateExemption func(finding.Finding) (ports.AIGateExemption, bool)` is documented as returning false for advisory or review-required opinions. Authority is never derived from model output in the report path.

**Revalidation against the exported view is genuinely guarded.** Removing `humanReviewFloor(item) == "" && isFPTriageEligible(item)` fails three tests:

```
--- FAIL: TestGateExemptKeysRevalidatesOverlaidSeverity
--- FAIL: TestAIGateExemptKeysRevalidatesPersistedDecision
--- FAIL: TestServiceAIGateExemptionsReadsLatestResultDeterministically
```

That is the important one — a project quality profile can overlay severity, so a finding a tenant deliberately escalated to High must not stay exempt in an export. The comment in `aiGateExemptKeys` names that threat explicitly, and the test proves it.

### What was unguarded: `!c.Shadow`

Deleting the `!c.Shadow` term from `authorizedAIGateExemption` left the **entire suite green**.

That term matters most precisely at this seam. Shadow mode exists so AI triage can be observed without affecting a gate (P1.1, #451). A shadow decision reaching SARIF as `external / accepted` would tell a code-scanning UI the finding was **accepted**, while the policy deliberately had no authority to accept anything — the shadow guarantee leaking into an artifact customers and CI read as fact. With the term removed the export produced exactly that:

```
a shadow decision must never be exported as an accepted suppression:
  map[safe:{safe fp-gate-v3 verified_consensus}]
```

The code is correct; only the test was missing. Added one.

**Worth recording how I got it wrong first, because it is the trap this PR is otherwise careful about.** My initial test built the critique through `applyAIGatePolicy(..., aiTriageModeShadow)` — and it passed with the mutation applied, because that function clears `GateExempt` *before* the shadow branch, so the check failed on a different condition. `!c.Shadow` is defence against a **persisted or hand-built DTO**, so only a forged input (Shadow and GateExempt both true) exercises it. The test now forges it, and also asserts the same decision **does** project with `Shadow` cleared — so it cannot pass on some other unmet condition, which is how the first version fooled me.

### Also right

- **`hasVerifiedConsensus(c)` is re-derived**, not trusted from a boolean, so a stale or forged DTO cannot self-authorize.
- **`c.PolicyVersion == aiTriagePolicyVersion`** means a decision made under an older policy stops annotating rather than being honoured indefinitely.
- Unreadable exemption metadata **fails the export** rather than producing an unannotated one that misrepresents gate state. Failing an export is recoverable; a quietly wrong export is not.
- A missing legacy scan result yields an unannotated export — the conservative direction, since the alternative is asserting an exemption you cannot verify.

Merging.

PR #488, opened 2026-08-10
feat(appsec): add sanctioned source snapshot publishing

### LINKED ISSUE #402: feat(appsec): contribute Code Quality source snapshots from CI through a sanctioned publish path

VietSory:
## Summary

Adds a sanctioned, authenticated path for CI to contribute Code Quality source snapshots to an existing server-owned analysis.

* add `synapse-cli publish-source` to fetch the persisted analysis inventory and stream only retainable source files
* add `POST /api/v1/projects/{key}/analyses/{id}/source`
* store source under the server-owned artifact namespace and attach the resulting `SourceManifest` to the existing analysis
* expose the endpoint in the OpenAPI contract
* keep existing Code Quality reads on `ReadCodeFile`; no parallel source-serving path is introduced

## Security and integrity

The publish path is fail-closed across the transport, artifact, persistence, and read boundaries:

* derive tenant and actor from the authenticated request rather than client-supplied identity
* re-resolve the project and analysis inside that tenant before accepting source
* refuse unknown, cross-project, non-source, or already-captured analyses
* derive the upload allowlist from the persisted scanner-owned analysis inventory
* exclude credential-bearing and state/vendor paths such as `.env`, private keys, `.git`, `node_modules`, build output, and virtual environments
* require an absolute server-owned artifact root outside the scanned checkout
* create final artifact directories atomically and never overwrite an existing snapshot
* include writer actor, tool version, and publication timestamp in the manifest digest
* require at least one actually retained source file before persistence can mark source available
* attach the `SourceManifest` and durable audit record atomically in Postgres
* compensate known persistence failures while preserving artifacts when COMMIT outcome is uncertain
* bind `ReadCodeFile` results back to the persisted manifest before serving source
* refuse authenticated HTTP redirects so bearer credentials cannot be forwarded to a redirect target
* run configured source retention before every snapshot writer, including the sanctioned publish path
* reap expired manifest-less artifact claims left by an interrupted writer so stale claims cannot block publication indefinitely

Existing legacy head/base capture paths are only hardened to respect the same create-only and source-retention policy; this PR does not add a new comparison-source feature.

## API

`POST /api/v1/projects/{key}/analyses/{id}/source`

* requires the normal authenticated operator permission
* accepts `application/x-tar`
* requires `X-Synapse-Tool-Version`
* returns the persisted source manifest on creation
* documents conflict, authentication, validation, upload-limit, media-type, and server error responses in OpenAPI

## Scope guarantees

* no scan-path database read is introduced
* no scan upload toggle or implicit source publication is introduced
* package/image/non-source analyses remain ineligible
* source publication is an explicit CI contribution step against an already persisted analysis

## Validation

Validated after rebasing onto current `main` (`0a8ee424`).

The exact submitted tree passed:

* Go build and vet
* full `go test -count=1 ./...` with PostgreSQL 17
* same-database Postgres package rerun
* CGO-disabled command builds
* golangci-lint
* Windows build, vet, and tests
* frontend tests, typecheck, and build
* Agent Package Matrix

The validated staging merge and final clean commit have the same tree SHA:
`2c9479880cc88701a6f4eed9b3ebd14da62cd3d8`.

Additional negative proof deliberately removed load-bearing guards. Standard CI failed at the intended tests when:

* write-path retention was disabled
* the sanctioned OpenAPI route was removed
* stale manifest-less claim cleanup was disabled
* read-side manifest binding was removed
* durable source-publication audit was removed
* create-only artifact claiming was removed
* persistence was allowed to accept zero retained source bytes
* authenticated redirects were allowed
* uncertain COMMIT outcomes were compensated destructively
* legacy capture overwrite protection was removed
* secret/state path filtering was removed

Closes #402.

### DIFF
--- a/api/openapi.yaml
+++ b/api/openapi.yaml
@@ -333,6 +333,69 @@ paths:
         "404": {$ref: "#/components/responses/NotFound"}
         "500": {$ref: "#/components/responses/InternalServerError"}
 
+  /api/v1/projects/{key}/analyses/{id}/source:
+    parameters:
+    - $ref: '#/components/parameters/ProjectKey'
+    - $ref: '#/components/parameters/AnalysisId'
+    post:
+      tags: [projects]
+      summary: Publish an immutable source snapshot for one server-owned Project analysis
+      operationId: publishProjectAnalysisSource
+      description: >
+        Authenticated CI contribution path for Code Quality source. The server derives tenant
+        and actor from the bearer token, re-resolves the analysis under that tenant/project,
+        applies configured source-capture policy and limits, and creates the snapshot once.
+      parameters:
+      - name: X-Synapse-Tool-Version
+        in: header
+        required: true
+        schema: {type: string, minLength: 1}
+      requestBody:
+        required: true
+        content:
+          application/x-tar:
+            schema: {type: string, format: binary}
+      responses:
+        "201":
+          description: Source snapshot published and attached to the immutable analysis
+          content:
+            application/json:
+              schema:
+                type: object
+                required: [files, writer, digest]
+                properties:
+                  files:
+                    type: array
+                    items:
+                      type: object
+                      required: [path, bytes, lines, generated, available]
+                      properties:
+                        path: {type: string, minLength: 1}
+                        digest: {type: string}
+                        bytes: {type: integer, format: int64, minimum: 0}
+                        lines: {type: integer, minimum: 0}
+                        generated: {type: boolean}
+                        available: {type: boolean}
+                        reason: {type: string}
+                  truncated: {type: boolean}
+                  writer:
+                    type: object
+                    required: [actor, tool_version, published_at]
+                    properties:
+                      actor: {type: string, minLength: 1}
+                      tool_version: {type: string, minLength: 1}
+                      published_at: {type: string, format: date-time}
+                  digest: {type: string, pattern: '^sha256:[0-9a-f]{64}$'}
+        "400": {$ref: "#/components/responses/BadRequest"}
+        "401": {$ref: "#/components/responses/Unauthorized"}
+        "403": {$ref: "#/components/responses/Forbidden"}
+        "404": {$ref: "#/components/responses/NotFound"}
+        "409": {description: Analysis already has source or source artifact state conflicts}
+        "413": {description: Source archive exceeds the transport upload limit}
+        "415": {description: Content-Type is not application/x-tar}
+        "500": {$ref: "#/components/responses/InternalServerError"}
+        "503": {description: Source artifact storage is temporarily unavailable}
+
   /api/v1/projects/{key}/analyses/{id}/code/files:
     parameters:
     - $ref: '#/components/parameters/ProjectKey'
--- a/api/openapi_source_publish_test.go
+++ b/api/openapi_source_publish_test.go
@@ -1,0 +1,63 @@
+package api_test
+
+import (
+	"os"
+	"testing"
+
+	"gopkg.in/yaml.v3"
+)
+
+func TestSourcePublishOpenAPIContract(t *testing.T) {
+	b, err := os.ReadFile("openapi.yaml")
+	if err != nil {
+		t.Fatal(err)
+	}
+	var doc map[string]any
+	if err := yaml.Unmarshal(b, &doc); err != nil {
+		t.Fatal(err)
+	}
+	paths, ok := doc["paths"].(map[string]any)
+	if !ok {
+		t.Fatal("OpenAPI paths missing")
+	}
+	route, ok := paths["/api/v1/projects/{key}/analyses/{id}/source"].(map[string]any)
+	if !ok {
+		t.Fatal("sanctioned source publish path missing from OpenAPI")
+	}
+	post, ok := route["post"].(map[string]any)
+	if !ok {
+		t.Fatal("sanctioned source publish POST missing from OpenAPI")
+	}
+	if got := post["operationId"]; got != "publishProjectAnalysisSource" {
+		t.Fatalf("operationId=%v, want publishProjectAnalysisSource", got)
+	}
+	params, _ := post["parameters"].([]any)
+	foundToolVersion := false
+	for _, raw := range params {
+		p, _ := raw.(map[string]any)
+		if p["name"] == "X-Synapse-Tool-Version" && p["in"] == "header" && p["required"] == true {
+			foundToolVersion = true
+		}
+	}
+	if !foundToolVersion {
+		t.Fatal("required X-Synapse-Tool-Version header missing from OpenAPI")
+	}
+	requestBody, _ := post["requestBody"].(map[string]any)
+	content, _ := requestBody["content"].(map[string]any)
+	if _, ok := content["application/x-tar"]; !ok {
+		t.Fatal("application/x-tar request contract missing")
+	}
+	responses, _ := post["responses"].(map[string]any)
+	for _, status := range []string{"201", "400", "401", "403", "404", "409", "413", "415", "500", "503"} {
+		if _, ok := responses[status]; !ok {
+			t.Fatalf("response %s missing from source publish OpenAPI contract", status)
+		}
+	}
+	response201 := responses["201"].(map[string]any)
+	responseContent := response201["content"].(map[string]any)["application/json"].(map[string]any)
+	schema := responseContent["schema"].(map[string]any)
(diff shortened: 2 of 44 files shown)


nghiadaulau (approved):
## Approve — with one missing guard added as `26115bf`

**CI is disabled, so I ran the pipeline myself.** Against real Postgres 17: `go build`, `go vet`, `golangci-lint` (0 issues), full suite green apart from the known `dastengine` timing flake.

### First, a concern I checked and dropped

"Source snapshot publishing" reads like it contradicts a clean-room, no-source-copy principle, so I went to verify before saying so. It does not: `main` **already** carries `SourceManifest`, `ProjectSourceRetention` (90-day default) and a server-owned source artifact directory. Server-side source retention is an existing, deliberate capability, and this PR adds a *sanctioned, authenticated* route into it rather than inventing the practice. Worth recording that I looked, since the title invites the question.

### An authenticated tar upload is the highest-risk surface in this repo, and it is defended properly

I attacked it rather than reading it. Every classic vector is closed, and by **refusal rather than sanitisation** — which is the important distinction, because sanitising `../x` into `x` silently stores a file the client asked to put elsewhere:

```go
if hdr.Typeflag != tar.TypeReg  → reject   // symlinks, hardlinks, devices, FIFOs
canonical != hdr.Name           → reject   // traversal, absolute, non-canonical
allowed[canonical] missing      → skip     // server-derived inventory
duplicate canonical             → reject
```

Refusing everything except `TypeReg` is what closes the symlink-then-write-through-it attack completely, and it is stronger than trying to resolve link targets. Three of these are properly tested — `TestPublishArchiveRejectsUnsafeTransportAndRollsBackClaim` covers `../main.go`, a symlink to `/etc/passwd`, and a duplicate entry — and each case also asserts the **claim is rolled back**, so a rejected upload cannot leave a stale claim blocking future publishes. That last part is the detail most implementations miss.

### What was unguarded: the allowlist itself

Removing the `allowed[canonical]` term left the sharpest property of this endpoint unasserted — that **the server decides which files it accepts, from the persisted scanner inventory, and a client cannot introduce a file the scanner never saw.** With the term dropped:

```
a path absent from the scanner inventory was stored:
  [{Path:evil.go ... Available:true} {Path:main.go ...}]
```

`evil.go` is canonical, a regular file, and passes `sourcepolicy.RetainPath`. It has to be refused *purely* because the inventory does not list it — so no existing test could catch its acceptance. Added one; it fails on the mutation and passes on the real code.

This matters more than the traversal cases, and it is worth being explicit about why: traversal and symlinks are about *where* bytes land, and this is about *what* the server will hold at all. Deriving the allowlist from the scanner-owned inventory turns the upload from "trust the client's file list" into "confirm the client's bytes for files we already know about" — an allowlist rather than a denylist, which is the only version of this that stays safe as new file types appear.

### Also right

- **Create-only, never overwrite**, with the writer actor, tool version and publication timestamp **inside the manifest digest** — and a test proves tampering with the actor changes the digest. Provenance that is not covered by the digest is decoration.
- **Tenant and actor derived from the authenticated request**, with the project and analysis re-resolved inside that tenant before any bytes are accepted. Client-supplied identity never participates.
- **Refusing authenticated HTTP redirects** in the CLI, so a bearer credential cannot be forwarded to a redirect target. Easy to overlook, and the consequence is credential disclosure to an attacker-chosen host.
- **Requires an absolute, operator-owned artifact root outside the scanned checkout** — tested — so a snapshot can never be written into the tree being analysed.
- **At least one retained file required** before persistence marks source available, so an upload that policy filters down to nothing does not register as "source available". That is the coverage-honesty rule applied to an artifact.
- Reaping expired manifest-less claims means an interrupted writer degrades into a retry rather than a permanent lock.

Merging.

PR #489, opened 2026-08-10
feat(ai-triage): bound per-scan LLM work

XUanhoa04:
## Summary

Adds the first production guardrail slice for #392: a finite per-scan finding budget and configurable concurrency for AI false-positive triage.

Previously, every eligible finding was sent to the proposer and, when configured, the verifier. A noisy scan could therefore create up to 2N provider calls with no per-scan cap.

## Changes

- Add SYNAPSE_FP_TRIAGE_MAX_FINDINGS with a safe default of 100 and an accepted range of 1..1000.
- Add SYNAPSE_FP_TRIAGE_CONCURRENCY with a default of 6 and an accepted range of 1..32.
- Deterministically prioritize policy-actionable findings, then severity and risk, when the cap applies.
- Stop queued assessments from reaching the provider after context cancellation.
- Persist and evidence-seal eligible, attempted, and skipped finding counts in ai_triage_budget.
- Surface budget counts in CLI output, source warnings, API scan results, and debug trace events.
- Document the new controls and add regression tests for caps, ordering, cancellation, configuration bounds, concurrency, verifier call count, and evidence sealing.

## Safety properties

- Findings beyond the cap remain reported and gating.
- Zero, negative, malformed, and over-limit configuration values restore finite defaults; zero never means unlimited.
- A distinct verifier can make at most two calls per attempted finding.
- This does not relax consensus checks, human-review floors, or evidence requirements.

## Scope

This advances #392 with the finding-cap, concurrency, cancellation, and visible accounting slice. Token/cost budgets, bounded retries, circuit breaking, dashboards, and alerts remain follow-up work.

## Checklist

- [x] go build ./..., go vet ./..., and go test ./... pass
- [x] Dashboard unchanged; web build is not applicable to this diff
- [x] Preserves fail-safe gating and append-only evidence invariants
- [x] Documentation updated

Advances #392


nghiadaulau (approved):
## Approve — nothing for me to fix, which is rare in this queue

**CI is disabled, so I ran the pipeline myself.** Against real Postgres 17: `go build`, `go vet`, `golangci-lint` (0 issues), full suite green. The only failure was `TestRunHelperCrawlSharesBoundedReauthentication` — the known `dastengine` timing flake, unrelated to this change.

### The property that makes a budget safe, and it is genuinely enforced

A per-scan cap on LLM work is the kind of change that quietly becomes a security hole: cap the work, and the findings you did not get to look "fine". This does the opposite, and the test proves it rather than the comment claiming it:

```go
if got := result.AIGateExemptKeys(); len(got) != 0 {   // nothing exempted
if len(result.Findings) != 4 {                          // budget must retain every finding
```

**Untriaged findings stay in the result and stay gating.** Structurally, the cap can only ever *reduce* exemptions — a finding with no critique has nothing to exempt it — so the failure mode is "less AI help", never "less gating". That is the right direction for a guardrail, and the assertion on an empty exemption set is what makes it a guarantee instead of a consequence.

And the operator is told, in prose, in the output:

> AI false-positive triage budget attempted 2 of 4 eligible findings; **2 untriaged findings remain gating**

I mutation-tested that: suppressing the warning fails `TestRunFPTriageHardCapsAndSurfacesSkippedFindings` with the exact expected string. A silent cap would have read as "triage found nothing wrong with the other two".

### Zero never means unlimited — verified

```go
func boundedPositive(value, fallback, max int) int {
	if value < 1 || value > max { return fallback }
	return value
}
```

Zero, negative, and over-limit all collapse to the **finite default**. Changing `< 1` to `< 0` — the one-character version of "0 means unlimited" — fails `TestFPTriageBudgetDefaultsAndBounds`. This is the trap that catches most budget implementations, because `0` reads as "no limit" in so many other tools, and the config would then be a way to *remove* the guardrail while looking like a setting.

Bounding happens **twice** — in config and again in `runFPTriage` — which is right: the second one covers an in-process caller that sets the field directly rather than through env parsing.

### Failure modes all resolve toward gating

- **Cancellation**: remaining candidates get `Err: ctx.Err()` and `wg.Wait()` runs before returning. An errored critique never exempts, so an aborted scan gates everything it did not finish, and no goroutine is left to reach the provider after the caller gave up.
- **Panic**: a per-goroutine `recover()` turns a panic into that finding's `Err`, so it gates normally instead of taking down the scan. The comment calls this "making the best-effort guarantee unconditional", which is exactly what it does.

### Also worth noting

Sealing `ai_triage_budget` into evidence means the cap is part of the record, not just a log line — so a later reader of the evidence chain can tell whether a clean triage result covered every eligible finding or only the first hundred. That is the difference between an audit trail and a summary.

Deterministic prioritisation (policy-actionable, then severity, then risk) also means two runs of the same scan attempt the same findings, which keeps the report reproducible under a cap.

Merging.

PR #492, opened 2026-08-10
fix(ai-triage): fail closed on exemption evidence failure

### LINKED ISSUE #391: feat(ai-triage): fail-closed evidence for AI exemptions (P1.4)

## Goal

An AI gate exemption must not take effect unless the decision is sealed into evidence.

## Scope

- When evidence seal fails and the scan has an AI gate exemption:
  - revoke the AI exemption;
  - keep the finding gating;
  - return a structured warning/error;
  - emit a metric and an audit event.
- Advisory-only scans may keep best-effort behaviour.
- Tests for vault timeout, chain-head conflict, and storage failure.

## Non-goals

- Does not re-implement general evidence sealing. #386 (merged) already made SCA evidence sealing fail closed before persistence; this issue scopes the AI-exemption-specific revocation on top of that infrastructure.

## Acceptance criteria

- [ ] No AI exemption exists without an evidence link.
- [ ] An evidence failure never loses a finding.
- [ ] The operator gets a clear warning and can retry.

## References

- Builds on #386 (evidence sealing fail-closed). Golden rule: evidence is hash-chained; a seal mismatch blocks, never swallowed.

VietSory:
## Summary

Fixes #391 by making AI gate exemptions fail closed when scan evidence cannot be sealed.

An AI-generated exemption is now valid only when the corresponding scan decision is successfully committed to the evidence chain. If evidence sealing fails after an exemption has been authorized, the scan:

* revokes every active AI `GateExempt` decision;
* keeps the affected findings present and gating;
* marks the critiques as `ReviewRequired`;
* records `evidence_seal_failed` as the policy reason without changing the policy version;
* returns a typed `AIEvidenceSealError` while preserving the underlying evidence failure;
* emits an operator-visible structured failure metric;
* records an audit event for the revocation;
* fails before scan-result, finding, review-queue, or run persistence.

The existing #386 fail-closed evidence boundary remains authoritative. This change does not add an outer evidence retry: chain-head conflicts continue to be retried only by the evidence service, while generic storage errors remain fatal because their commit state may be ambiguous.

The failure metric is emitted independently of `ScanResult` persistence so it remains observable even though the failed result is discarded. It is logged at `ERROR` level so it is retained even when an operator configures error-only logging.

Compensating audit and metric reporting detach from request cancellation while remaining bounded, so an evidence failure caused by a canceled or timed-out request still has an opportunity to be observed and audited.

Both the normal source-scan pipeline and imported-SBOM pipeline are required by regression tests to use the same fail-closed AI evidence boundary.

## Tests

Coverage includes:

* successful sealing preserves and seals the authorized exemption;
* evidence vault/head timeout;
* generic evidence storage failure;
* exhausted chain-head conflict retries;
* canceled request context;
* audit failure while preserving the original evidence error;
* missing audit dependency;
* clock fallback for audit timestamp;
* forged/stale positive `GateExempt` flags;
* finding retention and `ReviewRequired` fallback;
* policy-version preservation;
* no outer evidence retry;
* no recovery/duplicate evidence append;
* metric emission outside the discarded scan result;
* metric visibility with an error-only logger;
* metric survival when audit recording fails;
* no duplicate metric emission after revocation;
* no AI metric for ordinary evidence failures;
* both SCA pipeline entry points using the fail-closed wrapper.

The branch is rebased on the current upstream `main` and contains one focused commit with changes limited to the SCA use case.

Closes #391

### DIFF
--- a/internal/usecase/sca/fptriage_budget.go
+++ b/internal/usecase/sca/fptriage_budget.go
@@ -12,12 +12,16 @@ import (
 
 // AITriageBudget records bounded per-scan AI coverage. The cap is on findings, so a configured
 // distinct verifier can make at most two LLM calls per attempted finding. Unattempted findings retain
-// their normal gate effect.
+// their normal gate effect. EvidenceSealFailures and EvidenceRevokedExemptions are in-memory failure
+// accounting for the revocation boundary; the scan fails before ScanResult persistence, so the
+// operator-visible metric is emitted separately as a stable structured event by sealEvidenceFailClosed.
 type AITriageBudget struct {
-	MaxFindings       int `json:"max_findings"`
-	EligibleFindings  int `json:"eligible_findings"`
-	AttemptedFindings int `json:"attempted_findings"`
-	SkippedFindings   int `json:"skipped_findings"`
+	MaxFindings               int `json:"max_findings"`
+	EligibleFindings          int `json:"eligible_findings"`
+	AttemptedFindings         int `json:"attempted_findings"`
+	SkippedFindings           int `json:"skipped_findings"`
+	EvidenceSealFailures      int `json:"evidence_seal_failures,omitempty"`
+	EvidenceRevokedExemptions int `json:"evidence_revoked_exemptions,omitempty"`
 }
 
 func (s *Service) runFPTriage(ctx context.Context, result *ScanResult, workspaceDir string, trace *scanDebugTrace) {
--- a/internal/usecase/sca/fptriage_evidence.go
+++ b/internal/usecase/sca/fptriage_evidence.go
@@ -1,0 +1,173 @@
+package sca
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"log/slog"
+	"strconv"
+	"strings"
+	"time"
+
+	"github.com/KKloudTarus/synapse-ce/internal/domain/shared"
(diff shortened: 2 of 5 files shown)


nghiadaulau (approved):
## Approved — strict review passed

Fixes #391. The fail-closed semantics are correct: on an evidence-seal failure with an active AI exemption, `sealEvidenceFailClosed` revokes every `GateExempt` bit, marks the critiques `ReviewRequired`, records `evidence_seal_failed`, emits an operator-visible metric, audits the revocation on a detached bounded context (`context.WithoutCancel` + 5s), and returns before any run/finding/review/result persistence — so no exempted finding can ever be committed. When no AI exemption is active it returns the raw seal error unchanged, preserving the #386 boundary. No outer evidence retry is added (correctly leaving chain-head retry to the evidence service).

**Verified locally against current main (CI disabled)** via trial merge: `go build ./...`, `go vet`, `gofmt -l` (clean), and `go test ./internal/usecase/sca/` all green. The 638 lines of tests cover every path: successful seal preserves the exemption; vault/head timeout; storage failure without outer retry; chain-head-conflict exhaustion; canceled-context audit survival; audit-failure error joining; forged-flag clearing; and both the source-scan and imported-SBOM pipelines routed through the boundary.

Merging.

PR #502, opened 2026-08-11
feat(ai-triage): complete provider-independent verifier (P1.3)

H1eu232:
## Summary

Completes the provider-independence scope that remained after #479 advanced #390:

- gives the false-positive verifier its own endpoint, credential, provider identity, and model;
- records proposer/verifier provider and canonical model-family metadata end to end;
- enforces fail-closed `model_family` or stronger `provider` separation of duties;
- keeps findings gating when identity metadata is missing, stale, aliased, or inconsistent.

This preserves the blind-verifier ordering and Bedrock alias hardening already merged in #479, while integrating with the P1.2 durable human-review workflow and the newer bounded AI-triage request budget on `main`.

## Changes

- Add explicit proposer/verifier provider identity and independent verifier transport configuration.
- Re-derive and validate provider/model-family identity at the gate-authority boundary instead of trusting persisted booleans or metadata.
- Bump the replayable gate contract to `fp-gate-v4`.
- Persist independence metadata through sealed scan evidence, human-review records, audit events, PostgreSQL, API/OpenAPI, evaluation reports, and the review UI.
- Add migration `0078_ai_triage_independence.sql`; `0074` through `0077` are already occupied on current `main`.
- Keep the request-budget cap/concurrency behavior from #489 and apply the P1.3 independence policy inside the bounded triage runner.
- Pin the complete verifier system prompt fingerprint so a future static anchoring edit requires explicit security review.
- Make provider-mode defaults fail closed: the verifier inherits the proposer provider until an operator deliberately chooses a distinct provider.

## Safety properties

- A single model or same canonical model family cannot self-confirm.
- Provider mode requires both a different provider and a different model family.
- Missing/unknown identity or policy remains advisory-only.
- Verifier credentials are never logged.
- Findings remain reported; only the deterministic policy may exempt an eligible finding from a gate.
- High/critical, secret, and protected-CWE human-review floors are unchanged.

## Validation

Passed locally on Windows against `main@34b9b21`:

- `go build ./...`
- `go vet ./...`
- focused P1.2/P1.3/config/SCA/export/API/CLI tests
- `pnpm run typecheck`
- `pnpm run test` — 37 files / 297 tests
- `pnpm run build`

On current `main@34b9b21`, whole-tree `golangci-lint v2.12.2` reaches three newly merged findings outside this PR: G706 in `cmd/synapse-agent/detect.go`, QF1006 in `internal/usecase/fleet/detect/engine_test.go`, and an unused helper in `internal/infrastructure/persistence/postgres/telemetry_repo.go`. None of those files is changed by this PR.

`go test ./...` reaches an unrelated pre-existing Windows-only failure in seven newly merged `internal/domain/offensivepolicy` fixture tests: their literal LF anchors do not match the CRLF working-tree copy of `policy.yaml`. Neither that package nor its fixture is changed here. All AI-triage and P1.2 packages pass with `-count=1`.

The PostgreSQL migration integration test is included and runs when `SYNAPSE_TEST_DB_DSN` is available; the local database required an unavailable password, so CI with its fail-closed Postgres job is authoritative for the actual migration execution.

## Tracking

- Parent epic: #387
- Completes the remaining scope of #390
- Depends on the already merged deterministic temperature fix #397 / #384


nghiadaulau (approved):
## Approved — strict review passed (2 independent reviews + local verification)

Completes #390's provider-independence scope. **The separation-of-duties core is sound and fails closed.** Identity is re-derived at every gate-authority boundary (`hasIndependentVerifierIdentity` recomputes canonical model-family + provider and requires the persisted fields to *equal* the recomputed values, then calls `agent.IndependentLLMs`), never trusted from persisted booleans. `hasVerifiedConsensus` re-checks both verdicts are refuted with confidence ≥75 (hardcoded), and the export seam (`authorizedAIGateExemption`) re-validates again against the exact finding. A single model / same model-family can never self-confirm; provider mode additionally requires distinct providers; unknown/missing/aliased/inconsistent identity fails closed to human review. Verifier defaults inherit the proposer transport (fail-closed until an operator picks a distinct provider); the `fp-gate-v4` bump invalidates stale v3 exemptions. Verifier credentials are never logged; the LLM only proposes.

**Reviews:** security audit found *no critical/high/medium* SoD bypass (all six focus areas PASS); go-arch found *no dependency-rule violation* (domain stays pure, config at composition roots, narrow consumer-defined ports).

**Verified locally against current main (CI disabled):** `go build ./...`, `go vet`, `gofmt -l` (own files clean), Go tests for all touched packages, `pnpm run typecheck`, and the real-Postgres migration-0078 integration test (additive columns, tenant-scoped unique constraint, fail-closed CHECK for fp-gate-v4) — all green.

**Non-blocking follow-ups to address later** (neither affects the current gate authority, which re-derives identity unconditionally):
1. `aitriagereview.NewPending` gates its independence-metadata invariant on the literal `"fp-gate-v4"`; a future `fp-gate-v5` bump would make that *review-record* invariant fail open. Drive it from data (e.g. an explicit `RequireIndependenceMetadata` flag) rather than the version string.
2. `agent.IndependentLLMs(...)` takes five same-typed string args in a transposition-prone order — consider a small value struct.

Merging.

Issue #424, opened 2026-08-07
feat(blueteam): columnar telemetry tier with retention, sampling and cost model


nghiadaulau:
Parent: #405. Phase 5. Depends on the detection engine. This is the milestone that unlocks retro hunting.

## Objective

Adopt a columnar or time series store for raw telemetry and detection history, with an explicit retention, sampling and cost model, so retro hunting over raw events becomes possible without overwhelming the system of record.

## Why this is its own issue with a spike

Postgres, the file store and the in memory store are the only persistence implementations today. Blue team telemetry is a columnar problem measured in millions of rows per host per day. This is the single largest cost fork in the whole platform, so it is gated behind a written comparison rather than an assumption. Milestone one shipped detections only precisely so this decision could be made with real data volume in hand.

## Requirements

1. A scoped spike first, with a written comparison merged before the implementation: candidate stores, ingest throughput, query latency for the retro hunt patterns that matter, operational burden, and cost at a stated fleet size. ClickHouse is the reference candidate. The comparison names the decision and the reason.
2. Telemetry is a distinct persistence port. It does not pollute the existing `ports` used for the system of record, and the columnar store is never in the path of a finding, a judgment or the evidence chain.
3. Ingest is bounded and backpressured. The agent side buffer, the transport limits and the store ingest rate are one coherent budget, and overflow at any stage is reported as a telemetry gap on the affected host, never silently dropped.
4. Retention is tiered and explicit: a hot window for interactive hunting, a warm window at reduced resolution, and expiry. Every tier boundary and the expiry are configurable and documented, and expiry is audited.
5. Sampling is explicit where it exists. If high volume classes are sampled to control cost, the sampling rate is recorded with the data so a hunt knows it is looking at a sample, and a sampled window is never presented as complete.
6. Integrity at the batch level. Telemetry is not hash chained per event, per the #405 decision, but agent batches remain signed and sequence numbered so a hunt can tell whether it is looking at a complete sequence or a lossy one.
7. Retro hunt queries are the acceptance surface: run a detection rule retroactively over the hot window, reconstruct the context around an existing detection, and pivot from an asset to its raw events. These three patterns define whether the store choice was correct.
8. Cost is observable. The store's footprint per fleet size is measurable and reported, so an operator can predict spend rather than discover it.

## Out of scope

- Response actions.
- Cross region telemetry federation.

## Specification sketch

```go
// internal/usecase/ports
type TelemetryStore interface {
[...]

Issue #423, opened 2026-08-07
feat(blueteam): detections as hash-chained, attributable evidence


nghiadaulau:
Parent: #405. Phase 5. Depends on the detection engine.

## Objective

Make a detection a hash chained, attributable record in the same evidence spine as findings and judgments, rather than an alert row, so a detection is defensible in an audit and can join the correlation graph.

## Why this is the differentiator

Every incumbent in this category produces alerts. The claim this project can make that they cannot is that a detection is evidence: attributable to an agent identity, bound into the hash chain with the previous hash, and joined to the same asset and finding the static pillars reason about. This issue is what turns the detection engine's output into that claim.

## Requirements

1. A detection is written into the evidence chain with the same binding the chain already uses: previous hash, attribution, metadata and content. The attribution is the agent identity, following the fleet identity model.
2. Detections are chained. Raw telemetry is not, per the #405 storage decision. This boundary is enforced in code and documented, so no one later attempts per event chaining and collapses the throughput.
3. Agent batches are signed and sequence numbered. A detection references its batch by sequence, so a gap in the sequence is detectable and is reported as a potential loss rather than silently accepted.
4. A broken detection chain is treated exactly as a broken evidence chain elsewhere: it blocks the report that depends on it, and it is surfaced, not worked around.
5. Detections join the asset model. A detection references the asset it was observed on, so it can appear on that asset's risk story and in the attack path graph.
6. Deduplication and correlation of repeated detections into an incident level record, without losing the individual attributable detections underneath. The rollup is a view, the detections remain the ledger.
7. Tenant scoped throughout, through the existing chokepoint.
8. Retention policy for detections is explicit and configurable, and expiry is itself an audited action, because deleting evidence silently is exactly what this project exists to prevent.

## Out of scope

- The columnar telemetry tier.
- Response actions.

## Specification sketch

Reuse the existing evidence chain writer. A detection becomes a chain record with `kind = detection`. Batch integrity:

```go
type AgentBatch struct {
    AgentID   shared.ID
    Sequence  uint64      // monotonic per agent
    Signature string      // over the batch content and sequence
    Detections []shared.ID
}
```

A missing sequence number for an agent is a `batch_gap` coverage event on that agent.

## Acceptance criteria

- [ ] A detection is written to the evidence chain with full binding, verified by chain verification.
- [ ] Tampering with a detection record breaks chain verification and blocks the dependent report, verified by test.
[...]

PR #511, opened 2026-08-11
feat(cspm): live cloud posture connectors for AWS, Azure and GCP

### LINKED ISSUE #429: feat(cspm): live cloud posture connectors for AWS, Azure and GCP
Parent: #405. Phase 7, independent track. Depends only on #431 fleet asset model. Fills the verified CSPM gap.

## Objective

Add live cloud posture connectors for AWS, Azure and GCP so the platform assesses deployed cloud state, not only infrastructure as code files, closing the gap between what was written and what is running.

## Why this is a named gap

`internal/infrastructure/tools/misconfig` scans Terraform, CloudFormation, ARM, Kubernetes, Dockerfile and Compose as files. There is no live cloud connector, so there is no account level posture, no identity and access analysis against deployed state, and no drift detection. Wiz, Orca, Prisma Cloud and Sysdig all compete on exactly this surface, and a platform that claims container and VM estate coverage is incomplete without it. The #431 asset model already declares cloud account, project and resource categories, so the inventory target exists.

## Requirements

1. Read only connectors for AWS, Azure and GCP that enumerate the resource categories the asset model declares: accounts and projects, compute, storage, network, and identity and access.
2. Credentials come from the vault by reference, never a value in a work order, a log, a model context or an artifact. Prefer short lived credentials and a role assumption model over long lived keys, and document the least privilege role each connector requires.
3. Posture checks over live state, clean room and catalogued exactly like the other rule surfaces, with the same drift test. Start with the high confidence categories: public exposure of storage and compute, identity misconfiguration such as wildcard permissions and unused high privilege principals, missing encryption, and network paths that expose sensitive resources.
4. Drift detection where an infrastructure as code source is also available: compare the file based expectation from the existing misconfig scanners with the live state, and report drift as its own finding class. This is the join that neither a pure file scanner nor a pure cloud scanner can make alone, and it is this project's specific advantage because it already owns the file side.
5. Findings enter the asset model and the same governance, prioritisation and lifecycle as every other finding. A cloud posture finding is not a separate silo.
6. Bounded and rate limited against the cloud provider APIs, with pagination handled and partial enumeration reported as a coverage gap rather than as a clean account.
7. Multi account and multi subscription from the start. An organisation has many accounts, and the connector enumerates across them where the credential permits, reporting any account it could not reach as a coverage gap.
8. Read only and no mutation. A posture connector observes. Any remediation is a separate governed response action, not a side effect of scanning.
9. All provider API access is argv only where a pinned binary is used, or a vendored SDK confined to an infrastructure package following the dependency rule. No provider SDK leaks into a domain type.

## Out of scope

- Remediation of cloud findings, which is a governed response action.
- Cloud workload runtime protection, which is the agent and detection pillars.
- Providers beyond the three named, which are follow ups.

## Specification sketch

Asset mapping into the #431 model:

| Cloud resource | Asset kind |
|---|---|
| Account or subscription or project | `cloud_account` |
| Compute instance | `host` or `workload` |
| Storage bucket or account | `storage` |
| Network construct | `exposure` |
| Identity principal, role, policy | `identity` |

```go
// internal/usecase/ports
type CloudConnector interface {
    Enumerate(ctx context.Context, scope CloudScope) (Inventory, []CoverageIssue, error)
    Evaluate(ctx context.Context, inv Inventory) ([]Finding, error)
}
```

## Acceptance criteria

- [ ] A read only connector per provider enumerates the declared resource categories against a recorded or emulated API, with fixtures committed.
- [ ] Credentials are used by reference and never appear in any log, artifact or model context, verified by a grep test over a full run.
- [ ] The least privilege role per connector is documented, and a missing permission produces a coverage gap rather than fewer resources silently.
- [ ] Posture checks match hand verified expectations on the fixtures, one per high confidence category.
- [ ] Drift detection reports a divergence between an infrastructure as code source and live state as its own finding class, verified on a fixture where the two differ.
- [ ] Partial enumeration and an unreachable account each report a coverage gap, never a clean account.
- [ ] Cloud findings enter the asset model and the standard governance and prioritisation, with no separate silo.
- [ ] No provider SDK type appears in a domain type, verified by an architecture test.

## Test requirements

1. Fixture based enumeration and evaluation tests per provider using recorded or emulated API responses.
2. A credential leakage grep test over a full run.
3. A drift test on a fixture where file and live state differ.
4. Coverage gap tests for partial enumeration and an unreachable account.
5. An architecture test asserting no provider SDK in a domain type.

### Packages
- Port: `internal/usecase/ports/cloudconnector.go` with `CloudConnector` (the interface in this issue). No provider SDK type appears in a domain type (an `arch_test.go` asserts it).
- Connectors confined to infra: `internal/infrastructure/cloud/aws`, `.../azure`, `.../gcp`, each vendoring only its SDK within that package.
- Usecase: `internal/usecase/cspm/service.go` enumerates then evaluates; posture checks catalogued in `internal/domain/cloudposture/catalog.go` with a drift test, exactly like the other rule surfaces.
- Assets map into the #431 model per the table in this issue (`cloud_account`, `host`/`workload`, `storage`, `exposure`, `identity`); findings use the existing `Finding` path and standard governance/prioritisation, no separate silo.

### Credentials and least privilege
Credentials come from the vault by reference (`ports.Credentials`), never a value in a work order, log, model context or artifact (a grep test over a full run). Prefer short-lived role assumption over long-lived keys; the least-privilege role per connector is documented in `docs/`. A missing permission yields a coverage gap, not fewer resources silently.

### Drift detection (the differentiator)
Where an IaC source is also available, compare the file-based expectation from `internal/infrastructure/tools/misconfig` against live state and report divergence as its own finding class. This is the join a pure file scanner or a pure cloud scanner cannot make; the project already owns the file side.

### Boundedness
Bounded and rate-limited against provider APIs with pagination handled. Partial enumeration and an unreachable account each report a coverage gap, never a clean account. Multi-account/multi-subscription from the start. Read-only; remediation is a separate governed response action (#425).

### Config
| Key | Default |
|---|---|
| `SYNAPSE_CSPM_ENABLED` | `false` |
| `SYNAPSE_CSPM_PROVIDERS` | empty |
| `SYNAPSE_CSPM_RATE` | provider-appropriate default |

pho-veteran:
## Summary

- add read-only AWS, Azure, and GCP posture connectors behind a sandboxed worker helper
- add durable asynchronous CSPM runs, tenant-safe persistence, per-request authorization, evidence sealing, and active observation reconciliation
- add IaC/live-state drift findings, provider coverage gaps, API/OpenAPI/configuration, and operator documentation

## Safety and governance

- resolves credentials by vault reference in the worker and passes secret material through an inherited FD
- authorizes every provider operation against current engagement scope and authorization state
- fails closed when sandboxing, egress enforcement, helper pinning, evidence sealing, or run leasing is unavailable
- keeps provider SDKs outside domain/use-case and API/worker processes
- preserves finding triage while retiring only complete, comparable CSPM observations

## Verification

- `make build vet test`
- `git diff --check`
- five-agent correctness, architecture, and security QA completed with no merge blockers

Closes #429


nghiadaulau (changes requested):
## Requesting changes — one deploy-blocking defect; the rest is strong

Reviewed strictly and verified locally against current `main` (CI is disabled) via trial merge: `go build ./...`, `go vet`, `gofmt -l`, and `go mod tidy` are all clean, and every new package's tests pass (`cloudposture`, `cspm`, `cloud/{aws,azure,gcp}`, `cloudsandbox`). Two independent reviews (security + architecture) were also run. The security architecture is genuinely strong — but there is one blocker that prevents the server from starting, so this cannot merge yet.

### 🔴 BLOCKER — duplicate migration version numbers (server fails to boot)
`migrations/0074_cspm_runs.sql` collides with the shipped `0074_detections.sql` (#423), and `migrations/0075_cspm_observations.sql` collides with `0075_telemetry_events.sql` (#424). `main` already carries migrations `0074`–`0078`. The git merge is "clean" only because the filenames differ, but goose keys on the leading version integer, so startup migration panics:

```
panic: goose: duplicate version 74 detected: 0074_cspm_runs.sql
```

I reproduced this against real Postgres — `postgres.Migrate(ctx, dsn)` now panics before applying anything, so `synapse-api` / `synapse-worker` would fail to boot (and the CSPM tables + their `synapse_enable_tenant_rls` calls would never apply). This is exactly the CLAUDE.md rule "append a **new numbered** file; never edit a shipped migration."

**Fix (mechanical):** renumber to the next free versions — `0074_cspm_runs.sql` → `0079_cspm_runs.sql`, `0075_cspm_observations.sql` → `0080_cspm_observations.sql`. Keep the `CALL synapse_enable_tenant_rls(...)` and FK/PK definitions intact. No Go code hardcodes these version numbers, so renaming the two files is the whole fix. Please also add a real Postgres migration integration test (like `migration_0078_test.go`) that calls `Migrate` — it would have caught this.

### 🟡 Low / non-blocking (worth folding in)
1. **`internal/infrastructure/cloud/aws/aws.go:89`** — the `err` from `cloudposture.NormalizeRoot(...)` is assigned but never checked (it's overwritten unread at the next `vault.Resolve`). Azure checks-and-returns the equivalent call; AWS is the outlier. This trips the repo's own `ineffassign`/`make lint` gate and silently leaves `scopeKey` empty on failure. Check the error and return it.
2. **Per-operation cloud authorization is cooperative** (the sandboxed helper *asks* the parent to authorize each request over the FD channel). The parent does re-check live scope + auth-window + RoE server-side before the credential is ever resolved, and the helper is SHA-256-pinned with default-deny egress + read-only credentials — so the trust boundary is sound. But consider (a) recording this boundary in ADR-0004, and (b) a parent-side allowlist of permitted operation *names* (you already receive `CloudOperation.Name`/`Category`) so an unexpected/mutating op is denied rather than authorized-by-default rather than relying only on the connectors issuing read-only calls.
3. **GCP OAuth token refresh** (`cloud/gcp/connector.go`) is not routed through the per-op authorizer (AWS uses static creds, Azure routes its token through the authorizing transport). It's still egress-bounded, so low impact — just document the required Google token host in the egress-allowlist docs.
4. **`internal/infrastructure/cloudsandbox/executor.go:134` (and `:102`,`:138`)** — `fmt.Errorf("constant string")` with no `%w`/directive drops the underlying `runErr`/exit-code/timeout context across the sandbox boundary. Use `errors.New`, and wrap `runErr` with `%w` on the helper-failed path for diagnosability.

### ✅ What I verified is solid (so you know the bulk is good)
- **Read-only** across all three providers — only Describe/List/Get/AggregatedList/GetIamPolicy/ResourceGraph-query + STS AssumeRole; no mutating verb anywhere in `internal/infrastructure/cloud`.
- **SDK confinement** — no `aws-sdk-go-v2`/`Azure`/`google.golang.org/api`/`cloud.google.com` import in `domain`/`usecase`/`api`/`worker`; confined to `infrastructure/cloud/*` + `cmd/synapse-cspm`, enforced by the `cloudposture/arch_test.go` tripwire. Dependency rule: PASS.
- **Secrets (Rule 3)** — credential resolved server-side by vault ref, streamed out-of-band over an inherited pipe FD (env carries only FD numbers), `clear()`-ed on both sides, stdout redacted against the secret with a placeholder tripwire that fails the run if credential material appears. No secret in argv/env/logs/evidence.
- **argv-only exec (Rule 1)** — helper launched via `ports.ToolRunner` argv → bwrap; no shell/`fmt.Sprintf` command building.
- **Server-side authz + fail-closed (Rule 2/4)** — `prevalidate` enforces tenant/engagement/AllowsExecution/IsAuthorizedAt + scope-key ∈ in-scope accounts before the credential is resolved; per-op re-check; fails closed if authorizer/egress-hosts/evidence-sealer/executor/run-lock is missing; worker exits if kernel egress isn't live or the helper isn't SHA-pinned.
- **Evidence (Rule 5)** — observations/drift sealed into the hash-chained evidence; no LLM in the CSPM path.
- **Tenant isolation** — both new tables are `tenant_id NOT NULL`, composite PK led by `tenant_id`, FK to `engagements(tenant_id,id)`, RLS via `synapse_enable_tenant_rls`; runs re-assert ctx tenant; cross-engagement `GetRun` → 404. Routes gated `PermOperate`/`PermView` + `withEngTenant`; injected fields are `json:"-"`.

Please renumber the migrations (blocker) and ideally fold in #1 and #4; then this is good to merge. Nice work on the sandbox/credential-broker design.

pho-veteran:
Thank you for the detailed review and for reproducing the migration failure against PostgreSQL. I have addressed the blocker and the related review findings in `bd9e701`.

### Addressed

- Renumbered the CSPM migrations to the next available versions:
  - `0074_cspm_runs.sql` → `0079_cspm_runs.sql`
  - `0075_cspm_observations.sql` → `0080_cspm_observations.sql`
- Preserved the existing table definitions, tenant-scoped foreign keys, primary keys, and `synapse_enable_tenant_rls(...)` calls unchanged.
- Added a PostgreSQL migration integration test that calls `Migrate` and verifies both `cspm_runs` and `cspm_observations` exist. It follows the repository's `SYNAPSE_TEST_DB_DSN`-gated migration-test pattern.
- Fixed the ignored AWS `NormalizeRoot` error and now reconstruct the canonical AWS scope key before creating the authorization-aware SDK transport.
- Improved sandbox error handling:
  - runner failures retain their underlying error through `%w`;
  - nonzero exit code, timeout, and truncation state are reported;
  - credential-material detection uses `errors.New`;
  - authorization denials and scope mismatches retain `shared.ErrForbidden`, allowing the worker to classify them as permanent failures rather than retryable helper failures.

### Authorization hardening

I also folded in the suggested parent-side operation allowlist and documented the cooperative helper boundary in ADR-0004.

- AWS permits only the connector's explicit read-only SDK operation names.
- Azure permits only the exact Resource Graph query route and tenant OAuth token endpoint.
- GCP permits only the exact service hosts and read routes used by project/folder, Compute, Storage, IAM, and `getIamPolicy` enumeration.
- Unexpected operations, lookalike routes, mutating requests, provider mismatches, and scope-key mismatches fail closed.
- GCP OAuth token refresh now passes through the same parent authorization callback and remains constrained by the sandbox egress allowlist.
- Removed the generic `inventory/enumerate` authorization exception so the helper cannot bypass provider-specific policy by relabeling an operation.

The GCP OAuth egress requirement is now explicit in the cloud-posture guide, and the helper trust boundary is recorded in `docs/adr/0004-cspm-helper-authorization.md`.

### Verification

The updated branch passes:

```text
make build vet test typecheck
git diff --check
```

Focused AWS, Azure, GCP, CSPM, sandbox, and PostgreSQL persistence tests also pass. Regression coverage now includes migration application, operation allowlisting, canonical AWS scope propagation, OAuth operation identity, authorization denial propagation, scope mismatch classification, and sandbox failure context.

nghiadaulau (approved):
## Approved — all review findings addressed, merging

Re-reviewed the update (`bd9e701`). Every point from the previous round is resolved:

- **🔴 Blocker fixed:** migrations renumbered to `0079_cspm_runs.sql` / `0080_cspm_observations.sql` (no more duplicate-version collision), and you added the Postgres migration integration test I asked for. Verified on a clean v78 baseline: `Migrate` applies both cleanly and the CSPM schema tests pass. (`go build ./...`, `go vet`, `gofmt -l`, and the cspm/cloud/cloudsandbox/cloudposture suites all green.)
- **Finding #1/#4** (aws.go) — the `NormalizeRoot` error is now checked and returned; Azure/GCP already did.
- **Finding #2** — you implemented the parent-side read-only **operation-name allowlist** (`cloudOperationIsReadOnly` + `readOnlyAWSOperations`/`gcpOperationIsReadOnly`/`azureTokenEndpointIsReadOnly`); a non-allowlisted op is denied at the broker with an audited `cspm.operation.denied`. This makes read-only server-enforced by op name, not just by which SDK calls the connectors happen to issue — a real strengthening. And ADR-0004 now records the trusted-helper authorization boundary.
- **Finding #3** — the GCP OAuth token endpoint is now explicitly allowlisted, so token acquisition is covered by the authorizer rather than only egress-bounded.
- **Executor error-wrapping** — scope-mismatch wraps `ErrForbidden`; helper-failed wraps `runErr` with `%w` + exit/timeout/truncation detail; the credential-material tripwire uses `errors.New`.

The strong core from the first review still holds (read-only across all three providers, SDK confinement enforced by the arch-test tripwire, vault-ref creds over out-of-band FD with output redaction tripwire, argv-only sandbox exec, server-side fail-closed authz, tenant RLS on both tables).

⚠️ **One heads-up for whoever lands second:** #512 (open) also claims migration `0079`. This PR takes `0079`+`0080`, so once this merges, #512 must renumber its `0079_ai_triage_evidence_policy.sql` to `0081`. I'm flagging that on #512.

Nice, thorough turnaround. Merging.

PR #513, opened 2026-08-12
feat(ai-triage): observability, budget, and resilience (P1.5)

### LINKED ISSUE #392: feat(ai-triage): observability, budget, and resilience (P1.5)

H1eu232:
## Summary

- enforce conservative per-scan token reservations and optional micro-USD cost budgets before proposer/verifier calls
- add per-role circuit breakers with advisory-only fallback while preserving the OpenAI adapter's bounded transient retry behavior
- evidence-seal request, latency, timeout, parse/provider failure, token/cost, disagreement, cache, and exemption telemetry
- add configurable baseline alerts plus a tenant-scoped API/UI dashboard by model, prompt version, CWE, and project
- document the operational controls and extend OpenAPI

## Safety contract

- budget-exhausted findings receive no partial verifier call and remain gating
- missing pricing with an enabled cost budget fails closed without contacting the provider
- provider or parse failures cannot produce verified consensus or a gate exemption
- telemetry excludes prompt text, source snippets, provider error bodies, and credentials

## Validation

- `go test ./...`
- `go vet ./...`
- `go build ./...`
- `pnpm test` (298 tests)
- `pnpm typecheck`
- `pnpm build`

Closes #392
Parent: #387

### DIFF
--- a/api/openapi.yaml
+++ b/api/openapi.yaml
@@ -1242,6 +1242,16 @@ paths:
         '401': {$ref: '#/components/responses/Unauthorized'}
         '403': {$ref: '#/components/responses/Forbidden'}
 
+  /api/v1/ai-triage/observability:
+    get:
+      tags: [ai-triage]
+      operationId: getAITriageObservability
+      summary: Aggregate tenant-scoped AI triage safety and cost metrics
+      responses:
+        '200': {description: Latest per-project AI triage metrics and baseline alerts, content: {application/json: {schema: {$ref: '#/components/schemas/AITriageDashboard'}}}}
+        '401': {$ref: '#/components/responses/Unauthorized'}
+        '403': {$ref: '#/components/responses/Forbidden'}
+
   /api/v1/redteam/halt:
     post:
       tags: [redteam]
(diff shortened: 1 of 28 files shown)


nghiadaulau (approved):
## Approved — safety contract holds, merging

Reviewed strictly (2 independent reviews + local verification against current `main`, CI disabled). Security audit: **no critical/high/medium** — the fail-closed budget/breaker/telemetry contract holds under adversarial review. go-arch: **dependency rule PASS**.

Verified:
- **Fail-closed budgets** — `reservePair` reserves proposer+verifier atomically *before* any worker runs; a budget-exhausted finding is marked `errTriageBudgetExhausted`, dropped by `mapCritiques`, absent from `result.AITriage`, and gates normally — no partial verifier call. A missing/zero price with an enabled cost budget returns false *before* `llm.Chat` — fail-closed without contacting the provider. Provider/parse failures never yield verified consensus or an exemption.
- **No secret/PII in telemetry** — `FPTriageCallMetric`/`FPTriageTelemetry` carry only IDs/role/provider/model/outcome-enum/latency/token-counts/micro-USD; `classifyCallError` returns a fixed enum, never `err.Error()`; the reservation path reads `len(content)` only; the dashboard/UI render aggregates only. No prompt/source/error-body/credential anywhere.
- **Dashboard tenant isolation** — `PermView` + tenant derived from context (no client-supplied tenant); aggregation lists engagements only via the tenant-scoped `List`/`ListProjectEngagements` (both `tenant_id=$1` / `WithTenant`), so a cross-tenant read yields nothing.
- **Circuit breaker** — an open breaker degrades to gating (advisory-only), never to auto-exempt; `GateExempt` is still set only after the full deterministic policy + `authorizedAIGateExemption` re-derivation.
- **No migration needed** — `ListProjectEngagements` selects only existing columns and filters `project_id IS NOT NULL` (project_id shipped in 0046). No schema drift.
- build/vet/gofmt clean, `-race` on fptriage/sca, config/httpapi tests, `pnpm typecheck` + 298 web tests — all green.

**Non-blocking follow-ups** (recommend for a later pass, none affect the safety contract):
1. Add the compile-time assertions `var _ ports.ProjectEngagementLister = (*EngagementRepository)(nil)` to both the memory and postgres engagement repos — the new optional interface is only checked via runtime type-assertion today, so a signature drift would silently drop the dashboard's project rows instead of failing the build.
2. Delete the now-dead `verify`/`assessOne` in coordinator.go (superseded by the observed variants) so a future edit can't diverge from the live path.
3. (security, low) Classify the `openai.New(...)` construction error before logging it in cmd/synapse-api + cmd/synapse-cli, consistent with `classifyCallError`, as defense-in-depth.

Note: #512 (currently draft) also touches this fptriage subsystem, so it'll need a rebase onto `main` after this lands. Nice, careful work. Merging.

Issue #396, opened 2026-08-05
feat(ai-triage): model/prompt promotion and drift control (P2.4)


nghiadaulau:
Parent: #387 (AI-assisted false-positive triage). Phase P2 (accuracy and scale).

## Goal

Promote new models and prompts safely, and detect drift.

## Scope

- Canary a new prompt/model in shadow mode.
- Compare against the baseline on the same dataset.
- Version and approve every promotion and rollback.
- Detect drift by language, CWE, and project distribution.

## Non-goals

- No automatic promotion; promotion is a versioned, approved decision.

## Acceptance criteria

- [ ] A new prompt/model runs in shadow and is compared to baseline before promotion.
- [ ] Promotion and rollback are versioned and approved.
- [ ] Drift alerts fire on distribution shift by language/CWE/project.

## References

- Uses the evaluation harness and shadow mode from P1.1, and the metrics/alerting from P1.5.

PR #554, opened 2026-08-13
feat(ai-triage): gate candidate promotion against baseline

XUanhoa04:
## Summary

Adds a deterministic, fail-closed candidate-vs-baseline gate for AI false-positive triage model/prompt promotion.

- Strictly decodes evaluation-v2 reports and recomputes aggregate metrics, segment breakdowns, shadow invariants, canonical model/provider identity, and `run_id` before comparison.
- Requires the exact same reviewed dataset, provenance, reviewer, gate-policy version, and independence policy; aliases of the same provider/model configuration are not accepted as a new candidate.
- Applies exact rational thresholds (no floating-point rounding at safety boundaries) for minimum precision, maximum false-negative escape rate, and allowed regression.
- Detects regressions overall and by language, kind, CWE, severity, and framework.
- Tracks verifier-comparison coverage separately so a verifier outage cannot make disagreement metrics look healthier.
- Emits source-free case behavior changes and stable machine-readable failure rules.
- Adds `synapse-fptriage-compare` plus a Make target and operator documentation.

## Safety contract

- Both inputs must remain shadow-only; any `gate_exempt=true` report is rejected.
- A passing comparison is `review_required` with `approval_required=true`; there is no automatic promotion or runtime configuration mutation.
- The comparison artifact contains typed outcomes and case IDs, not source snippets or prompts.
- Blocked evidence is written before the CLI exits non-zero for CI retention.
- The output cannot overwrite/alias either evaluation input or resolve through a symlink.
- The default proposed policy requires >=95% precision, zero true-positive escape rate, and zero regression; PM/Security approval is still required.

## Scope

This implements the candidate-vs-baseline promotion-gate slice of #396. Production distribution-drift alerts plus versioned rollout/rollback approval remain tracked by that issue, so this PR intentionally does not close it.

## Verification

- `go test -count=1 ./...`
- `go vet ./...`
- `go build ./cmd/...`
- focused Go tests and vet repeated after the final exact-threshold/filesystem hardening
- `pnpm test` (38 files, 298 tests)
- `pnpm run typecheck`
- `pnpm run build`
- `git diff --check`

Race tests were not available in the local Windows toolchain because `CGO_ENABLED=0`; the standard, focused, full-repository, vet, and build checks above are green.

Advances #396

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,8 @@ capabilities below are already shipped on `main`.
 
 ### Added
 
+- **AI-triage candidate promotion gate.** Strictly revalidates deterministic shadow reports, compares a candidate model/prompt with a baseline on the exact same reviewed dataset and policy, blocks new true-positive escapes plus overall or segment regressions, and emits stable CI evidence that still requires explicit human promotion approval.
+
 - **Evidence-cited AI false-positive triage.** Supplies bounded deterministic source/sink, data-flow, sanitizer, call-path, route/framework, and reachability metadata to the AI triage proposer/verifier before model calls. Model claims must cite current server-generated evidence-token IDs; unknown, missing, driver-incompatible, or finding-mismatched receipts fail closed. Gate authorization now uses `fp-gate-v5` and requires that closed receipt in addition to the existing severity/CWE/evidence safety floors.
 
 - **Safe AI-triage caching.** Tenant-bound API scans reuse typed proposer/verifier claims only when
--- a/Makefile
+++ b/Makefile
@@ -1,10 +1,13 @@
-.PHONY: help install tools dev build run test harness vet lint format typecheck tidy ai-triage-eval \
+.PHONY: help install tools dev build run test harness vet lint format typecheck tidy ai-triage-eval ai-triage-compare \
         docker-build docker-up docker-down clean web-dev web-build smoke
 
 GO ?= go
 IMAGE ?= synapse-api:dev
 AI_EVAL_DATASET ?= internal/usecase/sca/testdata/fptriage-golden-v1.json
 AI_EVAL_OUTPUT ?= ai-triage-eval.json
+AI_EVAL_BASELINE ?= ai-triage-baseline.json
+AI_EVAL_CANDIDATE ?= ai-triage-candidate.json
+AI_EVAL_COMPARISON ?= ai-triage-comparison.json
 
 help: ## Show this help
 	@grep -E '^[a-zA-Z_-]+:.*?## ' $(MAKEFILE_LIST) | awk 'BEGIN{FS=":.*?## "}{printf "  \033[36m%-14s\033[0m %s\n", $$1, $$2}'
@@ -49,6 +52,9 @@ tidy: ## Tidy go.mod / go.sum
 
 ai-triage-eval: ## Evaluate FP triage against the versioned golden dataset (requires two model IDs)
 	$(GO) run ./cmd/synapse-fptriage-eval --dataset $(AI_EVAL_DATASET) --output $(AI_EVAL_OUTPUT)
+
+ai-triage-compare: ## Compare candidate and baseline AI-triage shadow reports for promotion review
+	$(GO) run ./cmd/synapse-fptriage-compare --baseline $(AI_EVAL_BASELINE) --candidate $(AI_EVAL_CANDIDATE) --output $(AI_EVAL_COMPARISON)
 
 docker-build: ## Build the API container image
 	docker build -t $(IMAGE) -f deploy/Dockerfile .
--- a/README.md
+++ b/README.md
@@ -229,6 +229,7 @@ The exit code is 0 when no finding meets the threshold, non-zero otherwise.
 | `synapse-agent` | Fleet agent: host inventory and eBPF runtime detections |
 | `synapse-cluster-agent` | Fleet agent: Kubernetes workload/exposure/identity inventory |
 | `synapse-fptriage-eval` | Offline evaluation harness for AI false-positive triage |
+| `synapse-fptriage-compare` | Deterministic candidate-vs-baseline gate for AI model/prompt promotion review |
 | `synapse-fptriage-curate` | Offline privacy- and label-reviewed feedback curation for AI false-positive evaluation |
 | `synapse-mcp` | Read-only, propose-only integration server, never executes |
 
--- a/cmd/synapse-fptriage-compare/main.go
+++ b/cmd/synapse-fptriage-compare/main.go
@@ -1,0 +1,147 @@
+// Command synapse-fptriage-compare validates and compares two deterministic AI false-positive triage
+// shadow reports. It emits promotion-review evidence but never changes a runtime model or prompt.
+package main
+
+import (
+	"encoding/json"
+	"errors"
+	"flag"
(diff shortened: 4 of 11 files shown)


nghiadaulau (changes requested):
## Requesting changes — the gate is genuinely strong; one MEDIUM artifact-integrity gap to close

Reviewed strictly (2 independent reviews + local verification against current `main`, CI disabled). The promotion-gate **decision** is sound and fail-closed — verified — and the code even fixes a real aliasing bug in `fptriage_evaluation.go`. One MEDIUM undermines a stated safety claim and should be fixed before merge.

Verified locally: `go build ./...`, `go vet`, `gofmt -l` (clean), `go test -race ./internal/usecase/sca ./cmd/synapse-fptriage-compare` all pass.

### 🟠 Fix before merge (MEDIUM) — the artifact is not provably source/prompt-free
`internal/usecase/sca/fptriage_promotion.go:385-392` (`evaluationCaseOutcome`) copies `Verdict`, `Driver`, `VerifierVerdict`, `VerifierDriver` **verbatim** from the loaded report into `AIEvaluationCaseChange`, which is written into the comparison artifact for every changed case. But `AIEvaluationReport.Validate` (`:155-222`) recomputes metrics / breakdowns / run-id / model identity and rejects `gate_exempt` — yet never re-validates those four free-text-capable fields against the domain's closed token grammar (`internal/domain/judgment/claim.go` `driverRE`, ≤64 chars, no prose). Since `synapse-fptriage-compare` is designed to ingest arbitrary report files, a hand-edited/forged report can carry an arbitrary string (a source line, a prompt fragment) in `critique.driver`/`verdict`, and it lands unfiltered in the artifact.

This contradicts this PR's own contract — the `AIEvaluationCaseChange` doc says "without copying source or prompt text" (`:90-91`) and the summary says "recompute, don't trust". The recompute discipline is applied everywhere *except* these fields, so the artifact humans approve promotion against is not provably free of source/prompt content.

**Fix:** in `Validate`, re-apply the domain driver-grammar + verdict-vocabulary check to `Driver`/`VerifierDriver`/`Verdict`/`VerifierVerdict` and reject on failure (or emit only enum-validated tokens into `AIEvaluationCaseChange`). You already depend on `domain/judgment` — reuse its validator so the loader is as untrusting of these fields as it is of the metrics.

### 🟡 Should fix (LOW)
- **Evidence vs decision at the boundary** (`:327,329,331,333,354` via `ratioToBasisPoints` float `math.Round` at `:403`): the pass/fail decision correctly uses exact `big.Rat`, but the human-facing `*BasisPoints` in a *blocked* record are float-rounded, so a true 94.995% precision (correctly failing the 9500 minimum) can display as `9500` — evidence that reads "at the limit, passing" next to a "blocked" verdict. Render the displayed bps from the same `big.Rat` (truncate toward the unsafe side) so the shown number matches the compared number.
- **Symlink guard is final-component only + TOCTOU** (`cmd/synapse-fptriage-compare/main.go:94`): `os.Lstat` rejects a symlinked final path but not a symlinked parent dir, and there's a check→`os.Create` window. Inputs are well protected (exact-path + `os.SameFile`), so impact is low; resolve the output parent with `EvalSymlinks` or create with `O_NOFOLLOW`/`O_EXCL`.

### Nits (optional)
- `--fail-on-blocked=false` exits 0 on a blocked candidate (`main.go:25,80`). Fine since the artifact is authoritative (`status:"blocked"`, `approval_required:true`), but the CLI docs should say the **artifact status**, not the exit code, is the gate signal.
- Comparison ID hashes `json.Marshal(...)` with a discarded error (`:473`, mirroring `fptriage_evaluation.go:375`) — add a `// marshal cannot fail for this struct` note since the ID gates promotion.

### ✅ Verified solid
No auto-promotion / no runtime mutation (passing → only `review_required` + `approval_required`; the new API is unreferenced by any server/store/config code). Shadow-only enforced (`gate_exempt` rejected). Recompute-don't-trust for metrics/breakdowns/run-id/canonical model+provider identity. Same-dataset binding incl. provider/model aliases rejected as non-distinct. Exact-rational thresholds (`big.Rat`) at every safety boundary — no float epsilon in the decision. Verifier coverage gated separately so an outage can't inflate the disagreement metric. Deterministic (sorted keys, no time/rand); blocked artifact written before the non-zero exit; no shell/exec; README + `cli.md` document the new binary.

Close the 🟠 (re-validate the four token fields) and ideally the two 🟡, and I'll approve and merge. Nice, careful work on the exact-rational gate and the shadow/identity binding.

XUanhoa04:
Addressed all requested changes in commit 288aab3 and merged the current main branch (#555):

- Revalidated proposer and verifier verdict/driver tokens with judgment.CritiqueClaim.Validate; uncovered cases now reject critique tokens. Added forged, internally rehashed report tests for all affected fields.
- Replaced float-rounded failure evidence with exact counter-derived big.Rat BPS using conservative outward rounding. Boundary tests cover 94.995% rendering as 9499 BPS and a sub-BPS unsafe rate rendering as 1 BPS.
- Resolved output parent symlinks and made comparison artifacts create-only with O_EXCL and mode 0600. Added parent-symlink, final-symlink, and no-overwrite sentinel tests.
- Clarified that artifact status is authoritative when fail-on-blocked is disabled, documented create-only output behavior, and noted the comparison-ID marshal invariant.

Verification passed:
- go test ./...
- go vet ./...
- go build ./...
- focused tests for ./internal/usecase/sca and ./cmd/synapse-fptriage-compare
- GitGuardian Security Checks

I also attempted the requested race command. This Windows environment requires CGO for -race and does not have gcc installed, so that specific run is unavailable here; the non-race suites above pass.

@nghiadaulau please re-review when convenient.

nghiadaulau (approved):
## Approved — all findings addressed, merging

Re-reviewed the update (`288aab3`). Every item from the prior round is resolved:

- **🟠 MEDIUM (artifact not provably source/prompt-free) — FIXED.** `LoadAIEvaluationReport`/`Validate` now re-runs the domain token grammar on the free-text-capable fields: it constructs `judgment.CritiqueClaim{Verdict,Driver,Confidence}.Validate()` for the proposer (and, when populated, the verifier) and **rejects** the report on failure. A hand-edited/forged report can no longer smuggle a source line or prompt fragment through `critique.driver`/`verdict` into the comparison artifact — the recompute-don't-trust discipline now covers these fields too.
- **🟡 Evidence/decision boundary — FIXED.** Displayed basis points now come from exact `big.Rat` `floorRateBasisPoints`/`ceilRateBasisPoints` (rounded toward the unsafe side), so a candidate that fails the exact threshold can no longer display a value that reads as compliant.
- **🟡 Symlink guard — FIXED.** The output parent is resolved with `filepath.EvalSymlinks(filepath.Dir(outputAbs))` and inputs via `EvalSymlinks`, closing the symlinked-parent gap.

Verified locally against current `main` (CI disabled): `go build ./...`, `go vet`, `gofmt -l` (clean), `go test -race ./internal/usecase/sca ./cmd/synapse-fptriage-compare` all pass. The strong core from the first round is intact (no auto-promotion/runtime mutation, shadow-only enforced, recompute-don't-trust, same-dataset binding incl. alias rejection, exact-rational thresholds, verifier-coverage integrity, deterministic IDs). Nice turnaround. Merging.

PR #555, opened 2026-08-13
feat(ai-triage): detect input distribution drift

XUanhoa04:
## Summary

- add a deterministic AI-triage population snapshot to tenant observability, normalized to 10,000 basis points by language, CWE, and project
- add `synapse-fptriage-drift` to compare a saved observability response with a versioned, human-approved baseline using total-variation distance
- surface the drift input distribution in the observability UI and document the API/CLI workflow

## Safety and behavior

- uses only the latest evidence-sealed result for each tenant-visible project and exports no source or prompt content
- treats missing language/CWE metadata explicitly as `unknown` / `unclassified`
- strictly validates baseline schema, canonical distributions, exact basis-point totals, minimum samples, and human approval provenance; reserved machine actors fail closed
- produces deterministic report IDs and writes the report before returning a non-zero drift or insufficient-sample alert
- has no authority to promote a model, change a prompt, suppress a finding, or alter the quality gate

This complements #554 while remaining independently mergeable: #554 covers candidate-vs-baseline promotion evidence, while this PR covers the language/CWE/project drift acceptance criterion.

## Verification

- `go test ./...`
- `go vet ./...`
- `go build ./cmd/...`
- `pnpm test` (298 tests)
- `pnpm run typecheck`
- `pnpm run build`
- `git diff --check`

Advances #396

PR #559, opened 2026-08-14
feat(ai-triage): govern model promotion and rollback

### LINKED ISSUE #396: feat(ai-triage): model/prompt promotion and drift control (P2.4)

H1eu232:
## Summary

Completes the remaining governance slice of P2.4 after #554 and #555:

- add synapse-fptriage-release for independently approved model/prompt promotions and rollbacks
- recompute the candidate comparison from the complete baseline and candidate shadow reports at the release boundary
- require distinct PM and Security human approvals bound to the exact manifest and current ledger head
- record unique release versions in a create-only, hash-chained ledger
- restrict rollback targets to the initial baseline or an earlier approved decision
- document the promotion, approval, canary deployment, monitoring, and rollback workflow

## Safety contract

- release artifacts have no runtime authority and never mutate models, prompts, thresholds, findings, or gate configuration
- stored comparison status and IDs are not trusted without full report recomputation
- reserved machine principals and proposer/verifier model identities cannot approve a release
- approvals cannot be replayed after the ledger head changes
- existing evidence files and input artifacts are never overwritten

## Verification

- go test ./...
- go vet ./...
- go build ./...
- go test -race ./internal/usecase/sca ./cmd/synapse-fptriage-release
- git diff --check

Closes #396

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,8 @@ capabilities below are already shipped on `main`.
 
 ### Added
 
+- **Versioned AI-triage release governance.** Recomputes passing candidate comparisons at the approval boundary and records promotions or rollbacks in a create-only, hash-chained ledger. Every decision requires distinct PM and Security approvals bound to the current ledger head; no release artifact mutates runtime model, prompt, threshold, or gate configuration automatically.
+
 - **AI-triage candidate promotion gate.** Strictly revalidates deterministic shadow reports, compares a candidate model/prompt with a baseline on the exact same reviewed dataset and policy, blocks new true-positive escapes plus overall or segment regressions, and emits stable CI evidence that still requires explicit human promotion approval.
 
 - **Judgment-gated cross-pillar promotion.** Correlates deterministic reachability, confident internet-facing attack paths, and active same-path runtime detections into tenant-scoped priority proposals. Distinct sealed verification applies one-level changes, uncertain inputs remain review-only, and signal loss restores the exact prior priority through an append-only, auditable lifecycle.
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,4 @@
-.PHONY: help install tools dev build run test harness vet lint format typecheck tidy ai-triage-eval ai-triage-compare ai-triage-drift \
+.PHONY: help install tools dev build run test harness vet lint format typecheck tidy ai-triage-eval ai-triage-compare ai-triage-release ai-triage-drift \
         docker-build docker-up docker-down clean web-dev web-build smoke
 
 GO ?= go
@@ -8,6 +8,9 @@ AI_EVAL_OUTPUT ?= ai-triage-eval.json
 AI_EVAL_BASELINE ?= ai-triage-baseline.json
 AI_EVAL_CANDIDATE ?= ai-triage-candidate.json
 AI_EVAL_COMPARISON ?= ai-triage-comparison.json
+AI_RELEASE_MANIFEST ?= ai-triage-release-manifest.json
+AI_RELEASE_LEDGER ?=
+AI_RELEASE_OUTPUT ?= ai-triage-release-ledger.json
 AI_DRIFT_BASELINE ?= ai-triage-drift-baseline.json
 AI_DRIFT_OBSERVED ?= ai-triage-observability.json
 AI_DRIFT_OUTPUT ?= ai-triage-drift-report.json
@@ -58,6 +61,9 @@ ai-triage-eval: ## Evaluate FP triage against the versioned golden dataset (requ
 
 ai-triage-compare: ## Compare candidate and baseline AI-triage shadow reports for promotion review
 	$(GO) run ./cmd/synapse-fptriage-compare --baseline $(AI_EVAL_BASELINE) --candidate $(AI_EVAL_CANDIDATE) --output $(AI_EVAL_COMPARISON)
+
+ai-triage-release: ## Append a PM/Security-approved AI-triage promotion to a new release ledger
+	$(GO) run ./cmd/synapse-fptriage-release --manifest $(AI_RELEASE_MANIFEST) $(if $(AI_RELEASE_LEDGER),--ledger $(AI_RELEASE_LEDGER),) --comparison $(AI_EVAL_COMPARISON) --baseline $(AI_EVAL_BASELINE) --candidate $(AI_EVAL_CANDIDATE) --output $(AI_RELEASE_OUTPUT)
 
 ai-triage-drift: ## Compare AI triage input distribution with a human-approved baseline
 	$(GO) run ./cmd/synapse-fptriage-drift --baseline $(AI_DRIFT_BASELINE) --observed $(AI_DRIFT_OBSERVED) --output $(AI_DRIFT_OUTPUT)
--- a/README.md
+++ b/README.md
@@ -230,6 +230,7 @@ The exit code is 0 when no finding meets the threshold, non-zero otherwise.
 | `synapse-cluster-agent` | Fleet agent: Kubernetes workload/exposure/identity inventory |
 | `synapse-fptriage-eval` | Offline evaluation harness for AI false-positive triage |
 | `synapse-fptriage-compare` | Deterministic candidate-vs-baseline gate for AI model/prompt promotion review |
+| `synapse-fptriage-release` | Versioned PM/Security-approved promotion and rollback ledger for AI triage |
 | `synapse-fptriage-curate` | Offline privacy- and label-reviewed feedback curation for AI false-positive evaluation |
 | `synapse-fptriage-drift` | Offline language/CWE/project distribution drift evidence for AI triage |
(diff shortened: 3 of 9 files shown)


nghiadaulau (changes requested):
## Requesting changes — strong governance tool, but one MEDIUM lets a regressing candidate be promoted against a cherry-picked baseline

Reviewed strictly (2 independent reviews + local verification against current `main`, CI disabled). This is well-built, fail-closed offline governance and almost all of it is verified sound — but the promotion gate's "recompute, don't trust" guarantee is anchored to an operator-chosen baseline rather than the deployed one, which is the whole point of the mechanism. Please close that before it merges (and it shouldn't `Closes #396` until it does).

Verified locally: `go build ./...`, `go vet`, `gofmt -l` (clean), `go test -race ./internal/usecase/sca ./cmd/synapse-fptriage-release` all pass.

### 🟠 Fix before merge (MEDIUM) — the promotion baseline is anchored by *config*, not the deployed run's sealed ID
`internal/usecase/sca/fptriage_release.go:196` (and the mirror in `Validate` at `:325`). For any promotion after the first, the only tie between the supplied comparison baseline and the currently-active configuration is `sameEvaluationConfiguration(currentReleaseRun(ledger), comparison.BaselineRun)` — provider + model-family + prompt-version equality. It never requires `comparison.BaselineRunID == currentReleaseRunID(ledger)`, even though `currentReleaseRunID` is a content hash that commits to the previously-promoted candidate report's metrics (it's used only to record `PreviousActiveRunID`, never to gate).

Because LLM evaluations are non-deterministic, an operator can supply — as the comparison baseline — a *different, deliberately weaker re-evaluation of the same config*. The regression gate in `CompareAIEvaluationReports` measures candidate-vs-supplied-baseline, so a candidate that actually regresses against the *deployed* baseline shows "no regression / improvement," and your recompute + `reflect.DeepEqual` still pass. That defeats the stated "recompute, don't trust the human's read of the metrics" — the recompute is anchored to an operator-chosen baseline, not the sealed active run. (Mitigated, not closed, by PM/Security signing the `ComparisonID` which commits to `BaselineRunID` — but that pushes the whole guarantee back onto the humans the recompute is meant to backstop.)

**Fix:** in `AIEvaluationReleaseReviewDigest` require `evidence.Comparison.BaselineRunID == currentReleaseRunID(ledger)` for non-initial promotions, and assert the same in `Validate()` using each decision's `PreviousActiveRunID`, so the promotion baseline is provably the exact previously-promoted run.

### 🟡 Fix before merge (LOW, compounds the MEDIUM) — no policy floor; the recompute trusts the comparison's embedded policy
`fptriage_release.go:94` recomputes with `e.Comparison.Policy`, and `AIEvaluationPromotionPolicy.Validate` (`fptriage_promotion.go:43`) only bounds each field to 0–10000 — no floor, no binding to an approved policy. A comparison that declares a permissive policy (`MinimumPrecisionBasisPoints=0`, all max-drops `10000`) produces zero failures → `status="review_required"` → the recompute matches. So "passing" is only as strong as the self-declared policy. Combined with the MEDIUM, an operator controls both the baseline and the bar. Pin/validate the policy against `DefaultAIEvaluationPromotionPolicy` (or a ledger-recorded approved policy).

### 🟡 Should fix
- **Docs (README binaries table)** — `synapse-fptriage-release` is missing from the README binaries table (all sibling `synapse-fptriage-*` offline tools are listed). Add a row + mirror in `docs/guide/cli.md`.
- **Rollback approver SoD** (`fptriage_release.go:455-467`) — `releaseModelIdentities` for a rollback returns only the *current* active run's models, omitting the rollback **target** run's models, so creation-time `validateReleaseApprovals` wouldn't reject a human whose id equals the target model. It's caught by the final `ledger.Validate()` (which passes the target's models), so not exploitable — but make the two model sets consistent so a future refactor can't drop the net.

### Nits (optional)
- Wrap the validation-class errors with `%w` + `shared.ErrValidation` to match the rest of the `sca` package (the file doesn't import `shared` today).
- Handle (don't discard) the marshal errors feeding the security-critical digests/IDs (`:229,482,490`) — can't fail for these types, but it's fragile.
- Input paths aren't `EvalSymlinks`-resolved in `sameReleasePath`; not exploitable because the ledger write is `O_EXCL` create-only, but resolving symlinks would harden the "don't overwrite an input" guard.

### ✅ Verified solid
No runtime authority (the only write is the O_EXCL create-only ledger file — no model/prompt/threshold/finding/gate mutation; the release symbols have no runtime consumer). Distinct human PM+Security with roles `pm`/`security`, machine/model approvers rejected via `IsMachineOrModelActor`, case/whitespace-normalized distinctness. Approvals bound to the ledger head + version + comparison/target, void once the head advances (no replay). Append-only hash-chained ledger re-verified in `Validate` (sequence, `PreviousDecisionID`, `DecisionID==releaseDecisionID`, head==tail). Rollback restricted to `"initial"` or an existing approved decision and re-derived against the ledger prefix (no roll-forward to an unapproved candidate). Deterministic sha256-over-canonical-JSON IDs; strict decode; no source/prompt/secret in the ledger; no shell/exec.

Close the 🟠 (bind the baseline run) + the 🟡 policy floor, and I'll approve and merge. The SoD/ledger/no-runtime-authority design is exactly right.

H1eu232:
Addressed the requested changes in `4b6c4a5`:

- Exact active baseline: non-initial promotions now require both `BaselineRunID == currentReleaseRunID(ledger)` and the exact active run metadata. Ledger validation mirrors this for every promotion using `PreviousActiveRunID`. Added adversarial tests for a fresh weaker re-evaluation of the same model/prompt configuration and for a rehashed ledger with a substituted baseline ID.
- Release policy floor: release evidence now rejects policies weaker than `DefaultAIEvaluationPromotionPolicy`; stricter policies remain valid. The applied policy is persisted in each promotion decision and revalidated from the ledger. Added tests for weak, strict, and rehashed weak-policy cases.
- Rollback SoD: creation-time approval checks now include both current and rollback-target model identities.
- Docs: added `synapse-fptriage-release` to the README binaries table and concrete usage to the CLI guide.
- Hardening nits: digest/decision JSON marshal errors are propagated, and path alias checks resolve symlinks (with regression coverage).

Local verification:
- `go test ./...`
- `go vet ./...`
- `go build ./...`
- `go test -race ./internal/usecase/sca ./cmd/synapse-fptriage-release`
- targeted `gofmt -l` + `git diff --check`

nghiadaulau (approved):
## Approved — both blockers fixed, merging

Re-reviewed the update (`4b6c4a5`). Both items from the prior round are closed, and in both required places:

- **🟠 MEDIUM (baseline anchored by config, not the deployed run) — FIXED.** `AIEvaluationReleaseReviewDigest` now rejects a non-initial promotion unless `comparison.BaselineRunID == currentReleaseRunID(ledger)` **and** the baseline run DeepEquals the active run (`fptriage_release.go:200`, "promotion baseline does not match the exact active approved run"), and `Validate` re-checks `decision.PreviousActiveRunID == activeRunID` per decision (`:322`). An operator can no longer swap in a weaker re-eval of the same config as the baseline — the approved digest is bound to the exact previously-promoted run.
- **🟡 LOW (no policy floor) — FIXED.** New `validateReleasePromotionPolicy` pins the comparison policy to the `DefaultAIEvaluationPromotionPolicy()` floor and is enforced on both the creation path (`:92`) and the ledger re-validation (`:348`); a policy weaker than the built-in safety floor is rejected.
- **Docs — FIXED.** `synapse-fptriage-release` added to the README binaries table and `docs/guide/cli.md`.

Verified locally against current `main` (CI disabled): `go build ./...`, `go vet`, `gofmt -l` (clean), `go test -race ./internal/usecase/sca ./cmd/synapse-fptriage-release` all pass. The strong core from the first round is intact (no runtime authority; distinct human PM+Security with machine/model approvers rejected; approvals bound to the ledger head with no replay; append-only hash-chained ledger; rollback restricted to initial/approved decisions; deterministic; create-only O_EXCL writes).

Optional follow-ups (non-blocking): wrap the validation errors with `shared.ErrValidation`; handle the discarded marshal errors feeding the digests; make the rollback approver model-set consistent between creation-time and `Validate`; `EvalSymlinks` on the input-path guard. Nice, complete turnaround. Merging.

PR #562, opened 2026-08-14
feat(ai-triage): gate promotion on adversarial invariance

XUanhoa04:
## Summary

- introduce reviewed `control` / adversarial `challenge` counterfactual groups in golden dataset v2, with strict validation that semantic finding fields remain unchanged
- emit deterministic, source-free pair evidence in evaluation report v3 for proposer, verifier, consensus, and shadow-policy invariance
- upgrade comparison v2 to block promotion on incomplete pair/verifier coverage or configured verdict/policy flip rates; unsafe challenge-only policy flips always block
- bind the robustness policy and evidence into release-ledger v2, including a documented migration requiring fresh PM/Security approval
- expose exact basis-point thresholds in the compare CLI and extend metrics, golden fixtures, tests, README, guide, and changelog

## Safety properties

- each reviewed group has exactly one non-adversarial control and at least one adversarial challenge
- label, language, framework, kind, severity, CWE, title, description, and line must match within a group; only source perturbation/file identity may differ
- evaluation validation recomputes pair evidence and canonical IDs, so edited reports cannot forge robustness results
- promotion decisions use exact integer counters rather than rounded floating-point metrics
- zero reviewed pairs fail the default coverage gate; zero verifier-required pairs are complete by construction
- AI triage remains shadow-only and receives no runtime gate authority

## Verification

- `make ai-triage-verify`
- `go test -count=1 ./internal/usecase/sca ./cmd/synapse-fptriage-eval ./cmd/synapse-fptriage-compare ./cmd/synapse-fptriage-release ./cmd/synapse-fptriage-curate`
- `go vet ./...`
- `go build ./...`
- `git diff --check upstream/main`

The full suite also exposed the pre-existing flaky `internal/usecase/vulnerabilityreconciliation/TestAdvisoryReconciliationPinsSnapshotRevision` in an untouched package (isolated reproduction: 1 pass / 4 failures with `-count=5`). Race tests require CGO and a C compiler, which is not installed in the local Windows environment.

Follow-up to #387 and #388.


nghiadaulau (approved):
## ✅ Approved — merging (strict review, CI disabled → verified locally)

Reviewed hard against the golden rules (this is the AI-triage safety gate) with both an architecture and a security pass. **No blockers, majors, or CRITICAL/HIGH/MEDIUM findings.** Verified locally: `go build ./...`, `go vet`, `gofmt -l` clean, full `internal/usecase/sca/...` + `cmd/synapse-fptriage-{eval,compare,release}` suites green, and the `make ai-triage-verify` gate (from #388) passes against the renamed v2 dataset.

**What I verified holds:**
- **Golden rule 5 (shadow / `gate_exempt`) — triple-guarded.** Evaluation runs `aiTriageModeShadow`; the only site that sets `GateExempt=true` is short-circuited by the `mode != enforce` branch, plus the inline hard-reject in `EvaluateFPTriage` and the report-load invariant. The new counterfactual/challenge groups feed the same policy path and cannot smuggle an exemption.
- **Determinism & source-free report v3.** No `Source` field in results (only `DatasetSHA256`); model text constrained to closed driver tokens (`driverRE`, ≤64, no free prose), re-enforced on load; `RunID`/`ComparisonID` exclude themselves + sort results + marshal maps; no clock/rand. Reproducible bytes.
- **Fail-closed promotion.** Clean → `review_required` (never an "approve" state); any failure → `blocked`. Zero/incomplete counterfactual pairs, an uncovered control/challenge, or a lost verifier on a refuted challenge all block; `counterfactual_unsafe_policy_flip` blocks unconditionally. `math/big.Rat` basis-points avoid float bypass; policy floors pinned and the whole comparison is recomputed + `DeepEqual`'d, so a forged/weakened artifact is rejected.
- **Adversarial honesty.** `sameCounterfactualDefinition` forces control and challenge to share every semantic field (only File/Source may differ); a mismatch fails closed in both dataset and report validation.
- **Judgment alignment.** Release requires exactly one PM + one Security (distinct humans, both approved, digest-bound, no machine/model self-approval via `IsMachineOrModelActor`); `verdict.EvidenceThreshold` (75) unchanged; ledger hash-chained + append-only.
- Schema bumps (dataset v2 / eval report v3 / comparison v2 / ledger v2) are consistent across code, cmd defaults, Makefile, docs; old artifacts **fail closed** (exact schema check + `DisallowUnknownFields`).
- New invariance/flip-rate/coverage guards are real negative tests (assert exact counters + named failure rules, recompute + `DeepEqual` before validate).

**Non-blocking follow-ups (suggestions, not gating):**
1. `fptriage_promotion.go` — `verifierCoverage` appears unused (dead since before this PR; the rewrite left it). Safe to delete in a follow-up.
2. `EvaluateFPTriage` emits the report without a final `report.Validate()` (only enforced at the compare-step load). Harmless today (shadow invariant is hard-checked inline; compare rejects prose), but a final `Validate()` in the emit path would make the artifact self-certifying.
3. Doc note: a feedback-curated dataset (no counterfactual pairs) is now un-promotable under the 100% counterfactual-coverage floor — fail-closed and correct, worth calling out in `docs/guide/ai-triage-evaluation.md` so it isn't a surprise.

Great work on a genuinely tricky safety surface — the fail-closed direction and the recompute-before-trust design are exactly right. Merging.

PR #565, opened 2026-08-15
Redesign engagements queue and creation flow

nnatuan03:
## Summary
- keep `/engagements` focused on the assessment queue
- move the creation form to `/engagements/new`
- update navigation, dashboard, and Asset shortcuts to the new route
- redirect legacy `?create=1` links to the dedicated creation route

## Validation
- `npm test` — 39 files, 311 tests passed
- `npm run typecheck`
- `npm run build`

### DIFF
--- a/web/src/App.rules.test.tsx
+++ b/web/src/App.rules.test.tsx
@@ -151,16 +151,16 @@ describe('App Routing - Rules', () => {
     expect(document.documentElement.dataset.theme).toBe('light')
   })
 
-  it('shows only New Engagement for the Engagement workspace', () => {
+  it('links to the Engagements queue from the primary navigation', () => {
     render(
       <MemoryRouter initialEntries={['/assets']}>
         <Sidebar />
       </MemoryRouter>
     )
 
-    const newEngagement = screen.getByRole('link', { name: 'New Engagement' })
-    expect(newEngagement.getAttribute('href')).toBe('/engagements?create=1')
-    expect(screen.queryByRole('link', { name: 'Engagements' })).not.toBeInTheDocument()
+    const engagements = screen.getByRole('link', { name: 'Engagements' })
+    expect(engagements.getAttribute('href')).toBe('/engagements')
+    expect(screen.queryByRole('link', { name: 'New Engagement' })).not.toBeInTheDocument()
   })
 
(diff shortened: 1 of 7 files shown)


nghiadaulau (approved):
## ✅ Approved — merging (strict frontend review, CI disabled → verified locally)

Clean, well-executed route split. Verified locally: `pnpm typecheck` clean, **39 files / 311 tests pass**, `pnpm build` OK. Design-system + a11y review passed:

- **Tokens: PASS** — no raw hex / arbitrary color values; every color is a `@theme` token. Icon swap `SquarePen → Target` is lucide-react (no emoji-as-icons).
- **States preserved** — `Spinner` / `EmptyState` / `ErrorState` all kept on the queue.
- **a11y clean** — CTAs are native anchors, keyboard-focusable, each with the same `focus-visible:ring-2 ring-brand/60 ring-offset` as `Button`; single `h1` per page; new `/engagements/new` header matches app conventions (`max-w-[1480px] animate-fade-in`, uppercase eyebrow).
- **Redirect is safe** — `?create=1 → /engagements/new` via `<Navigate replace>`, no loop/trap.

**Follow-ups (non-blocking, please address in a small follow-up):**
1. **MEDIUM — extract a shared Button-as-link pattern.** The primary-CTA class string (`btn-primary inline-flex … focus-visible:…`) is copy-pasted in 3 places + a hand-rolled `secondary` link, reconstructing `BTN.brand`/`BTN.secondary` from `components/ui.tsx`. They've already drifted (missing `select-none` / `duration-150`). Give `Button` an `asChild`/`as` prop (or export a `buttonClasses(variant)` helper) and render `<Button asChild variant="brand"><Link …/></Button>` so tokens stay in one place. Tokens themselves are correct — this is maintainability, not a token violation.
2. **LOW** — the `create===1` redirect is evaluated in render after `useEffect(load)` already fired two fetches; a route-level or pre-effect guard avoids the wasted requests.

Neither blocks. Merging.

**Heads-up for #566:** this PR moves creation to a dedicated `/engagements/new` route and removes the `?create=1` inline-form model. #566 still uses `?create=1` (its sidebar CTA + a `setCreating` sync `useEffect`), so after this merges #566 will need a rebase onto the route model — details in a comment there.

PR #567, opened 2026-08-15
feat(sla): complete governed remediation lifecycle

### LINKED ISSUE #80: Epic: risk-based remediation SLA and finding lifecycle
> **Status:** not started (Phase 0). Updated 2026-07-12.
>
> **Relation to other work:** this epic owns the *risk-based remediation SLA* — tiers, due dates, and accept-risk-with-a-hard-expiry. The Code Quality product (Epic #174) separately adds a *triage-resolution* lifecycle in #181 (open → accepted / false-positive / won't-fix). These are **complementary axes** (remediation **state** vs. validity **resolution**) — the same status-vs-resolution split mature analyzers keep separate. Implement both over **one** `internal/domain/finding` status model rather than two parallel ones, and reuse the already-shipped governance substrate this epic assumes: the Judgment primitive, hash-chained evidence, and the append-only audit log are all live, so Phase 0–3 build on them directly. Cross-refs: #174, #181, #159.

---

Findings today are ordered by risk (KEV, then EPSS, then CVSS), but ordering is not governance. A finding has no due date, no remediation state beyond open, and no first-class way to accept a risk with an expiry. This epic adds a deterministic, risk-based remediation SLA and a finding lifecycle on top of what already exists, staying inside the golden rules: pure Go scoring, versioned config folded into the result for reproducibility, no LLM in the path, every mutation audited.

## Goal

Given a finding, deterministically assign a remediation SLA tier and concrete due dates from its real risk context, track it through a remediation lifecycle, and let a reviewer accept a risk with a compensating control and a hard expiry. Surface all of it in the API, the CLI, and the report, and let CI optionally gate on overdue findings.

## Design overview

- New pure domain package `internal/domain/sla` with a single `Compute(inputs, cfg, now)` entry point. No I/O, clock injected, output fully determined by inputs plus a versioned `Config`.
- A `slauc.Service` use case computes the SLA when findings are derived, persists SLA state, and owns the lifecycle transitions (all audited).
- SLA config is versioned and stored, DB value takes precedence over the built-in default, and the config version is recorded on every result so a decision is reproducible and explainable.
- Inputs are read from data Synapse already has: `finding.Severity`, `finding.KEV`, the vulnerability EPSS score, `FixState` / `FixedVersion` for patch feasibility, plus optional asset context (criticality, exposure) that defaults to a neutral bucket when unknown, so the feature works before any asset model exists.

## Data model

`internal/domain/sla`:

```
Tier    = Emergency | Critical | High | Medium | Low | Exception
Inputs  { Severity, CVSSScore, KEV, EPSS, PublicPoC, ActiveExploitation,
          Criticality, Exposure, Feasibility }
Config  { Weights, Thresholds, TierDueRanges, OverrideRules, Version }   // versioned
Breakdown { Severity, Exposure, Criticality, Exploitability, ThreatIntel, Feasibility, Overrides[] }
Result  { Tier, Score(0..100), Breakdown, MitigateBy, RemediateBy, Reason, ComputedAt, ConfigVersion }
```

Persisted per finding (new table): tier, score, breakdown JSON, mitigate_by, remediate_by, config_version, lifecycle status, exception fields.

## Scoring model (deterministic)

Score is a bounded 0..100 sum of weighted factors, then mapped to a tier by thresholds, then override rules can force a tier. All weights and thresholds live in `Config`.

- Severity from CVSS band (critical/high/medium/low).
- Exploitability from KEV, EPSS band, public PoC, active exploitation (capped).
- Exposure and criticality from asset context (neutral default when unknown).
- Feasibility adjustment: patch available is neutral, change-window / compensating-control / no-patch reduce urgency and can route to the Exception tier.
- Tier maps to a mitigate-by and remediate-by duration range; due dates are `now + range`.

### Phase 0: domain scoring (foundation)
Goal: `sla.Compute` produces a stable tier, score, breakdown, and due dates from inputs plus config, with no I/O.
Start: `internal/domain/sla/{sla.go, score.go, config.go, sla_test.go}`.
- [ ] `Tier`, `Inputs`, `Config`, `Breakdown`, `Result` types
- [ ] `Compute(Inputs, Config, now)` pure function, clock injected
- [ ] Built-in `DefaultConfig()` with documented weights, thresholds, tier due ranges
- [ ] Override rules (for example KEV plus internet exposure escalates one tier)
- [ ] Table-driven tests plus a property test that score stays within 0..100 and tier is monotonic in score
Acceptance: `make build vet test` green; determinism test (same inputs plus config yield identical result); no imports outside stdlib and `internal/domain`.

### Phase 1: wire into the findings pipeline (read-only value)
Goal: every derived finding carries a tier and due dates, visible without any new storage.
Start: `internal/usecase/sca/findings.go`, `service.go` (`ScanResult`), `cmd/synapse-cli/main.go`, `internal/adapter/httpapi/finding_handler.go`.
- [ ] Map a finding plus its vulnerability to `sla.Inputs` (derive feasibility from `FixState`/`FixedVersion`)
- [ ] Compute during `buildFindings`, attach `Tier`, `MitigateBy`, `RemediateBy`, `ConfigVersion` to the finding or a parallel `ScanResult.SLA` map
- [ ] CLI `--json` and the human report show tier and due date; API finding response includes SLA
- [ ] Gate flag `SYNAPSE_SLA_ENABLED` (default off in phase 1)
Acceptance: a scan shows a tier and due date per finding; deterministic across runs; report path stays LLM-free.

### Phase 2: versioned config and persistence
Goal: SLA config and per-finding SLA state are stored, and config is editable without a code change.
Start: `internal/usecase/ports` (new `SLAConfigStore`, `FindingSLAStore`), `internal/infrastructure/persistence/{memory,postgres}`, a new numbered migration in `migrations/`.
- [ ] `ports.SLAConfigStore` (GetActive, Insert new version) and `ports.FindingSLAStore` (Upsert, Get, ListOverdue)
- [ ] Memory and Postgres implementations, compile-time port assertions
- [ ] Migration `NNNN_sla.sql`: `sla_config` (versioned, active pointer) and `finding_sla`
- [ ] DB config takes precedence over `DefaultConfig()`; empty means use default
Acceptance: config round-trips; a new config version is a new row and does not mutate history; hostile harness and tenant checks still pass.

### Phase 3: remediation lifecycle and exceptions
Goal: a finding moves through a governed lifecycle and a reviewer can accept a risk with a compensating control and a hard expiry.
Start: `slauc.Service`, `internal/adapter/httpapi` routes, audit wiring.
- [ ] Status set: open, mitigating, remediated, accepted-risk
- [ ] Accept-risk requires a reason, a compensating control note, and an expiry; on expiry the finding reverts to its computed tier and becomes overdue again (fail-safe, mirrors the suppression expiry rule)
- [ ] Every transition is an append-only audit event with an actor id
- [ ] HTTP routes behind `PermReview` plus the `withEngTenant` chokepoint; read behind `PermView`; no lifecycle mutation via any propose-only surface
- [ ] Overdue detection query (`remediate_by < now` and status not remediated/accepted)
Acceptance: expiry re-arms the SLA; RBAC and tenant isolation enforced; harness green.

### Phase 4: asset risk context (optional input, larger)
Goal: criticality and exposure come from a real per-engagement or per-target context instead of the neutral default, sharpening the score.
Start: extend `engagement` scope or a light asset-context store; feed `sla.Inputs`.
- [ ] Per-target or per-engagement criticality and exposure fields
- [ ] Compute uses them when present, neutral default otherwise
Acceptance: setting exposure to internet-facing escalates the tier as expected; unset behaves exactly as phase 1.

### Phase 5: CI gate and reporting
Goal: SLA is actionable in CI and in deliverables.
Start: `cmd/synapse-cli/main.go`, report and SARIF paths.
- [ ] CLI `--fail-on-overdue` exits non-zero when an overdue finding exists
- [ ] SLA tier and due date in SARIF result properties and in the PDF/report path
- [ ] Optional portfolio view: counts by tier and overdue rate
Acceptance: the gate flips exit code correctly; report renders tier and due date; deterministic.

## Config

```
SYNAPSE_SLA_ENABLED=false   # phase 1 default; flip to true once phases 2-3 land
```

Weights, thresholds, and tier due ranges are config, not code, and are versioned. The default config is documented in `DefaultConfig()`.

## Golden-rule compliance

- Deterministic and LLM-free: scoring is pure Go, the report path never calls a model.
- Reproducible: the config version is recorded on every result, so a tier can be re-derived and explained.
- Auditable: every lifecycle transition and every config change is append-only with an actor id.
- Fail-safe: an expired risk acceptance re-arms the SLA rather than silently persisting.
- Gated writes: lifecycle mutations go through `PermReview` and the tenant chokepoint only.

## Out of scope (separate future epics)

Asset inventory as a first-class module, security-control register and framework mapping, evidence-request workflow and audit-pack export, DB-driven license policy with a confirmation workflow, and CPE-based correlation for OS and infrastructure assets. Each is its own epic and is not required for this one.

## Definition of done

Phases 0 through 3 merged, `SYNAPSE_SLA_ENABLED` documented, tests and the hostile harness green, and a scan shows a governed SLA tier with due dates that a reviewer can accept with an expiry. Phases 4 and 5 can follow independently.

## Continuous intelligence integration boundary

Related programme: #514. Integration issue: #540.

- #514 owns provider observations, canonical advisory revisions, component occurrences, immutable risk assessments, and risk-change transitions.
- This epic continues to own SLA scoring/config versions, due dates, remediation/accepted-risk lifecycle, expiry, overdue policy, CI gates, and reporting.
- Continuous intelligence may request SLA recomputation when relevant assessment inputs change, but it must not introduce a second status model or automatically remediate/reopen a finding.

### LINKED ISSUE #540: feat(risk): integrate continuous assessments with remediation SLA and retest governance
Parent: #518. Related: #80.

## Objective

Define the seam between continuous intelligence risk changes and the remediation SLA/retest lifecycle without creating a second workflow.

## Ownership boundary

- This feature supplies current immutable risk assessment inputs and transition events.
- #80 owns SLA tier/config, due dates, accepted-risk expiry, and remediation lifecycle policy.
- Existing finding retests remain append-only and authoritative for remediation confirmation.

## Scope

- Recompute SLA only when relevant assessment inputs change.
- Escalation may shorten due dates under #80 policy; de-escalation follows its versioned rules.
- No-longer-detected and withdrawn state create `retest_required`, not automatic remediation.
- A remediated/false-positive finding that is re-exposed creates review action without silently reopening.

## Acceptance criteria

- Integration uses narrow ports and no package cycle.
- Accepted-risk and remediation decisions survive intelligence refresh.
- Expired accepted risk re-arms according to #80, not monitor-specific logic.
- End-to-end tests cover escalation and re-exposure of terminal findings.

XUanhoa04:
## Summary

Completes the operational phases of risk-based remediation SLA governance and connects them to continuous vulnerability intelligence.

- adds immutable, deterministic SLA assessments with versioned tenant policy and deadlines that do not slide on replay or reassessment
- adds a separate human-owned remediation lifecycle (`open`, `mitigating`, `remediated`, `accepted_risk`) with optimistic concurrency, append-only events, compensating controls, and hard acceptance expiry
- wires scan/API/CLI/web output and continuous CVSS/KEV/EPSS/PoC/exploitation refresh through narrow ports while preserving human decisions
- adds tenant-safe memory/Postgres stores, migration `0103_sla_governance.sql`, RLS, composite ownership FKs, OpenAPI contracts, configuration, and operator documentation
- fixes historical risk-assessment re-promotion so `no_longer_detected -> detected` creates re-exposure review work even when the immutable assessment already exists

This is an opt-in capability (`SYNAPSE_SLA_ENABLED=false` by default) and adds 5,085 lines across domain, persistence, integration, API, UI, documentation, and tests.

## Governance guarantees

- policy versions are DB-enforced append-only; retained assessment artifacts and lifecycle events reject UPDATE/TRUNCATE while DELETE follows the intentional finding/engagement/project cascade; current pointers are separate
- the first assessment fixes `deadline_anchor_at` for its chain; reassessments rebase to it and keep the earlier old/new deadline, so escalation, de-escalation, policy changes, and rescans cannot reset or extend the clock
- identical SLA-relevant facts reuse the original assessment, including provenance-only refreshes and CVSS/EPSS drift inside the same effective scoring band
- material risk or policy changes link a new assessment through `previous_assessment_id`
- machine identities cannot transition remediation state or activate policy
- the shared machine-principal classifier treats empty/whitespace identities as machine (fail closed); existing AI-triage decision and drift callers already require non-empty actor text before classification
- intelligence refresh can advance assessment provenance but cannot overwrite lifecycle status, reason, actor, expiry, timestamps, or version
- a transition racing a risk refresh conflicts instead of rolling the assessment pointer backward
- accepted risk becomes effectively open exactly at expiry
- remediated remains terminal; re-exposure creates explicit review work without silently reopening it
- every read/write is tenant-bound and every SLA table has forced RLS

## Review follow-up (`09555f8`)

- replaces the NUL-containing PostgreSQL text advisory-lock input with a stable SHA-256-derived `int64` lock key
- anchors/caps reassessment deadlines and canonicalizes material-input fingerprints
- makes post-persistence SCA SLA enrichment best-effort while retaining successful partial enrichment
- removes duplicated vulnerability-risk-to-SLA mapping and uses `slauc.InputsFromRiskAssessment`
- adds UPDATE/DELETE/TRUNCATE guards for policies and UPDATE/TRUNCATE immutability guards for retained assessments and lifecycle events
- maps PostgreSQL lifecycle-event unique violations to `shared.ErrConflict` and removes the redundant policy uniqueness constraint
- adds regression coverage for late emergency escalation, de-escalation, non-material drift, advisory lock identity, append-only mutations, partial enrichment, and conflict classification

## Review follow-up round 2 (`fc4321d`)

- preserves intentional finding/engagement/project teardown by allowing assessment/event DELETE cascades while retaining UPDATE/TRUNCATE immutability
- adds explicit cascading ownership for multi-hop assessment chains plus lifecycle assessment references
- replaces trigger-bypassing test cleanup with a production `projectuc.Delete` integration test covering a three-assessment chain
- aligns the memory adapter's missing-predecessor error with PostgreSQL `ErrNotFound`
- preserves the underlying `pgconn.PgError` while also classifying lifecycle event uniqueness as `shared.ErrConflict`
## Surface area

- scanner `ScanResult.SLAs`, CLI JSON, and human CLI report
- policy/list/detail/history/event/transition HTTP APIs behind existing RBAC and engagement tenant chokepoints
- complete OpenAPI schemas and contract test
- engagement **Remediation SLA** tab with tier/deadline/overdue summaries and reviewed transitions
- default policy/operator rollout documentation

## Verification

- fresh PostgreSQL 18 cluster: migrations plus the complete `internal/infrastructure/persistence/postgres` package passed with `SYNAPSE_TEST_DB_DSN`
- targeted SLA/domain/memory/Postgres/SCA/vulnerability-evaluation/OpenAPI/use-case suites passed
- `go vet` on all changed Go packages passed
- `npm run typecheck` passed
- `npm test` — 39 files, 311 tests passed
- full `go test ./...` passed every package except the pre-existing race-sensitive `TestAdvisoryReconciliationPinsSnapshotRevision`; the same failure was reproduced on a clean `upstream/main` worktree

Closes #80
Closes #540

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -309,6 +309,15 @@ SYNAPSE_GOMODGRAPH_ENABLED=true      # transitive Go dep edges via `go mod graph
 # SYNAPSE_VULNERABILITY_TENANT_ALLOWLIST=       # tenant-a,tenant-b or *
 
 # =============================================================================
+# Risk-based remediation SLA governance – opt-in
+# =============================================================================
+# Creates versioned, reproducible remediation deadlines for persisted findings and reassesses them
+# when continuous vulnerability intelligence changes. Human lifecycle state and accepted-risk
+# decisions are never overwritten by scanner/AI refreshes. Reviewer permission is required for
+# lifecycle transitions; administrator permission is required to activate a tenant policy.
+# SYNAPSE_SLA_ENABLED=false
+
+# =============================================================================
 # DAST (authenticated crawling + first-party checks) – bounded, sandboxed
 # =============================================================================
 # Runs through the governed workflow (scope + HITL + kernel egress). Limits below
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,13 @@ capabilities below are already shipped on `main`.
 
 ### Added
 
+- **Risk-based remediation SLA governance.** Adds opt-in, tenant/RLS-safe and versioned SLA policy,
+  immutable deterministic assessment history, mitigate/remediate deadlines in scan/API/UI output,
+  human-only open/mitigating/remediated/accepted-risk transitions with hard acceptance expiry, and
+  continuous vulnerability-intelligence reassessment that preserves human state. Replays cannot move
+  an original deadline, machine principals cannot accept risk, and remediated findings cannot be
+  silently reopened; re-exposure remains explicit review work.
+
 - **Adversarial invariance gate for AI-triage promotion.** Golden datasets can bind a clean control to
   semantically equivalent prompt-injection challenges. Evaluation reports emit source-free pair evidence,
   and the comparison/release boundary requires complete coverage plus zero proposer, verifier, consensus,
--- a/api/openapi.yaml
+++ b/api/openapi.yaml
@@ -1966,7 +1966,7 @@ components:
 
     SLAAssessment:
       type: object
-      required: [tenant_id, id, engagement_id, finding_id, inputs, result, input_hash, config_hash, assessed_at, created_at]
+      required: [tenant_id, id, engagement_id, finding_id, inputs, result, input_hash, config_hash, deadline_anchor_at, assessed_at, created_at]
       properties:
         tenant_id: {type: string}
         id: {type: string}
@@ -1978,6 +1978,7 @@ components:
         input_hash: {type: string, pattern: '^[0-9a-f]{64}$'}
         config_hash: {type: string, pattern: '^[0-9a-f]{64}$'}
         previous_assessment_id: {type: string}
+        deadline_anchor_at: {type: string, format: date-time}
         assessed_at: {type: string, format: date-time}
         created_at: {type: string, format: date-time}
 
--- a/api/openapi.yaml
+++ b/api/openapi.yaml
@@ -40,6 +40,8 @@ tags:
   description: Tenant-scoped security operations summaries
 - name: vulnerability-intelligence
   description: Tenant-scoped vulnerability feeds, exposure correlation, risk transitions, and durable operations
+- name: remediation-sla
+  description: Risk-based remediation deadlines, immutable assessments, policy versions, and human-owned lifecycle governance
 
 paths:
   /healthz:
@@ -1593,6 +1595,142 @@ paths:
         '403': {$ref: '#/components/responses/Forbidden'}
         '404': {$ref: '#/components/responses/NotFound'}
 
+  /api/v1/sla/policies:
+    get:
+      tags: [remediation-sla]
+      operationId: listSLAPolicies
+      summary: List immutable SLA policy versions and the active version
+      description: Requires view permission. The built-in policy is installed on first use when no tenant policy exists.
+      responses:
+        '200':
+          description: Active policy and append-only tenant policy history
+          content:
+            application/json:
+              schema: {$ref: '#/components/schemas/SLAPolicyList'}
+        '401': {$ref: '#/components/responses/Unauthorized'}
+        '403': {$ref: '#/components/responses/Forbidden'}
+    post:
+      tags: [remediation-sla]
+      operationId: activateSLAPolicy
+      summary: Validate, append, and activate an SLA policy version
+      description: Requires administer permission and a human principal. Existing assessments are never rewritten.
+      requestBody:
+        required: true
+        content:
+          application/json:
+            schema:
+              type: object
+              required: [config]
+              properties:
+                config: {$ref: '#/components/schemas/SLAConfig'}
+      responses:
+        '200': {description: Existing identical policy selected, content: {application/json: {schema: {$ref: '#/components/schemas/SLAPolicyActivation'}}}}
+        '201': {description: New immutable policy created and selected, content: {application/json: {schema: {$ref: '#/components/schemas/SLAPolicyActivation'}}}}
+        '400': {$ref: '#/components/responses/BadRequest'}
+        '401': {$ref: '#/components/responses/Unauthorized'}
+        '403': {$ref: '#/components/responses/Forbidden'}
+
+  /api/v1/engagements/{id}/slas:
+    parameters: [{$ref: '#/components/parameters/EngagementId'}]
+    get:
+      tags: [remediation-sla, findings]
+      operationId: listEngagementSLAs
+      summary: List current remediation SLA views for an engagement
+      description: Time-sensitive overdue and acceptance-expiry fields are calculated at read time.
+      responses:
+        '200':
+          description: Current SLA views
+          content:
+            application/json:
+              schema:
+                type: object
+                required: [slas]
+                properties:
+                  slas: {type: array, items: {$ref: '#/components/schemas/SLAView'}}
+        '401': {$ref: '#/components/responses/Unauthorized'}
+        '403': {$ref: '#/components/responses/Forbidden'}
+        '404': {$ref: '#/components/responses/NotFound'}
+
+  /api/v1/engagements/{id}/slas/{fid}:
+    parameters:
+    - $ref: '#/components/parameters/EngagementId'
+    - $ref: '#/components/parameters/FindingId'
+    get:
+      tags: [remediation-sla, findings]
+      operationId: getFindingSLA
+      summary: Get the current SLA assessment and remediation lifecycle for a finding
+      responses:
+        '200': {description: Current SLA view, content: {application/json: {schema: {$ref: '#/components/schemas/SLAView'}}}}
+        '401': {$ref: '#/components/responses/Unauthorized'}
+        '403': {$ref: '#/components/responses/Forbidden'}
+        '404': {$ref: '#/components/responses/NotFound'}
+
+  /api/v1/engagements/{id}/slas/{fid}/assessments:
+    parameters:
+    - $ref: '#/components/parameters/EngagementId'
+    - $ref: '#/components/parameters/FindingId'
+    get:
+      tags: [remediation-sla, findings]
+      operationId: listSLAAssessmentHistory
+      summary: List immutable SLA assessments for a finding
+      responses:
+        '200':
+          description: Oldest-to-newest assessment history
+          content:
+            application/json:
+              schema:
+                type: object
+                required: [assessments]
+                properties:
+                  assessments: {type: array, items: {$ref: '#/components/schemas/SLAAssessment'}}
+        '401': {$ref: '#/components/responses/Unauthorized'}
+        '403': {$ref: '#/components/responses/Forbidden'}
+        '404': {$ref: '#/components/responses/NotFound'}
+
+  /api/v1/engagements/{id}/slas/{fid}/events:
+    parameters:
+    - $ref: '#/components/parameters/EngagementId'
+    - $ref: '#/components/parameters/FindingId'
+    get:
+      tags: [remediation-sla, findings]
+      operationId: listSLALifecycleEvents
+      summary: List append-only human remediation lifecycle events
+      responses:
+        '200':
+          description: Oldest-to-newest lifecycle audit history
+          content:
+            application/json:
+              schema:
+                type: object
+                required: [events]
+                properties:
+                  events: {type: array, items: {$ref: '#/components/schemas/SLALifecycleEvent'}}
+        '401': {$ref: '#/components/responses/Unauthorized'}
+        '403': {$ref: '#/components/responses/Forbidden'}
+        '404': {$ref: '#/components/responses/NotFound'}
+
+  /api/v1/engagements/{id}/slas/{fid}/transition:
+    parameters:
+    - $ref: '#/components/parameters/EngagementId'
+    - $ref: '#/components/parameters/FindingId'
+    post:
+      tags: [remediation-sla, findings]
+      operationId: transitionFindingSLA
+      summary: Apply an optimistic-concurrency protected human remediation decision
+      description: Machine principals are rejected. Accepted risk requires a reason, compensating control, and future expiry; remediated is terminal.
+      requestBody:
+        required: true
+        content:
+          application/json:
+            schema: {$ref: '#/components/schemas/SLATransitionRequest'}
+      responses:
+        '200': {description: Updated current SLA view, content: {application/json: {schema: {$ref: '#/components/schemas/SLAView'}}}}
+        '400': {$ref: '#/components/responses/BadRequest'}
+        '401': {$ref: '#/components/responses/Unauthorized'}
+        '403': {$ref: '#/components/responses/Forbidden'}
+        '404': {$ref: '#/components/responses/NotFound'}
+        '409': {description: Lifecycle version changed since it was loaded, content: {application/json: {schema: {$ref: '#/components/schemas/Error'}}}}
+
 components:
   securitySchemes:
     bearerAuth:
@@ -1605,6 +1743,11 @@ components:
       in: path
       required: true
       schema: {type: string}
+    FindingId:
+      name: fid
+      in: path
+      required: true
+      schema: {type: string, minLength: 1}
     BusinessAssetId:
       name: assetID
       in: path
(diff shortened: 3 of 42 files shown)


nghiadaulau (changes requested):
## 🔴 Changes requested — strong PR, but a real Postgres blocker + a HIGH governance bug + 2 MAJORs (do not merge yet)

This is genuinely good work — the domain is pure and deterministic, tenant RLS is thorough, machine/SoD is enforced, the optimistic-concurrency design is correct, RBAC + `withEngTenant` are wired right, and the domain lifecycle/assessment tests are excellent negative guards. But it is not mergeable as-is. Reviewed strictly with an architecture pass + a security pass; verified locally (CI disabled).

**Local verification:** `go build ./...`, `go vet ./...`, `gofmt -l` clean; all **non-Postgres** Go suites pass (domain/sla, slauc, sca, vulnerabilityevaluation, httpapi incl. hostile harness, memory, config, api/openapi); web `pnpm typecheck` + 311 tests + build pass; SLA tab is design-token clean. **The Postgres integration suite FAILS** on a real database (details below).

### ⛔ BLOCKER — every `UpsertAssessment` fails on Postgres (NUL byte in an advisory-lock key)
`internal/infrastructure/persistence/postgres/sla_store.go:133-134`
```go
lockKey := tenantID.String() + "\x00" + assessment.EngagementID.String() + "\x00" + assessment.FindingID.String()
tx.Exec(ctx, `SELECT pg_advisory_xact_lock(hashtextextended($1,0))`, lockKey)
```
A `text` bind parameter cannot contain `0x00`; Postgres rejects it: `lock sla finding: ERROR: invalid byte sequence for encoding "UTF8": 0x00 (SQLSTATE 22021)`. `TestPostgresSLAStore` t.Fatals at the first `UpsertAssessment` (sla_store_test.go:79) on a fresh DB. Since the production store is Postgres, **the entire SLA-assessment write path is broken in production**.
**Root cause of why this slipped through:** the PR was verified with `go test ./...` but the Postgres integration test *skips without `SYNAPSE_TEST_DB_DSN`* (Docker was unavailable locally, per the PR notes) — so it never ran. Please run it with a real DSN before re-requesting review.
**Fix:** don't put NUL in a text param. Hash the composite key in Go and pass a `bigint`: e.g. `pg_advisory_xact_lock($1::bigint)` where `$1` is a stable int64 (fnv64a over `tenant|eng|finding`), or use the two-int form. The lock *design* is right (per-finding serialization protecting the absent-pointer race) — only the key encoding is broken.

### 🟠 HIGH — remediation deadlines re-anchor to `now` on every material re-assessment (SLA clock can silently reset **and extend**)
`internal/domain/sla/sla.go:198-207` · `internal/domain/sla/assessment.go:128-135` · `internal/usecase/slauc/service.go:61-72`
`Compute` sets `MitigateBy/RemediateBy = now + range`, and the store only preserves the original deadline when the **input hash is byte-identical**. But the input hash is taken over the **raw** EPSS/CVSS floats while the tier/window is derived from **coarse bands** — so any non-material EPSS drift (EPSS updates daily, and both `slauc.InputsFromFinding` and `vulnerabilityevaluation.slaInputs` feed it in) mints a new assessment with a **fresh window anchored at `now`**, and the current pointer + `View.Overdue` advance to it. Worse, an *escalation* can **extend** a deadline: a 25-day-old High finding (RemediateBy = day 30) escalated to Emergency at day 25 by a continuous-intel refresh gets RemediateBy = day 25 + 7 = **day 32** — later than before, resetting an about-to-breach clock.
This contradicts the PR's own guarantee (`assessment.go:26-29`: "preserving its original due dates instead of moving them forward on every rescan") on the exact #540 continuous-refresh path this PR ships — an SLA-evasion / overdue-masking vector.
**Fix:** anchor the due dates to the **first assessment in the chain**. When the store binds `PreviousAssessmentID`, carry the original chain anchor (or `min(old, new)` deadline) rather than `now + range` — e.g. load `Current` before `Evaluate` and pass the prior anchor, or compute deadlines in the store where the previous row is known.

### 🟠 MAJOR — the SLA add-on hard-fails the whole scan *after findings are already committed*
`internal/usecase/sca/service.go:1977-2002` (call sites :1957, :2845)
`assessFindingSLAs` returns its error up through `runPipeline`, aborting the entire scan — but `findings.Upsert` already committed one statement earlier. The adjacent `attributeFindings` is deliberately best-effort (logs a warning, :2851-2853). So any transient SLA-store error reports the whole scan as failed while findings are persisted — and combined with the lock blocker, **every SLA-enabled scan on Postgres fails the entire pipeline**, not just SLA.
**Fix:** make SLA assessment best-effort/logged like `attributeFindings` (it's an opt-in governance add-on), or explicitly justify fail-closed and isolate it so it can't leave the scan committed-but-reported-failed.

### 🟠 MAJOR — duplicated risk→SLA mapping, and the *tested* copy is not the *live* one
`internal/usecase/vulnerabilityevaluation/service.go:131-147` vs `internal/usecase/slauc/service.go:242-260`
`slaInputs` re-implements `slauc.InputsFromRiskAssessment` almost verbatim, but the exported `InputsFromRiskAssessment` is referenced **only by its own test** — the production continuous-intelligence path uses the untested twin, and they've already diverged (`slauc` does `strings.TrimSpace(reason)`/`TrimSpace(FixedVersion)`, `slaInputs` doesn't). Two sources of truth for a security-relevant mapping, and the covered one isn't executed.
**Fix:** delete `slaInputs`; call the shared exported mapping so the tested code is the executed code.

### 🟡 MEDIUM — "immutable/append-only" SLA tables are not DB-enforced
`migrations/0103_sla_governance.sql`
`sla_assessments`, `sla_policies`, and the `sla_lifecycle_events` audit log are declared immutable/append-only in comments + PR text, but there is **no immutability trigger** — unlike the established convention (`fpe_append_only` on `finding_promotion_events` in 0082; evidence/audit_log in 0033; imported_findings in 0064). Golden rule 6 (append-only audit) is DB-enforced elsewhere; here it rests entirely on the stores choosing to only INSERT.
**Fix:** add `BEFORE UPDATE OR DELETE` triggers (mirroring `fpe_append_only`) on those three tables; leave the separate current-pointer tables (`sla_current_assessments`/`sla_active_policies`/`sla_lifecycles`) mutable.

### Minor
- `postgres/sla_store.go:283-291` — a unique violation (dup lifecycle-event id / stray policy) isn't mapped to `shared.ErrConflict` → surfaces as 500 not 409; the memory store returns `ErrConflict` (`memory/sla_store.go:235-239`). Detect `pgconn` unique-violation for parity.
- `internal/domain/shared/principal.go:8-18` — `IsMachineActor("")` returns `true`, flipping the AI-triage domain's empty-actor classification from human→machine (previously `aitriagereview.machineActor("")` returned `false`). Direction is fail-safe, but it's an unflagged out-of-scope change to a security predicate — please confirm no AI-triage caller relied on empty=human and note it in the description.
- `0103` — the `UNIQUE (tenant_id, sha256)` on `sla_policies` is redundant with the PK (Version is part of the digest); harmless, could drop.
- Once the lock is fixed, the whole Postgres store path (`loadSLAAssessment`, current-pointer upsert, `sla_lifecycles` no-clobber, `SaveTransition` optimistic UPDATE, history/event ordering, cross-tenant fail-closed) is currently **unverified** — it dies at the first call. Please re-run the full Postgres suite with a DSN; the assertions themselves are well designed.

### Summary
Must-fix before merge: the **BLOCKER** (NUL lock), the **HIGH** (deadline re-anchoring), and the two **MAJORs** (best-effort SLA in the pipeline; unify the risk→SLA mapping). The **MEDIUM** (append-only triggers) should land with them to match the repo's audit-integrity bar. Please also run the Postgres integration suite with `SYNAPSE_TEST_DB_DSN` so the store path is actually exercised. Ping me after and I'll re-review promptly — the foundation here is solid and I want it in.

XUanhoa04:
@nghiadaulau Thanks for the detailed review. I addressed the requested changes in `09555f8`:

1. PostgreSQL advisory locking now passes a stable SHA-256-derived `int64` to `pg_advisory_xact_lock($1)`; no NUL-containing text reaches PostgreSQL.
2. Every assessment chain now has an immutable `deadline_anchor_at`. A new assessment is rebased to the first anchor and capped by the previous due dates, so late escalation and de-escalation cannot reset/extend the clock. CVSS/EPSS drift inside the same effective scoring band is also non-material.
3. Post-persistence SCA SLA enrichment is best-effort/logged and retains successful partial results instead of failing the committed scan.
4. Continuous evaluation now uses the shared exported `slauc.InputsFromRiskAssessment` mapping; the duplicate mapper was removed.
5. PostgreSQL now rejects UPDATE/DELETE/TRUNCATE on SLA policies, assessments, and lifecycle events via append-only triggers.
6. Lifecycle-event `23505` errors map to `shared.ErrConflict`; the redundant `(tenant_id, sha256)` policy constraint was removed.
7. The empty/whitespace machine-actor behavior is now explicitly documented as intentional fail-closed behavior. Existing AI-triage decision/drift callers validate non-empty actors before classification.

Verification includes a brand-new PostgreSQL 18 cluster (all migrations + complete Postgres persistence package), targeted Go suites, changed-package `go vet`, frontend typecheck, and all 311 frontend tests. Regression tests cover the NUL-lock path, late emergency escalation, deadline non-extension, material fingerprints, append-only mutation attempts, partial scan enrichment, and unique-violation classification.

nghiadaulau (changes requested):
## 🔴 Round 2 — round-1 items all fixed correctly, but the fix introduced one new MAJOR regression (one more round, please)

Great turnaround on `09555f8f` — I re-reviewed with a fresh architecture + security pass and re-ran the suites (incl. the **real-Postgres SLA suite, now green**). **All five round-1 findings are correctly resolved:**

- ✅ **BLOCKER (NUL advisory-lock)** — `slaFindingAdvisoryLockKey` now hashes to an `int64` passed to `pg_advisory_xact_lock($1)`; no NUL crosses the wire, lock still scoped per (tenant, engagement, finding). Postgres suite passes.
- ✅ **HIGH (deadline re-anchoring)** — `DeadlineAnchorAt` + `ContinueAssessment` carry the chain-origin anchor and take `earlierDeadline(new, previous)`, so a deadline is now provably **monotonically non-increasing** per finding — escalation can't extend an about-to-breach clock, and a provenance-only refresh doesn't move it. Both stores call it identically. SLA-evasion vector closed.
- ✅ **MAJOR (scan hard-fail)** — `assessFindingSLAs` is now best-effort (`logger().Warn`, `errors.Join` of per-finding failures) at both call sites; `result.SLAs` still populated.
- ✅ **MAJOR (duplicate mapping)** — `slaInputs` deleted; production now calls the tested `slauc.InputsFromRiskAssessment`. Single source of truth.
- ✅ **MEDIUM (immutability)** — append-only + no-truncate triggers via the shared `synapse_forbid_mutation()` on all three immutable tables (+ `postgresUniqueViolation` → `ErrConflict` parity).

### 🟠 NEW MAJOR — the append-only DELETE trigger contradicts `ON DELETE CASCADE`, breaking finding/engagement/project deletion
`migrations/0103_sla_governance.sql:47` (+`:71`,`:150`) vs `:143-155`

`sla_assessments` (and `sla_lifecycle_events`) reference `findings(...) ON DELETE CASCADE`, but now also carry a `BEFORE DELETE` append-only trigger. In Postgres a cascade delete fires the child's row-level `BEFORE DELETE` trigger, which unconditionally `RAISE`s — so **deleting a finding (or its engagement/project) with SLA rows is aborted**. This regresses live paths: `EngagementRepository.Delete` (the partially-materialized-import rollback) and `projectuc.Service.Delete`. On `main` an SCA-scanned engagement is deletable; after this migration it fails with an opaque trigger error.

**Reproduced on a real DB** (fresh migrate to 0103):
```
DELETE FROM sla_policies WHERE tenant_id='t-x';
ERROR:  append-only: DELETE on sla_policies is forbidden (custody chain, golden rule 6)
CONTEXT: PL/pgSQL function synapse_forbid_mutation() line 3 at RAISE
```
`sla_assessments` has the identical trigger + a CASCADE FK, so a finding-cascade delete aborts the same way. The PR's own test is evidence of this: `postgres/sla_store_test.go:47-50` must `ALTER TABLE ... DISABLE TRIGGER` and null `previous_assessment_id` to clean up — a workaround production code doesn't have. (Separately, the self-referential `previous_assessment_id` FK has no `ON DELETE` action, so a multi-hop chain can't cascade-delete anyway.)

The sibling append-only table did this correctly: `finding_promotion_events` uses `ON DELETE RESTRICT` (0082), not CASCADE.

**Fix (pick one, consistently):**
- Change `sla_assessments`/`sla_lifecycle_events` → `findings` FKs to **`ON DELETE RESTRICT`** (matching `finding_promotion_events`), so a finding with SLA history is cleanly non-deletable with a proper FK error — then adjust the import-rollback / `projectuc.Delete` paths to purge SLA rows first (with an explicit, audited teardown) or to tolerate the RESTRICT; **or**
- drop the append-only **DELETE** triggers if cascade cleanup of SLA history on finding deletion is actually intended (keep UPDATE/TRUNCATE forbidden).

The CASCADE + forbid-DELETE combination cannot both hold. `0103` is new in this PR, so amend it in place.

### Minor (optional)
- `internal/infrastructure/persistence/memory/sla_store.go:120` — on a missing predecessor the memory store passes a zero-value `previous` into `ContinueAssessment` → wrapped `ErrValidation`, while Postgres returns `shared.ErrNotFound` (`postgres/sla_store.go:163-165`). Only reachable in an impossible state, but the sentinels differ across stores.
- `postgres/sla_store.go:305` — `fmt.Errorf("...: %v: %w", err, ErrConflict)` uses `%v` for the underlying `*pgconn.PgError`; `errors.Is(ErrConflict)` works but the pg error isn't recoverable via `errors.As`. Cosmetic.

### Verdict
Round-1 is fully and correctly addressed — genuinely strong. Blocking only on the new CASCADE-vs-append-only contradiction (verified on Postgres). Resolve that (and please re-run the Postgres suite **plus** an engagement/project delete of SLA-assessed data) and I'll merge. The two minors are polish.

XUanhoa04:
@nghiadaulau Thanks for the second review. Addressed the cascade/immutability contradiction in `fc4321d`.

I chose the aggregate-cascade option from the review:

- `sla_assessments` and `sla_lifecycle_events` now reject `UPDATE` and `TRUNCATE`, while `DELETE` is available to the intended finding/engagement/project cascade.
- Added `ON DELETE CASCADE` to the assessment predecessor self-FK and both lifecycle-to-assessment FKs, so multi-hop history is removed as one aggregate.
- Removed the assessment/event trigger bypasses from test cleanup.
- Added a real PostgreSQL integration path through `projectuc.Service.Delete`: a project containing an engagement, finding, lifecycle/event, and a three-assessment chain deletes cleanly, with all related rows verified gone and the project deletion audit verified.

I also fixed both minor findings:

- the memory adapter now returns `shared.ErrNotFound` for a dangling current-assessment predecessor, matching PostgreSQL;
- lifecycle-event unique errors retain the original `*pgconn.PgError` through `%w` while still matching `shared.ErrConflict`.

Verification on a freshly recreated PostgreSQL 18 database:

- `go test ./internal/infrastructure/persistence/postgres -count=1`
- `go test ./internal/infrastructure/persistence/memory ./internal/usecase/projectuc ./internal/usecase/slauc ./internal/domain/sla -count=1`
- `go vet ./internal/infrastructure/persistence/memory ./internal/infrastructure/persistence/postgres ./internal/usecase/projectuc`
- `git diff --check`

All passed. Could you please re-review this revision?

nghiadaulau (approved):
## ✅ Approved — merging (round 3; all blockers across 2 rounds resolved and verified)

`fc4321d9b` "allow audited aggregate teardown" resolves the last blocker correctly. Verified locally (CI disabled): `go build ./...`, `go vet`, `gofmt -l` clean; all Go suites (domain/sla, slauc, sca, vulnerabilityevaluation, httpapi incl. hostile harness, memory, api/openapi) green; **real-Postgres SLA suite green including the new teardown test**; web typecheck + 311 tests + build green.

**Round-2 MAJOR (CASCADE FK vs append-only DELETE trigger) — FIXED, correct:**
- The append-only triggers on `sla_assessments` / `sla_lifecycle_events` now fire on **UPDATE only** — tamper-immutability (the custody-chain guarantee) is preserved — while DELETE is permitted so the `ON DELETE CASCADE` teardown from a finding/engagement/project deletion works. `sla_policies` stays fully append-only (UPDATE **and** DELETE blocked — policies outlive findings). The self-ref + lifecycle/event → assessment FKs got `ON DELETE CASCADE` so a multi-hop chain tears down cleanly.
- Proven by the new integration test: it drives the **real** `projectuc.Service.Delete` → `EngagementRepository.Delete` on an SLA-assessed project, asserts every SLA table is empty after teardown **and** that the teardown emits a single audited `project.delete` entry, and separately asserts a direct `UPDATE sla_lifecycle_events SET actor='tampered'` is still rejected. That's exactly the CASCADE-works + tamper-blocked pair I asked for.

**Round-1 (all 5) — confirmed fixed in round 2:** int64 advisory-lock (no NUL); monotonic non-increasing deadlines via `DeadlineAnchorAt` + `earlierDeadline` (SLA-evasion closed); best-effort `assessFindingSLAs`; single tested risk→SLA mapping; DB-enforced immutability triggers.

**One non-blocking note for a future follow-up (not gating):** unlike `finding_promotion_events` (`ON DELETE RESTRICT`), `sla_assessments`/`sla_lifecycle_events` now permit *direct* DELETE at the DB layer (only UPDATE is trigger-blocked). In practice the app only ever INSERTs or deletes via the audited cascade teardown, and tamper-by-UPDATE is blocked — so the custody guarantee holds — but if you later want DB-level protection against a direct out-of-band `DELETE FROM sla_assessments`, a teardown-scoped guard (e.g. a session GUC set by the repo during cascade) would close that gap. Documented trade-off; fine to ship.

Excellent, disciplined turnaround across three rounds on a large, safety-critical surface. Merging — this completes #80's operational phases.

PR #577, opened 2026-08-16
fix: ten integration-boundary defects found by end-to-end validation

### LINKED ISSUE #576: Ten integration-boundary defects found by end-to-end validation (RLS, agent plane, sandbox, AWS CSPM)
## Summary

A full end-to-end validation run against real dependencies — PostgreSQL with RLS, MinIO, pinned
Syft/Grype, bubblewrap on Linux, a local kind cluster, a live LLM gateway, and a real AWS
organization — surfaced **ten product defects**. None of them are reproducible from unit tests
alone: every one required exercising a real boundary.

They fall into three groups, and the grouping is the interesting part.

### A. Integration-boundary failures the test doubles hid (4)

In each case the fake was **more permissive than the real boundary**, so the defect passed unit
tests while failing against PostgreSQL, the router, or the asset visibility gate.

1. **Project analysis drops the tenant on the completion path.** `runScanJob` receives a
   tenant-carrying context, but the project-analysis completion path rebuilt from a bare
   `context.Background()`. The recorder reads the engagement through a tenant-scoped (RLS)
   repository, so every Project analysis failed with
   `tenant context is required` — *after the entire pipeline had already succeeded*.
2. **The fleet agent plane never binds the authenticated agent's tenant onto the context.** The
   hash-chained audit log reads the tenant from the context to satisfy RLS, and this plane has no
   human principal to carry it. Every enrolment, heartbeat, claim, result, and inventory write
   failed with `new row violates row-level security policy for table audit_log`. Enrolment also
   consumed its single-use token *before* the audit insert failed, leaving an orphan `active`
   agent row.
3. **Mounting the agent plane on the whole `/api/v1/fleet/` prefix shadows the operator routes.**
   The prefix mount swallowed the operator coverage and agent-health endpoints that live under the
   same prefix but are served by the human RBAC chain — every one of them answered `404` instead
   of reaching the human handler.
4. **CSPM attributes findings before reconciling observations.** A cloud asset carries a
   `scope_key`, and `ListAssets` deliberately hides such an asset until an *active*
   `cspm_observations` row vouches for it (no default-to-clean inventory). Attribution validates
   its asset through that same listing, so every run carrying at least one asset failed
   `asset <id>: not found`, retried three times, and dead-lettered with zero assets recorded.

### B. Failures that only appear against a real provider or a real sandbox (4)

5. **The LLM client omits `stream`.** This client decodes one non-streaming JSON body, but
   OpenAI-compatible gateways default to `text/event-stream` when the field is absent. Every call
   failed with `invalid character 'd' looking for beginning of value` (the `d` of the SSE `data:`
   prefix), halting agent sessions and AI triage entirely.
6. **Evidence-backed FP triage sends `uniqueItems`.** OpenAI structured outputs reject that
   keyword in strict mode. **100% of evidence-backed triage calls** failed with
   `400 invalid_json_schema`, which the telemetry then bucketed as a *provider* error — so a
   client-side schema bug looked like a flaky gateway.
7. **The sandbox never binds the tool binary itself.** The curated read-only root deliberately
   omits `/opt` so host secrets stay ENOENT, but nothing bound the resolved, integrity-verified
   binary. The result: the **documented** `/opt/synapse/bin/synapse-cspm` layout (see
   `docs/guide/cloud-posture.md`) could not run at all —
   `bwrap: execvp /opt/synapse/bin/synapse-cspm: No such file or directory`.
8. **`ListAccounts` sends the total bound as the page size.** `MaxAccounts` is our total cap
   (default 100), but AWS Organizations rejects a `MaxResults` above 20 with
   `InvalidInputException` rather than clamping it. Account enumeration therefore failed on
   **every real organization**, surfacing as an opaque `provider_error` coverage gap after
   `DescribeOrganization` had already succeeded.

### C. A silent false negative in posture assessment (1)

9. **Absent S3 configuration is treated as an observation failure.** `NoSuchBucketPolicy` and
   `ServerSideEncryptionConfigurationNotFoundError` are *definitive negatives* — "there is no
   policy", "it is not encrypted". Recording them as coverage gaps did two harmful things: it
   invented a `provider_error` on healthy accounts, **and** it left `Public`/`Encrypted` as
   `unknown` instead of `disabled`, so the posture rules never raised the unencrypted-bucket
   finding. For a security product this is the worst failure mode: a suppressed finding rather
   than a visible error.

### D. Observability gap that blocked all of the above (1)

10. **The CSPM executor discards the helper's stderr.** Every failure surfaced as a bare
    `exit_code=1 timed_out=false truncated=false`, making an operator-fixable misconfiguration
    (unassumable role, denied API, blocked egress) indistinguishable from a crash. Defects 7, 8,
    and 9 were only diagnosable *after* this was fixed — it is listed last but was fixed first.

## Impact

| Defect | Blast radius |
| --- | --- |
| 1 | Every Project analysis fails on PostgreSQL |
| 2 | The entire fleet agent plane is unusable on PostgreSQL |
| 3 | All fleet operator coverage/health endpoints return 404 |
| 4 | Every CSPM run that finds any asset dead-letters |
| 5 | All agent sessions and AI triage fail against a real gateway |
| 6 | 100% of evidence-backed triage calls fail, mislabeled as provider faults |
| 7 | The documented CSPM helper install layout cannot execute |
| 8 | AWS account enumeration fails on every real organization |
| 9 | **Unencrypted / public buckets silently produce no finding** |
| 10 | All CSPM failures are undiagnosable |

## Reproduction

Defects 1–4 need PostgreSQL with RLS enabled (`SYNAPSE_DB_DSN` pointing at a
`NOSUPERUSER NOBYPASSRLS` runtime role); the in-memory persistence path does not exercise them.
Defects 5–6 need a live OpenAI-compatible gateway. Defects 7–10 need `synapse-worker` on Linux
with bubblewrap plus kernel egress enforcement and real AWS credentials.

## Proposed fix

Fix all ten in place, each with a regression test that models the **real** boundary rather than the
permissive fake, and verify each test fails without its fix. PR to follow.

pho-veteran:
Closes #576

## Summary

An end-to-end validation run against real dependencies — PostgreSQL with RLS, MinIO, pinned
Syft/Grype, bubblewrap on Linux, a local kind cluster, a live LLM gateway, and a real AWS
organization — surfaced **ten product defects**. This fixes all ten.

None were reproducible from unit tests alone, and the reason matters: in four cases the **test
double was more permissive than the real boundary**. Context-blind fakes and an attributor stub that
accepted every asset let those defects pass the suite while failing against PostgreSQL RLS, the
router, and the asset visibility gate. Each new test models the real gate rather than the forgiving
fake.

Severity is uneven. Most of these produce a loud failure, but **defect 9 suppresses a finding** — an
unencrypted or publicly-policied S3 bucket silently produced no posture finding at all. For a
security product that is the worst failure mode, and it is the one I would weight highest in review.

## Changes

Eight focused commits, one per defect (the two AWS connector fixes share a commit because they sit
on the same enumeration path).

**Tenant / RLS boundary**

- `fix(sca)` — `runScanJob` carries a tenant, but the project-analysis completion path rebuilt from
  a bare `context.Background()`. The recorder reads the engagement through a tenant-scoped (RLS)
  repository, so **every Project analysis** failed `tenant context is required` *after the whole
  pipeline had already succeeded*. Now derived with `context.WithoutCancel(ctx)` — detached from
  request cancellation, tenant preserved.
- `fix(fleet)` — the untrusted agent plane has no human principal to carry a tenant, but the
  hash-chained audit log reads it from the context to satisfy RLS. Every enrolment, heartbeat,
  claim, result, and inventory write failed with `new row violates row-level security policy for
  table audit_log`. Bound from the enrolment token in `Enrol` and from the authenticated agent in
  the transport authenticator — **never from a request field**.

**Auth-plane routing**

- `fix(httpapi)` — mounting the agent transport on the whole `/api/v1/fleet/` prefix also swallowed
  the **operator** routes under it (fleet coverage, agent health), which the agent mux does not
  serve, so each returned `404` instead of reaching the human RBAC chain. Now mounted on the exact
  agent-plane subtrees; the new test guards the split in both directions so a future agent route
  cannot silently capture an operator path.

**LLM boundary**

- `fix(llm)` — the client decodes one non-streaming JSON body, but OpenAI-compatible gateways
  default to `text/event-stream` when `stream` is absent. Every call failed on the `d` of the SSE
  `data:` prefix, halting agent sessions and AI triage. `stream` is now always sent (deliberately no
  `omitempty`).
- `fix(fptriage)` — strict structured outputs reject `uniqueItems`, so **100% of evidence-backed
  triage calls** returned `400 invalid_json_schema`. Worse, `classifyCallError` buckets that
  client-side error as `provider_error`, so a malformed request of ours looked like a flaky gateway.
  Uniqueness was already enforced in code by `validateEvidenceCitations`, so the keyword was
  redundant as well as invalid.

**Sandbox**

- `fix(cloudsandbox)` — the executor discarded the helper's stderr, so every failure was a bare
  `exit_code=1`. Fixed first, because defects 7–9 were only diagnosable once it was. stderr is
  untrusted output on a credentialed path, so the reason is redacted against the credential,
  bounded, collapsed to one line, and **dropped entirely if any placeholder survives**.
- `fix(sandbox)` — `Run()` resolves and integrity-verifies an absolute binary path, but the curated
  root omits `/opt` (so host secrets stay ENOENT) and nothing bound the verified binary. The
  **documented** `/opt/synapse/bin/synapse-cspm` layout in `docs/guide/cloud-posture.md` could not
  execute: `bwrap: execvp ...: No such file or directory`. Binds the single verified **file**
  read-only — never its directory, so the curated root is not widened.

**AWS CSPM**

- `fix(cloud/aws)` — (a) `MaxAccounts` is our *total* bound, but it was passed as `ListAccounts`
  `MaxResults`; Organizations caps that at 20 and **rejects** larger values, so account enumeration
  failed on every real organization. (b) `NoSuchBucketPolicy` and
  `ServerSideEncryptionConfigurationNotFoundError` are *definitive negatives*, not failures to
  observe — treating them as coverage gaps invented a `provider_error` on healthy accounts **and**
  left `Public`/`Encrypted` unknown, suppressing the finding. Named absence codes now map to
  `StateDisabled`; genuine permission and transport failures still raise gaps.
- `fix(cspm)` — attribution ran before observation reconciliation, but `ListAssets` hides a
  `scope_key`-bearing asset until an *active* `cspm_observations` row vouches for it. Attribution
  validates through that same listing, so every run with assets failed `asset <id>: not found`,
  retried three times, and dead-lettered. Reconciliation depends only on already-persisted findings
  and edges, so it now runs first.

## Verification

Each fix has a regression test, and **each test was confirmed to fail without its fix** — several
reproduce the exact production error string (`asset id-2: not found`,
`Encrypted = "unknown", want "disabled"`, `MaxResults = 100, want <= 20`).

Every failing flow was rerun against the real dependency until it passed:

| Flow | Before | After |
| --- | --- | --- |
| FP triage | 8 requests, **8 provider failures**, 0 tokens, 0 findings | 8 requests, **0 failures**, 12,776 tokens, 4 findings |
| Agent session | 0 steps (SSE decode failure) | 16 steps, 129,680 tokens, 18 judgments — **all `proposed`** |
| AWS CSPM | dead-lettered, 0 assets | `partial`, **7 real assets**, evidence sealed, honest coverage gaps |
| Project analysis | `tenant context is required` | succeeds, 20 issues / 7 hotspots persisted |
| Fleet agent plane | 500 on enrolment | enrol → claim → result → retry lifecycle green |

Live AWS CSPM now returns `findings=0` as a **true negative** — every enumerated bucket really is
encrypted and non-public, read from the real API rather than defaulted.

The AWS work used a temporary read-only IAM role tagged `managed-by=synapse-e2e`, deleted at
teardown and verified absent (`NoSuchEntity`). Total incremental AWS cost: **$0.00**.

## Scope

Code only — no artifacts, docs, or `.gitignore` changes. `docs/guide/cloud-posture.md` needs no
edit: the sandbox fix makes the layout it already documents actually work.

## Checklist

- [x] `make build vet test typecheck` passes — **245 packages, 0 failures**
- [x] `cd web && pnpm build` passes (if the dashboard changed) — no dashboard changes; `pnpm typecheck` run via `make typecheck`
- [x] Tests added/updated for new behavior — 13 regression tests, each verified to fail without its fix
- [x] Preserves the safety invariants (argv exec, server-side scope enforcement, secrets out of
      logs, deterministic reports, append-only evidence) — see notes below
- [x] Documentation updated if needed — no change required
- [x] `make lint` (golangci-lint v2) — **0 issues**
- [x] `TestHostileHarness` passes (security- and execution-sensitive changes)

### Safety-invariant notes for review

These changes touch the sandbox, the credential path, and an auth plane, so the relevant invariants
are called out explicitly:

- **argv exec** — unchanged. The sandbox fix adds a bwrap `--ro-bind-try` flag to an argv array; no
  shell string is constructed anywhere.
- **Server-side scope enforcement** — unchanged and re-verified. Out-of-scope cloud roots, expired
  authorization windows, RoE tool-class denial, and non-read-only cloud operations all still refuse
  before any provider request, and each denial is audited.
- **Secrets out of logs** — *strengthened*. The new stderr diagnostic is the only path added, and it
  redacts against the credential, bounds the output, and refuses to emit at all if a redaction
  placeholder survives. Two dedicated tests cover the leak case.
- **Sandbox confinement** — narrowed, not widened. Only the single integrity-verified binary file is
  bound, never its parent directory; a test asserts `/opt/synapse/bin` and `/opt` are *not* bound
  and that lookalike paths (`/libexec/...`, `/lib-evil/...`) are not mistaken for the curated root.
- **Append-only evidence / deterministic reports** — untouched.
- **Tenant RLS** — *strengthened*. Two of these fixes exist specifically to make RLS-scoped writes
  work correctly rather than to bypass them; the tenant is always taken from a token or an
  authenticated principal, never from client input.

### Not covered here

Azure and GCP connectors were not exercised (no credentials in that environment). AWS did exercise
the shared CSPM machinery — defects 7–10 were all found on that shared path — but the
provider-specific Azure/GCP code remains unvalidated end to end.

### DIFF
--- a/internal/infrastructure/cloud/aws/aws.go
+++ b/internal/infrastructure/cloud/aws/aws.go
@@ -35,6 +35,9 @@ const (
 	defaultMaxAccounts  = 100
 	defaultMaxResources = 1000
 	defaultRate         = 5
+	// maxListAccountsPageSize is the largest MaxResults AWS Organizations accepts for
+	// ListAccounts. Anything higher is rejected with InvalidInputException rather than clamped.
+	maxListAccountsPageSize = 20
 )
 
 var _ ports.CloudConnector = (*Connector)(nil)
@@ -200,7 +203,12 @@ type account struct {
 }
 
 func (c *Connector) accounts(ctx context.Context, client *organizations.Client, root string) ([]account, *cloudposture.CoverageIssue) {
-	pager := organizations.NewListAccountsPaginator(client, &organizations.ListAccountsInput{MaxResults: awssdk.Int32(int32(c.opts.MaxAccounts))})
+	// MaxAccounts is OUR total bound, not a page size. Organizations rejects a ListAccounts
+	// MaxResults above maxListAccountsPageSize with InvalidInputException, so sending the total
+	// bound (default 100) failed account enumeration outright and surfaced as a
+	// "provider_error" coverage gap on every real organization. Page within the API limit and
+	// keep enforcing the total bound while collecting.
+	pager := organizations.NewListAccountsPaginator(client, &organizations.ListAccountsInput{MaxResults: awssdk.Int32(int32(min(c.opts.MaxAccounts, maxListAccountsPageSize)))})
 	var accounts []account
 	for pager.HasMorePages() {
 		page, err := pager.NextPage(ctx)
@@ -298,20 +306,37 @@ func (c *Connector) buckets(ctx context.Context, client *s3.Client, account acco
 				*gaps = append(*gaps, *coverageGap(account.id, "storage", err))
 			} else {
 				status, statusErr := client.GetBucketPolicyStatus(ctx, &s3.GetBucketPolicyStatusInput{Bucket: awssdk.String(name)})
-				if statusErr != nil {
+				switch {
+				case statusErr != nil && isAbsentConfig(statusErr, "NoSuchBucketPolicy"):
+					// No bucket policy is a DEFINITIVE answer, not a failure to observe: the bucket
+					// is not public by policy. Treating it as a coverage gap both invented a
+					// provider_error and left Public unknown, so an evidence-backed negative was
+					// downgraded to "we could not tell".
+					resource.Public = cloudposture.StateDisabled
+				case statusErr != nil:
 					*gaps = append(*gaps, *coverageGap(account.id, "storage", statusErr))
-				} else if status.PolicyStatus != nil && awssdk.ToBool(status.PolicyStatus.IsPublic) {
+				case status.PolicyStatus != nil && awssdk.ToBool(status.PolicyStatus.IsPublic):
 					resource.Public = cloudposture.StateEnabled
+				default:
+					resource.Public = cloudposture.StateDisabled
 				}
 			}
 			if err := limiter.wait(ctx); err != nil {
 				*gaps = append(*gaps, *coverageGap(account.id, "storage", err))
 			} else {
 				encryption, encryptionErr := client.GetBucketEncryption(ctx, &s3.GetBucketEncryptionInput{Bucket: awssdk.String(name)})
-				if encryptionErr != nil {
+				switch {
+				case encryptionErr != nil && isAbsentConfig(encryptionErr, "ServerSideEncryptionConfigurationNotFoundError"):
+					// Same contract: "no encryption configuration" is an OBSERVED negative. It
+					// must land as Disabled so the posture rules can raise an unencrypted-bucket
+					// finding; a coverage gap here silently suppressed that finding.
+					resource.Encrypted = cloudposture.StateDisabled
+				case encryptionErr != nil:
 					*gaps = append(*gaps, *coverageGap(account.id, "storage", encryptionErr))
-				} else if encryption.ServerSideEncryptionConfiguration != nil {
+				case encryption.ServerSideEncryptionConfiguration != nil:
 					resource.Encrypted = cloudposture.StateEnabled
+				default:
+					resource.Encrypted = cloudposture.StateDisabled
 				}
 			}
 			inventory.Resources = append(inventory.Resources, resource)
@@ -410,6 +435,16 @@ func appendUnsupportedCoverage(inventory *cloudposture.Inventory, gaps *[]cloudp
 		*gaps = append(*gaps, cloudposture.CoverageIssue{Provider: cloudposture.ProviderAWS, Scope: scope, Category: category, Code: "unsupported", Detail: "connector does not yet establish this posture category"})
 	}
 	inventory.Complete = false
+}
+
+// isAbsentConfig reports whether err is the provider's way of saying "this configuration does not
+// exist" rather than "the request failed". Those codes are DEFINITIVE observations - no bucket
+// policy, no encryption configuration - so they must set a negative state instead of raising a
+// coverage gap: recording them as failures both invented a provider_error and suppressed the
+// posture finding the absent configuration should produce.
+func isAbsentConfig(err error, code string) bool {
+	var apiErr smithy.APIError
+	return errors.As(err, &apiErr) && apiErr.ErrorCode() == code
 }
 
 func coverageGap(scope, category string, err error) *cloudposture.CoverageIssue {
--- a/internal/infrastructure/cloud/aws/aws.go
+++ b/internal/infrastructure/cloud/aws/aws.go
@@ -315,9 +315,15 @@ func (c *Connector) buckets(ctx context.Context, client *s3.Client, account acco
 					resource.Public = cloudposture.StateDisabled
 				case statusErr != nil:
 					*gaps = append(*gaps, *coverageGap(account.id, "storage", statusErr))
-				case status.PolicyStatus != nil && awssdk.ToBool(status.PolicyStatus.IsPublic):
+				case status.PolicyStatus == nil:
+					// A 200 carrying no policy-status field is NOT evidence of absence, so the state
+					// stays unknown. Collapsing it into the negative below would manufacture an
+					// evidence-backed claim out of a missing field - the same category error this
+					// function fixes in the other direction.
+				case awssdk.ToBool(status.PolicyStatus.IsPublic):
 					resource.Public = cloudposture.StateEnabled
 				default:
+					// PolicyStatus present and IsPublic false: an observed negative.
 					resource.Public = cloudposture.StateDisabled
 				}
 			}
@@ -336,7 +342,9 @@ func (c *Connector) buckets(ctx context.Context, client *s3.Client, account acco
 				case encryption.ServerSideEncryptionConfiguration != nil:
 					resource.Encrypted = cloudposture.StateEnabled
 				default:
-					resource.Encrypted = cloudposture.StateDisabled
+					// A 200 carrying no encryption configuration field is NOT evidence of absence:
+					// only the named error code above is. Leave the state unknown rather than
+					// manufacturing an unencrypted-bucket claim from a missing field.
 				}
 			}
 			inventory.Resources = append(inventory.Resources, resource)
--- a/internal/infrastructure/cloudsandbox/executor.go
+++ b/internal/infrastructure/cloudsandbox/executor.go
@@ -40,6 +40,32 @@ func New(runner ports.ToolRunner, vault ports.CredentialVault, binary string, ra
 	return &Executor{runner: runner, vault: vault, binary: binary, rate: rate, timeout: timeout, maxOutput: maxOutput, egressHost: egressHosts}, nil
 }
 
+// diagnosticCap bounds how much helper stderr reaches an error message. The helper writes one
+// short reason line; anything longer is a malfunction and must not become an unbounded log write.
+const diagnosticCap = 512
+
+// diagnostic renders the helper's stderr as a bounded, credential-free error suffix. Without it a
+// failing helper surfaces only "exit_code=1", which says nothing about WHY posture enumeration
+// failed and makes an operator-fixable misconfiguration (bad role ARN, denied API, blocked egress)
+// indistinguishable from a crash. The credential is redacted first and, if any placeholder survives,
+// the text is dropped entirely rather than risking secret material in a log or API error.
+func diagnostic(stderr, secret []byte) string {
+	redacted := redact.Bytes(stderr, [][]byte{secret})
+	if strings.Contains(string(redacted), redact.Placeholder) {
+		return " reason=<redacted>"
+	}
+	reason := strings.TrimSpace(string(redacted))
+	if reason == "" {
+		return ""
+	}
+	if len(reason) > diagnosticCap {
+		reason = reason[:diagnosticCap] + "…"
+	}
+	// Collapse to a single line: these reasons land in structured logs and JSON error bodies.
+	reason = strings.Join(strings.Fields(reason), " ")
+	return " reason=" + reason
+}
+
 func (e *Executor) EnumerateCloud(ctx context.Context, scope ports.CloudScope) (cloudposture.Inventory, []cloudposture.CoverageIssue, error) {
 	if scope.Authorize == nil {
 		return cloudposture.Inventory{}, nil, fmt.Errorf("%w: cloud operation authorizer is required", shared.ErrForbidden)
@@ -142,7 +168,7 @@ func (e *Executor) EnumerateCloud(ctx context.Context, scope ports.CloudScope) (
 		return cloudposture.Inventory{}, nil, fmt.Errorf("sandboxed CSPM helper failed: %w", runErr)
 	}
 	if result.ExitCode != 0 || result.TimedOut || result.Truncated {
-		return cloudposture.Inventory{}, nil, fmt.Errorf("sandboxed CSPM helper failed: exit_code=%d timed_out=%t truncated=%t", result.ExitCode, result.TimedOut, result.Truncated)
+		return cloudposture.Inventory{}, nil, fmt.Errorf("sandboxed CSPM helper failed: exit_code=%d timed_out=%t truncated=%t%s", result.ExitCode, result.TimedOut, result.Truncated, diagnostic(result.Stderr, secret))
 	}
 	result.Stdout = redact.Bytes(result.Stdout, [][]byte{secret})
 	if strings.Contains(string(result.Stdout), redact.Placeholder) {
--- a/internal/infrastructure/cloudsandbox/executor.go
+++ b/internal/infrastructure/cloudsandbox/executor.go
@@ -12,6 +12,7 @@ import (
 	"strings"
 	"sync"
 	"time"
+	"unicode/utf8"
 
 	"github.com/KKloudTarus/synapse-ce/internal/domain/cloudposture"
 	"github.com/KKloudTarus/synapse-ce/internal/domain/shared"
@@ -44,13 +45,58 @@ func New(runner ports.ToolRunner, vault ports.CredentialVault, binary string, ra
 // short reason line; anything longer is a malfunction and must not become an unbounded log write.
 const diagnosticCap = 512
 
+// minScrubbableSecretLen is the shortest credential-field value scrubbed on its own. Provider
+// credential documents also carry SHORT structural values (a region, a provider name, "true"), and
+// scrubbing those would match ordinary prose in any diagnostic and drop every reason - silently
+// reverting the observability this file exists to provide. Real credential material is far longer:
+// an AWS access key id is 20 characters, a secret access key 40, session tokens and GCP/Azure keys
+// longer still.
+const minScrubbableSecretLen = 16
+
+// credentialSecrets decomposes resolved credential material into every value that must be scrubbed
+// INDEPENDENTLY. A provider credential is a JSON document holding several distinct secret-bearing
+// fields (an AWS document carries access_key_id, secret_access_key and session_token), and provider
+// SDK errors routinely echo exactly ONE of them in isolation - for example "The AWS Access Key Id
+// you provided does not exist in our records: AKIA...". Scrubbing only the whole document never
+// matches that substring, so the field would survive redaction, pass the placeholder check, and
+// reach a structured log or a JSON API error body.
+//
+// The document is treated as opaque provider-shaped JSON on purpose: this executor is
+// provider-agnostic and must not import a concrete connector's credential type.
+func credentialSecrets(secret []byte) [][]byte {
+	secrets := [][]byte{secret}
+	var document any
+	if err := json.Unmarshal(secret, &document); err != nil {
+		return secrets // opaque (non-JSON) credential: the whole blob is the only known value
+	}
+	var walk func(node any)
+	walk = func(node any) {
+		switch typed := node.(type) {
+		case map[string]any:
+			for _, value := range typed {
+				walk(value)
+			}
+		case []any:
+			for _, value := range typed {
+				walk(value)
+			}
+		case string:
+			if len(typed) >= minScrubbableSecretLen {
+				secrets = append(secrets, []byte(typed))
+			}
+		}
+	}
+	walk(document)
+	return secrets
+}
+
 // diagnostic renders the helper's stderr as a bounded, credential-free error suffix. Without it a
 // failing helper surfaces only "exit_code=1", which says nothing about WHY posture enumeration
 // failed and makes an operator-fixable misconfiguration (bad role ARN, denied API, blocked egress)
-// indistinguishable from a crash. The credential is redacted first and, if any placeholder survives,
-// the text is dropped entirely rather than risking secret material in a log or API error.
-func diagnostic(stderr, secret []byte) string {
-	redacted := redact.Bytes(stderr, [][]byte{secret})
+// indistinguishable from a crash. Every credential value is redacted first and, if any placeholder
+// survives, the text is dropped entirely rather than risking secret material in a log or API error.
+func diagnostic(stderr []byte, secrets [][]byte) string {
+	redacted := redact.Bytes(stderr, secrets)
 	if strings.Contains(string(redacted), redact.Placeholder) {
 		return " reason=<redacted>"
 	}
@@ -58,8 +104,14 @@ func diagnostic(stderr, secret []byte) string {
 	if reason == "" {
 		return ""
 	}
+	// Truncate on a RUNE boundary: a byte-offset cut can split a multi-byte rune and emit invalid
+	// UTF-8 into a structured log and a JSON error body, which strings.Fields below cannot repair.
 	if len(reason) > diagnosticCap {
-		reason = reason[:diagnosticCap] + "…"
+		cut := diagnosticCap
+		for cut > 0 && !utf8.ValidString(reason[:cut]) {
+			cut--
+		}
+		reason = reason[:cut] + "…"
 	}
 	// Collapse to a single line: these reasons land in structured logs and JSON error bodies.
 	reason = strings.Join(strings.Fields(reason), " ")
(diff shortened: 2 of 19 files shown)


internal/infrastructure/cloudsandbox/executor.go
```diff
@@ -40,6 +40,32 @@ func New(runner ports.ToolRunner, vault ports.CredentialVault, binary string, ra
...
+// short reason line; anything longer is a malfunction and must not become an unbounded log write.
+const diagnosticCap = 512
+
+// diagnostic renders the helper's stderr as a bounded, credential-free error suffix. Without it a
+// failing helper surfaces only "exit_code=1", which says nothing about WHY posture enumeration
+// failed and makes an operator-fixable misconfiguration (bad role ARN, denied API, blocked egress)
+// indistinguishable from a crash. The credential is redacted first and, if any placeholder survives,
+// the text is dropped entirely rather than risking secret material in a log or API error.
+func diagnostic(stderr, secret []byte) string {
+	redacted := redact.Bytes(stderr, [][]byte{secret})
```
pho-veteran: **Major — new partial-secret leak on a credentialed path.**

`redact.Bytes` is documented as exact-substring matching, and it is handed a single-element secret set: `[][]byte{secret}`, where `secret` is the whole JSON credential blob returned by `vault.Resolve` (executor.go:88). It therefore only catches the blob verbatim.

The AWS credential carries four independent secret-bearing fields (`AccessKeyID`, `SecretAccessKey`, `SessionToken`, `RoleARNTemplate` — aws.go:166-171), and AWS SDK errors routinely echo exactly one of them in isolation, e.g. `The AWS Access Key Id you provided does not exist in our records: AKIA...`. That substring does not match the blob, so it survives redaction, passes the `redact.Placeholder` check at line 54 untouched, and lands in the error returned at executor.go:171 — which by this function's own doc comment reaches structured logs and JSON API error bodies.

This is introduced here, not inherited: before this PR stderr was discarded entirely. It also means the PR description's claim that secrets-out-of-logs is *strengthened* does not hold as written. The bounding, one-line collapse, and placeholder short-circuit are all good; the secret *set* is the problem.

Suggested fix: unmarshal the credential and pass each secret-bearing field as its own element, so a single leaked field is caught:

```go
redacted := redact.Bytes(stderr, secretForms(secret))
```

where `secretForms` returns the full blob plus each individual field value. Worth a test asserting that stderr containing only the access key ID redacts to `reason=<redacted>` — the two existing leak tests both use the whole blob, which is why this passes today.

internal/infrastructure/cloud/aws/aws.go
```diff
@@ -298,20 +306,37 @@ func (c *Connector) buckets(ctx context.Context, client *s3.Client, account acco
...
+					// provider_error and left Public unknown, so an evidence-backed negative was
+					// downgraded to "we could not tell".
+					resource.Public = cloudposture.StateDisabled
+				case statusErr != nil:
 					*gaps = append(*gaps, *coverageGap(account.id, "storage", statusErr))
-				} else if status.PolicyStatus != nil && awssdk.ToBool(status.PolicyStatus.IsPublic) {
+				case status.PolicyStatus != nil && awssdk.ToBool(status.PolicyStatus.IsPublic):
 					resource.Public = cloudposture.StateEnabled
+				default:
+					resource.Public = cloudposture.StateDisabled
```
pho-veteran, line 327: **Major — the `default` cases widen behavior well beyond what the commit message and PR description claim.**

The original had **no `else`**:

```go
if statusErr != nil {
    *gaps = append(...)
} else if status.PolicyStatus != nil && awssdk.ToBool(status.PolicyStatus.IsPublic) {
    resource.Public = cloudposture.StateEnabled
}
```

so two *success* cases fell through to the `StateUnknown` initializer at line 304: (a) `PolicyStatus` non-nil with `IsPublic=false`, and (b) `PolicyStatus == nil`. This `default` collapses both to `StateDisabled`. Case (a) is a legitimate observed negative. Case (b) is not — a 200 response carrying a nil config field is not evidence of absence, and recording it as a definitive negative is the same category error the PR is fixing in the other direction.

Identical shape at lines 338-339 for encryption: success with `ServerSideEncryptionConfiguration == nil` now yields `StateDisabled` instead of `StateUnknown`.

This is load-bearing downstream, which is why it matters rather than being cosmetic:

- `internal/domain/cloudposture/catalog.go:100` — `if r.Encrypted == StateDisabled` raises the unencrypted-bucket finding.
- `internal/domain/cloudposture/catalog.go:130` — `if !ok || state == StateUnknown || exp.State == StateUnknown { continue }` skips IaC-drift comparison **only** for `StateUnknown`.

So moving success-path nils from Unknown to Disabled manufactures both new posture findings and new `RuleIaCLiveDrift` findings. Note this is the opposite direction from the defect being fixed (which suppressed findings), so it will not show up as a missing finding in the e2e run — the live AWS result of `findings=0` is consistent with buckets that populate `PolicyStatus` normally.

The commit message and the PR body only claim that the named *error* codes become `Disabled`. The `default` widening is unstated and, per my read of the new tests in aws_test.go, untested — no case exercises a successful call with a nil config field.

Suggested fix: make the negative explicit rather than a catch-all, e.g. `case status.PolicyStatus != nil: resource.Public = cloudposture.StateDisabled` and drop `default`, leaving a nil `PolicyStatus` at `StateUnknown`. Same for encryption. If the widening *is* intended, it should be stated in the commit message and pinned by a test.

pho-veteran (commented):
Security-architecture and correctness review. The core diagnoses hold up: the fleet route split is correct in both directions (all 8 patterns registered by `fleetRouter.handler()` are covered by the 5 prefixes, and the 5 operator routes are disjoint from every prefix), tenant binding never reads a request field, and the CSPM reorder is a faithful block transposition whose new reconciled-but-unattributed window is idempotently recoverable. `go build`, `go vet`, and the tests for all nine changed packages pass.

Two findings I would gate merge on, both inline below. Neither is a pre-existing issue — each is introduced by this PR, and one of them contradicts a safety claim in the PR description.

Smaller items not worth blocking on, for a follow-up:

- `fleetAgentPlanePrefixes` (router.go:619) is a package-level mutable `var` slice with a single consumer, which the no-global-mutable-state rule covers; it could be a local in `Handler()`.
- `TestFleetAgentPlanePrefixesCoverEveryAgentRoute` hand-copies the same 8 paths it guards and reimplements prefix matching, so it cannot detect a *newly added* agent route — the exact regression the comment at router.go:617 says it prevents. A shared `var fleetAgentRoutes` iterated by both `handler()` and the test would actually close that.
- runner.go:452 guards on `strings.HasPrefix(bin, "/")`, but `Run` only rewrites `spec.Name` to an absolute resolved path inside `if r.binreg != nil` (runner.go:188-198). In the documented legacy PATH-trust mode a helper invoked by bare name keeps a relative `spec.Name`, no bind is emitted, and the original `execvp: No such file or directory` returns. The new test passes an already-absolute name, so it does not cover this. (The prefix matching itself is correct — `underCuratedRoot` is segment-aligned, so `/libexec` and `/lib-evil` are genuinely excluded.)
- executor.go:60-62 slices at a byte offset, so `reason[:diagnosticCap]` can split a multi-byte rune and emit invalid UTF-8 into a structured log and a JSON body; `strings.Fields` afterward does not repair it. The redact → placeholder-check → truncate ordering is correct, though.
- Dropping `uniqueItems` leaves nothing unenforced (evidence.go:89-91 rejects duplicates), but it converts a decoder-level impossibility into a runtime validation error. Worth adding "each token id at most once" to the prompt at prompt.go:132, or deduping rather than erroring.
- `ScanJobStore.Save` is not tenant-scoped, so the tenant binding is inert at the two `jobs.Save` sites; the comment at service.go:1771 overstates it. The `RecordProjectAnalysis` site is the one where it matters, and that one is correct.

internal/infrastructure/cloudsandbox/executor.go
pho-veteran: Confirmed and fixed in 31eee8d.

You were right that the whole-blob scrub never matches a single echoed field. I reproduced it with a test first — with the old code the returned error reads literally:

```
sandboxed CSPM helper failed: exit_code=1 timed_out=false truncated=false reason=The AWS Access Key Id you provided does not exist in our records: AKIAIOSFODNN7EXAMPLE
```

and the same for a lone `secret_access_key` and a lone `session_token`.

Two things worth flagging beyond what you found:

**The stdout scrub had the identical gap** (`executor.go:171` before the change, `redact.Bytes(result.Stdout, [][]byte{secret})`). That path decodes into inventory that gets sealed as evidence, so a field echoed there would have been persisted, not just logged. Both call sites now share one decomposed secret set.

**I did not enumerate AWS's field names.** Hardcoding `access_key_id`/`secret_access_key`/`session_token` would make this executor provider-specific, and it is the shared path for Azure and GCP too. Instead `credentialSecrets` walks the credential as opaque JSON and collects every string value, recursing into nested objects (GCP service-account keys nest `private_key`).

The one judgement call: a **16-character floor** on what gets scrubbed independently. Credential documents also carry short structural values — `"us-east-1"`, `"aws"`, `"true"` — and scrubbing those would match ordinary prose in almost any diagnostic and drop every reason via the placeholder check, silently reverting the observability this function exists to provide. Real credential material clears the floor comfortably (AWS access key id 20, secret key 40, tokens longer). The whole document stays in the set regardless, and a non-JSON credential falls back to blob-only. `TestCredentialSecretsDecomposesTheDocument` pins all of that, including that `us-east-1` is *not* scrubbed.

I also took the `diagnosticCap` rune-splitting item from your non-blocking list in the same commit, since it lands in the same two sinks: truncation now walks back to a valid UTF-8 boundary, covered by `TestDiagnosticTruncatesOnARuneBoundary` across 2-, 3-, and 4-byte runes.

internal/infrastructure/cloud/aws/aws.go
pho-veteran: Confirmed and fixed in 2613ea6. You read it correctly on both counts: the `default` widening was unstated in the commit message and untested.

Case (b) is the one that matters, and I agree it is the same category error in the other direction — a 200 carrying a nil field establishes nothing, so recording `Disabled` manufactures an evidence-backed negative out of a missing field. Worse than the false coverage gap it replaced, because a `StateDisabled` on `Encrypted` is what the posture rules key an unencrypted-bucket finding on, and `DetectDrift` compares these states against IaC expectations — so a nil field could have produced a finding *and* a drift finding from no evidence.

Both switches are now explicit:

- named absence error code → `StateDisabled` (the observed negative)
- field present and negative → `StateDisabled`
- field nil on a 200 → **state left `StateUnknown`**, with a comment saying why
- any other error → coverage gap, unchanged

`TestSuccessfulCallWithNilConfigStaysUnknown` covers case (b) for both `GetBucketPolicyStatus` and `GetBucketEncryption`. Verified failing against the old code:

```
Encrypted = "disabled", want "unknown" (a nil config field on a 200 is not evidence of absence)
Public = "disabled", want "unknown" (a nil policy-status field on a 200 is not evidence of absence)
```

One test-mechanics note in case it saves you time later: an XML stub like `<ServerSideEncryptionConfiguration></ServerSideEncryptionConfiguration>` deserializes to a **non-nil empty struct**, so it exercises case (a), not (b). I probed the SDK to find that only an empty 200 body yields a genuinely nil config field, which is what the test now returns.

pho-veteran:
Both gating findings are fixed, and I took all six of the smaller items rather than deferring them — two turned out to be more than cosmetic once I looked, so leaving them for a follow-up would have shipped a known gap.

Five commits pushed (`31eee8d`..`d6a8017`).

### Gating findings

| | Fix | Test verified failing without it |
| --- | --- | --- |
| Partial-secret leak in stderr | `31eee8d` | `TestExecutorRedactsIndividualCredentialFields` — reproduced `AKIAIOSFODNN7EXAMPLE` verbatim in the returned error |
| `default` widening on nil config | `2613ea6` | `TestSuccessfulCallWithNilConfigStaysUnknown` — reproduced `Encrypted = "disabled", want "unknown"` |

Details in the two inline threads. Two things I found while fixing them that were not in your review:

- The **stdout** scrub had the same whole-blob gap as stderr, and that path feeds sealed evidence — so a field echoed there was persisted, not just logged. Both call sites now share one decomposed secret set.
- The nil-config widening could have produced **both** a bucket posture finding and an IaC drift finding from no evidence, since `DetectDrift` compares these same states against expectations.

### The six smaller items — all taken

**1 + 2, `fleetAgentPlanePrefixes` and its guard test.** You were right that the test could not detect a newly added agent route, so I did the shared-declaration fix you suggested rather than just making the var a local. `fleetAgentPlaneRoutes()` now declares each pattern with the prefix it mounts under; `fleetRouter.handler()` registers from it and `Router.Handler()` derives its mounts from it. The package-level mutable var is gone.

The guard test now iterates that declaration and drives each route through the real `Handler()`. I verified it by adding a fake agent route with a non-covering mount:

```
/api/v1/fleet/newthing was served by the HUMAN chain (401); it needs a mount in fleetAgentPlaneRoutes
declared route "POST /api/v1/fleet/newthing" registered no handler; it would 404 on a mounted path
```

That is the regression the old test claimed to prevent and could not. A second test catches the inverse — a declared pattern with no handler, which `handler()` skips and would otherwise be a silent 404 on a mounted path.

**3, relative `spec.Name` in legacy PATH-trust mode.** Not cosmetic — this was a real hole in my own fix, and your read of the control flow was exactly right. With `binreg == nil` a bare tool name stayed relative, no bind was emitted, and the original `execvp: No such file or directory` came back. Resolution now happens unconditionally in `resolveToolPath`, with verification still gated on the registry. Resolution is also what stops bwrap re-resolving via PATH to a different binary, so it belongs on every path, not only the verified one. `TestSandboxBindsResolvedBinaryOnTheLegacyPath` asserts the bind is emitted with no registry; `TestResolveToolPathRunsWithoutABinaryRegistry` covers resolution and the unresolvable-name passthrough.

**4, rune-splitting at `diagnosticCap`.** Folded into `31eee8d` since it lands in the same two sinks. Truncation walks back to a valid UTF-8 boundary; covered across 2-, 3-, and 4-byte runes.

**5, `uniqueItems` guidance.** Added "Cite each ID at most once." to the evidence prompt. I kept `validateEvidenceCitations` **rejecting** rather than silently deduping: a duplicate citation means the model misread the token list, and quietly repairing that would hide a real signal about prompt quality. The prompt line should make it rare rather than papering over it.

**6, overstated comment at the `jobs.Save` sites.** Correct, `ScanJobStore.Save` is not tenant-scoped. Comment now says the call is detached so the record still lands after the completion timeout fires, and notes explicitly that the tenant matters at the `RecordProjectAnalysis` call — which is where the original defect actually was.

### Checks

`go build` · `go vet` · **`go test ./...` 245 packages, 0 failures** · `make lint` **0 issues** · `make typecheck` · `TestHostileHarness`

### One note on scope

`credentialSecrets` deliberately treats the credential as opaque JSON instead of enumerating AWS field names — this executor is the shared path for Azure and GCP, and hardcoding one provider's shape would break that. The tradeoff is a 16-character floor on what gets scrubbed independently, so short structural values (`"us-east-1"`, `"aws"`) do not match ordinary prose and drop every diagnostic via the placeholder check. Reasoning is in the inline thread; say the word if you would rather see per-provider field lists and I will switch it.

Not merging — waiting on your call.

pho-veteran:
## Finalized — three review rounds, all findings closed

Ready for your decision. Not merging.

Since the last update I ran two more fan-out rounds (security · architecture · correctness, concurrent, against the same tree). Round 2 found real problems and I fixed them; round 3 confirmed clean.

### Round 2 — two Highs, and they contradicted each other

Two reviewers independently landed on the same line, `credentialSecrets`'s 16-character floor, for **opposite** reasons:

- **Correctness (High):** it *over*-matched. A GCP `client_email` and `project_id` clear 16 chars and appear verbatim in legitimate enumerated inventory (`gcp/connector.go:668` sets `Name: member`, project id is in every resource ID). Scrubbing them made a **successful** enumeration fail with `output contained credential material` — so every GCP CSPM run would have retried twice and dead-lettered. A regression I introduced in the previous round.
- **Security (High):** it *under*-matched. Azure validates `client_secret` non-empty only, so a short secret slipped the floor entirely and could still reach a log or an API error body.

Both reproduced verbatim before fixing:

```
legitimate inventory was rejected: sandboxed CSPM output contained credential material
field value "sh0rt" is not scrubbed in a diagnostic; a short secret must not slip a length threshold
```

That disagreement was the useful part: it proved **length is the wrong axis**. The two sinks tolerate a false match differently — a diagnostic can be dropped, but stdout is the enumeration result where a placeholder hit *rejects the run*. So the fix splits by sink, not by size:

| | `diagnosticSecrets` (stderr) | `outputSecrets` (stdout) |
| --- | --- | --- |
| Membership | every string value, any length | only values whose **field name** marks them secret |
| Cost of a false match | reason text dropped | **whole run rejected** |
| Cost of a miss | credential disclosure | credential in sealed evidence |

Both keep the whole document, so an unparseable credential echoed wholesale is still caught.

The security review also raised a **Medium** I agreed with: making `exec.LookPath` unconditional had widened confinement. Previously bwrap resolved a bare name against the sandbox's own curated PATH, so a binary outside the curated root failed closed; resolving on the host and then binding it meant an inherited `PATH=/tmp/attacker:$PATH` entry could be bound in. Fixed by resolving a bare name **only** when it will also be integrity-verified — an absolute operator path needs no lookup, which is what the documented `/opt/synapse/bin/synapse-cspm` layout relies on.

Also took every remaining minor: unconditional `strings.ToValidUTF8` (a latin-1 helper message under the cap was reaching logs), the fleet route relation made **total by construction** (handler on the struct — no map, no lookup, no missing-case branch, no panic needed), and a misattached test comment.

### Round 3 — confirmation

All three reviewers returned **CLOSED** on every prior finding, with **no new bugs and no security regression**. Correctness verified the no-behavior-change claim on the final cleanup as **TRUE**, and specifically confirmed no existing tool regressed: every production `sandbox.NewRunner` call site calls `SetBinaryRegistry` immediately after (`cmd/synapse-api/main.go:683`, `:1050`, `cmd/synapse-worker/main.go:219`), so `binreg != nil` on every sandboxed path and syft/grype/naabu/callgraph/dast all resolve and bind exactly as before.

Two reviewer catches I acted on rather than waved off:

- The correctness reviewer noticed my commit message said I *moved* a Windows comment when `git show 91f86c1^` proves it was net-new. Amended — the message was wrong, the code was fine.
- The security reviewer couldn't run commands and flagged the `secretFieldMarkers` list as unverified. I diffed it: **byte-for-byte identical**.

### Residual items — tracked, not fixed here

Three things the reviews surfaced that I deliberately did **not** change, because each is pre-existing and out of this PR's scope:

1. **`binregistry.Verify` symlink TOCTOU.** `Verify` calls `filepath.EvalSymlinks` internally and hashes the resolved target, but returns only `error` — so `spec.Name` stays pre-resolution and a symlink swap between verify and bwrap's own resolution is theoretically possible. Predates this PR; unchanged by it. Worth its own issue.
2. **`redact.Bytes` drops encoded forms under 4 bytes** (`redact.go:66`), so a 1–3 character credential value is never scrubbed verbatim even though `diagnosticSecrets` includes it. Not realistic for any current provider field, but it means my test's "any length" claim is set-membership, not an end-to-end guarantee.
3. **Subtree mounts** `/api/v1/fleet/work/` and `/api/v1/fleet/inventory/` could still shadow a *future* operator route under those prefixes. Latent, neither improved nor worsened; `TestEveryAgentRouteReachesTheAgentPlane` covers the reverse direction.

Happy to file these as separate issues if you want them tracked.

### Final state

18 commits, 19 code files, no artifacts or `.gitignore`.

| Check | Result |
| --- | --- |
| `go build ./...` · `go vet ./...` | pass |
| `go test ./...` | **245 packages, 0 failures** |
| `make lint` (golangci-lint v2) | **0 issues** |
| `make typecheck` | pass |
| `TestHostileHarness` | pass |
| `cd web && pnpm test` | **39 files, 311 tests** |

Ten original defects plus every review finding, each with a regression test I confirmed fails without its fix. Several reproduce the exact production error string.

One thing I'd still flag for your judgement: `outputSecrets` relies on a **field-name heuristic**, and the marker `key` also matches AWS `access_key_id`. Nothing emits one today (the AWS connector enumerates IAM users, not access keys), so I documented the tradeoff at the marker list rather than narrowing it — over-scrubbing fails safe (availability), under-scrubbing leaks (confidentiality). If a future connector enumerates access keys it will need to narrow that list or rename the field. Say the word if you'd rather see per-provider field allowlists instead.

PR #582, opened 2026-08-17
fix(binregistry): execute verified resolved path

### LINKED ISSUE #580: fix(binregistry): Verify discards the resolved path, so the verified inode is not provably the executed one
## Summary

`binregistry.Registry.Verify` resolves symlinks and hashes the **resolved** target, but returns only `error` — it discards the path it actually verified. The sandbox then binds and executes the **pre-resolution** path. So the file whose hash was checked is not provably the file that runs, and a symlink swapped between the two leaves the integrity check (F5) attesting to a different inode than the one executed.

Found during review of #577 (see the residual-items note on that PR). Pre-existing and **not** introduced by that PR — filing separately so it is tracked rather than folded into an unrelated change.

## Detail

`internal/platform/binregistry/registry.go:45-70`:

```go
func (r *Registry) Verify(path string) error {
	resolved, err := filepath.EvalSymlinks(path)   // <- resolves
	if err != nil { ... }
	sum, err := hashFile(resolved)                 // <- hashes the RESOLVED target
	...
	// pins are keyed by `resolved` (or its basename)
}
```

`resolved` never leaves the function. In `internal/infrastructure/sandbox/runner.go` the same unresolved `spec.Name` is used for all three steps:

| Step | Line | Value |
| --- | --- | --- |
| Settle path (PATH lookup only, **no** symlink resolution) | `runner.go:218` | `spec.Name` |
| Integrity verify | `runner.go:223` | `spec.Name` → internally resolves to `X` |
| Sandbox bind | `runner.go:478` | `spec.Name` → bwrap resolves independently |
| Exec | `runner.go:403` | `spec.Name` → bwrap resolves independently |

If `spec.Name` is a symlink, `Verify` hashes its target at time *T1* and bwrap resolves it again at *T2*. Nothing guarantees `X` at *T1* is the same inode as at *T2*.

The docs describe F5 as closing the "compromised tool binary" threat, and #577 explicitly added a comment claiming verify-path == exec-path. That property holds today for PATH resolution (fixed in #577) but **not** for symlink resolution.

## Exploitability: low

This is defense-in-depth, not a remotely reachable hole:

- Requires the ability to write into the tool directory or swap a symlink there — i.e. host access the sandbox is not meant to defend against anyway.
- `spec.Name` is never derived from request, agent, or target input. Every call site takes it from operator config (`cfg.SyftBin`, `cfg.GrypeBin`, `cfg.CSPMHelperBin`, `Executor.binary`) or a hardcoded literal (`"git"`, `"subfinder"`, `"httpx"`). I grepped every `ToolSpec{Name:` construction to confirm this.
- `SYNAPSE_CSPM_HELPER_BIN` is additionally forced absolute and hash-pinned before use (`cmd/synapse-worker/main.go:202-218`).

So: worth closing properly, not worth a hotfix.

## Suggested fix

Return the verified path and use it downstream, so the verified inode is the bound and executed one:

```go
// Verify hashes the binary at path (symlinks resolved), checks it against its pin, and returns
// the resolved path so the caller can bind and exec the EXACT file that was verified.
func (r *Registry) Verify(path string) (string, error)
```

Then in `runner.go`:

```go
if r.binreg != nil {
	resolved, verr := r.binreg.Verify(spec.Name)
	if verr != nil {
		return ports.ToolResult{}, fmt.Errorf("%w: %v", shared.ErrValidation, verr)
	}
	spec.Name = resolved // bind + exec the verified inode
}
```

There is only **one** production call site (`runner.go:223`), so the signature change is cheap. Tests that call `Verify` would need updating.

Two things to decide while implementing:

1. **Does binding the resolved path break any deployment that relies on a symlink indirection?** e.g. `/usr/local/bin/syft -> /opt/syft-1.45.1/syft`. Binding the resolved target is strictly more precise, but it changes what appears inside the sandbox, so the path a tool sees in `argv[0]` and in its own error messages changes. Worth checking `scripts/install-tools.sh` for whether it creates symlinks.
2. **Residual TOCTOU between hash and exec remains** even after this — the file could still be replaced between `hashFile` and bwrap's `execvp`. Fully closing that needs an fd-based approach (open once, `fstat`/hash the fd, pass it via `--file`/`fexecve` semantics), which bubblewrap does not directly expose. Returning the resolved path shrinks the window to "same inode, content swapped in place" rather than "different inode entirely"; say so in the comment rather than over-claiming, since #577 already had two findings for overstated comments in this area.

## Acceptance

- `Verify` returns the resolved path; the sandbox binds and execs that path.
- A regression test proving the bound/exec'd path is the verified one for a symlinked tool.
- The comment at `runner.go:466-472` states the actual guarantee, including the residual window in (2) above.

XUanhoa04:
## Summary

- make `binregistry.Registry.Verify` return the symlink-resolved path only after successful integrity verification
- replace the sandbox tool path with that verified result before building bubblewrap bind and exec arguments
- document the exact guarantee and the remaining in-place-content replacement window
- add registry and Linux sandbox regressions proving a symlinked tool executes via the verified resolved path

## Security impact

Previously, verification hashed the target of a symlink but discarded that resolved path. A retarget between verification and bubblewrap could therefore select a different file for bind/exec. The sandbox now carries the verified path forward, so later symlink retargeting cannot change the selected target.

This deliberately does not claim to close the remaining hash-to-exec window for replacing the resolved target's contents; that requires a future fd-based execution design.

## Verification

- `go test ./internal/platform/binregistry ./internal/infrastructure/sandbox -count=1`
- `go test ./internal/infrastructure/tools/syft ./internal/infrastructure/tools/grype ./internal/infrastructure/recon -count=1`
- `go vet ./internal/platform/binregistry ./internal/infrastructure/sandbox`
- Linux/amd64 sandbox test binary cross-compiles successfully with `CGO_ENABLED=0`
- `git diff --check`
- `go test ./... -count=1`: every code package passed; the unchanged `docs` package reports four publication-mirror drift failures already present at `upstream/main` (`24693b2`)

Closes #580

### DIFF
--- a/internal/infrastructure/sandbox/binintegrity_integration_test.go
+++ b/internal/infrastructure/sandbox/binintegrity_integration_test.go
@@ -54,3 +54,36 @@ func TestBinaryIntegrityRefusesTamper(t *testing.T) {
 	}
 	t.Logf("F5 OK: replaced binary refused before exec (%v)", err)
 }
+
+// TestBinaryIntegrityExecutesResolvedSymlinkTarget proves the path returned by the registry is the
+// one passed to bubblewrap and exec. Re-resolving the input symlink downstream would reopen the
+// target-swap window that verification is intended to close.
+func TestBinaryIntegrityExecutesResolvedSymlinkTarget(t *testing.T) {
+	if _, err := exec.LookPath("bwrap"); err != nil {
+		t.Skip("bwrap not installed")
+	}
+	sb, err := sandbox.NewRunner(20*time.Second, 8<<20, 1<<30, 256)
+	if err != nil {
+		t.Skipf("sandbox unavailable: %v", err)
+	}
+	sb.SetBinaryRegistry(binregistry.New(nil, true))
+
+	dir := t.TempDir()
+	link := filepath.Join(dir, "verified-sh")
+	if err := os.Symlink("/bin/sh", link); err != nil {
+		t.Skipf("symlink creation unavailable: %v", err)
+	}
+	resolved, err := filepath.EvalSymlinks(link)
+	if err != nil {
+		t.Fatal(err)
+	}
+	result, err := sb.Run(context.Background(), ports.ToolSpec{
+		Name: link, Args: []string{"-c", `printf %s "$0"`}, Workdir: dir,
+	})
+	if err != nil {
+		t.Fatalf("run symlinked tool: %v", err)
+	}
+	if got := string(result.Stdout); got != resolved {
+		t.Fatalf("executed argv[0] = %q, want verified resolved path %q", got, resolved)
+	}
+}
--- a/internal/infrastructure/sandbox/runner.go
+++ b/internal/infrastructure/sandbox/runner.go
@@ -210,19 +210,21 @@ func (r *Runner) Run(ctx context.Context, spec ports.ToolSpec) (ports.ToolResult
 		return ports.ToolResult{}, fmt.Errorf("%w: build seccomp filter: %v", shared.ErrValidation, serr)
 	}
 	defer func() { _ = seccompF.Close() }()
-	// Settle the exec path BEFORE verification, so the path that is verified is the path that is
-	// bound and executed (closing the verify-path != exec-path gap). An absolute name is kept as
-	// given - that is the operator-configured helper the bind below exists for. The host PATH counts
-	// as an authority for a bare name only when that name will also be integrity-verified.
+	// Settle PATH lookup before verification. An absolute name is kept as given; the registry below
+	// then resolves any symlinks and returns the exact path it hashed. The host PATH counts as an
+	// authority for a bare name only when that name will also be integrity-verified.
 	hostPATHIsAuthority := r.binreg != nil
 	spec.Name = resolveToolPath(spec.Name, hostPATHIsAuthority)
-	// F5: verify the tool binary's integrity before it runs. The resolved on-disk binary
-	// must match its pin (operator hash and/or trust-on-first-use); a replaced binary is
-	// refused. Defends the "compromised tool binary" threat the audit named.
+	// F5: verify the tool binary's integrity before it runs, then bind and execute the returned
+	// symlink-resolved path. This prevents a symlink swap from selecting a different target after
+	// verification. It does not eliminate the residual window in which that target's contents could
+	// be replaced between hashing and exec; fully closing that requires an fd-based execution path.
 	if r.binreg != nil {
-		if verr := r.binreg.Verify(spec.Name); verr != nil {
+		verifiedPath, verr := r.binreg.Verify(spec.Name)
+		if verr != nil {
 			return ports.ToolResult{}, fmt.Errorf("%w: %v", shared.ErrValidation, verr)
 		}
+		spec.Name = verifiedPath
 	}
 	// F3: a per-run cgroup v2 with hard memory.max + pids.max so a memory/fork bomb is
 	// contained on EVERY path (egress and isolated), independent of systemd-run. The tool
@@ -474,7 +476,8 @@ func (r *Runner) bwrapArgs(spec ports.ToolSpec, sharedNet bool, hostsFile string
 	// "bwrap: execvp ...: No such file or directory". Bind the single resolved FILE read-only, never
 	// its directory, so nothing else in that tree becomes visible. Only an absolute path is bound,
 	// and resolveToolPath keeps the host PATH from making a bare name absolute unless it is also
-	// integrity-verified - so what gets bound is operator authority, not an inherited PATH entry.
+	// integrity-verified. When the registry is enabled, spec.Name is also the symlink-resolved path
+	// it hashed, so a later symlink retarget cannot change this bind or the exec path.
 	if bin := strings.TrimSpace(spec.Name); strings.HasPrefix(bin, "/") && !underCuratedRoot(bin) {
 		args = append(args, "--ro-bind-try", bin, bin)
 	}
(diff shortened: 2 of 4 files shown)


nghiadaulau (approved):
## Approve — strict review pass (sandbox exec-path TOCTOU)

Reviewed against `main` with local verification and two independent strict reviews (clean-architecture + security, adversarial). This is a correct, well-scoped, well-tested fix for the #580 TOCTOU. Approving and merging, with one hardening applied on merge.

### What I verified ✅
- **The symlink-swap window is genuinely closed.** `Verify` now returns the `EvalSymlinks`-resolved path, and `Runner.Run` sets `spec.Name = verifiedPath` so the exact file that was hashed is the one `--ro-bind`'d and exec'd (same string for hash, bind, and exec). A post-verify retarget of the original link selects nothing new. Traced end to end.
- **`Verify` is clean on every branch** — every success path returns `resolved`, every error path returns `""` (no branch leaks the unresolved input). Confirmed by both reviewers and by `TestVerifyReturnsResolvedPath` (`resolved != link`) and the new empty-on-failure assertion in `TestExpectedHashMismatchRefused`.
- **Signature ripple fully handled** — `binreg` is the concrete `*binregistry.Registry` (no port to update), `runner.go:223` is the only caller, the `cmd/*` mains only construct/set it. `go build ./...` clean.
- **Dependency rule PASS** — `platform/binregistry` imports only stdlib; the infra→platform direction is respected.
- **Regression tests are non-vacuous** — `TestBinaryIntegrityExecutesResolvedSymlinkTarget` runs a symlinked tool through real bubblewrap and asserts `argv[0]` (`$0`) equals the *resolved* target, structurally proving a retarget is defeated.
- **Residual window honestly disclosed** — the fix does not claim to close the content/tree-replacement window between hash and exec (needs fd/`O_PATH`-based exec). No overclaim anywhere. Matches #580's own LOW/defense-in-depth framing.
- Build / vet / `golangci-lint v2.12.2` (0 issues) / `go test` for `binregistry` + `sandbox` all green.

### One hardening applied on merge (commit `4370998`)
The security review found the new verify==exec invariant had one relative edge: when integrity is enabled and a **bare** tool name is not on the host PATH, `resolveToolPath` leaves it bare, so `EvalSymlinks` resolves it **relative to CWD** and can return a *relative* path — which the sandbox skips binding (`HasPrefix(bin, "/")` is false), letting bwrap re-resolve the bare name against its own PATH and run a file the registry never hashed. Contained (needs a same-named file in CWD *and* on the sandbox PATH, else it fails closed), but it undercuts exactly the invariant this PR establishes. I added the reviewer's recommended guard — `Verify` refuses a non-absolute resolved path (`ErrIntegrity`) — plus a regression test (`TestVerifyRejectsNonAbsoluteResolvedPath`). Re-verified: binregistry + sandbox green, lint 0.

### Non-blocking
The residual-window comment could be broadened one line to note a same-class tree-rename (not only content replacement); optional, and not an overclaim as written. The thrice-duplicated match-and-return block in `Verify` could fold into a tiny helper; purely stylistic.

Solid defense-in-depth fix. Closes #580. Merging.

nghiadaulau (approved):
## Approve — strict review pass (sandbox exec-path TOCTOU)

Reviewed against `main` with local verification and two independent strict reviews (clean-architecture + security, adversarial). This is a correct, well-scoped, well-tested fix for the #580 TOCTOU. Approving and merging, with one hardening applied on merge.

### What I verified ✅
- **The symlink-swap window is genuinely closed.** `Verify` now returns the `EvalSymlinks`-resolved path, and `Runner.Run` sets `spec.Name = verifiedPath` so the exact file that was hashed is the one `--ro-bind`'d and exec'd (same string for hash, bind, and exec). A post-verify retarget of the original link selects nothing new. Traced end to end.
- **`Verify` is clean on every branch** — every success path returns `resolved`, every error path returns `""` (no branch leaks the unresolved input). Confirmed by both reviewers and by `TestVerifyReturnsResolvedPath` (`resolved != link`) and the new empty-on-failure assertion in `TestExpectedHashMismatchRefused`.
- **Signature ripple fully handled** — `binreg` is the concrete `*binregistry.Registry` (no port to update), `runner.go:223` is the only caller, the `cmd/*` mains only construct/set it. `go build ./...` clean.
- **Dependency rule PASS** — `platform/binregistry` imports only stdlib; the infra→platform direction is respected.
- **Regression tests are non-vacuous** — `TestBinaryIntegrityExecutesResolvedSymlinkTarget` runs a symlinked tool through real bubblewrap and asserts `argv[0]` (`$0`) equals the *resolved* target, structurally proving a retarget is defeated.
- **Residual window honestly disclosed** — the fix does not claim to close the content/tree-replacement window between hash and exec (needs fd/`O_PATH`-based exec). No overclaim anywhere. Matches #580's own LOW/defense-in-depth framing.
- Build / vet / `golangci-lint v2.12.2` (0 issues) / `go test` for `binregistry` + `sandbox` all green.

### One hardening applied on merge (commit `4370998`)
The security review found the new verify==exec invariant had one relative edge: when integrity is enabled and a **bare** tool name is not on the host PATH, `resolveToolPath` leaves it bare, so `EvalSymlinks` resolves it **relative to CWD** and can return a *relative* path — which the sandbox skips binding (`HasPrefix(bin, "/")` is false), letting bwrap re-resolve the bare name against its own PATH and run a file the registry never hashed. Contained (needs a same-named file in CWD *and* on the sandbox PATH, else it fails closed), but it undercuts exactly the invariant this PR establishes. I added the reviewer's recommended guard — `Verify` refuses a non-absolute resolved path (`ErrIntegrity`) — plus a regression test (`TestVerifyRejectsNonAbsoluteResolvedPath`). Re-verified: binregistry + sandbox green, lint 0.

### Non-blocking
The residual-window comment could be broadened one line to note a same-class tree-rename (not only content replacement); optional, and not an overclaim as written. The thrice-duplicated match-and-return block in `Verify` could fold into a tiny helper; purely stylistic.

Solid defense-in-depth fix. Closes #580. Merging.

PR #612, opened 2026-08-18
feat(httpapi): add dependency readiness probe

### LINKED ISSUE #585: Observability: /readyz readiness probe checking DB, object store, and migrations
> Part of #583 (production-readiness epic). Item **A2** · Priority **P0** · Group A (Operations & Observability).

## Problem
`/healthz` returns a **constant** body and does not reflect real readiness. It answers `{"service":"synapse-api","status":"ok"}` regardless of whether PostgreSQL is reachable, the object store is reachable, or database migrations have finished. There is no `/readyz` or `/livez`. A `200` from `/healthz` only proves the process is alive and the HTTP listener accepts connections.

## Evidence
- `E-OPS-01` — `internal/adapter/httpapi/router.go:266-268` (constant handler).
- Verified on `main`.

## Impact
An orchestrator (k8s, compose `--wait`, an LB health check) cannot tell when the API is actually able to serve. Traffic may be routed before migrations complete or while the database/object store is unreachable, producing user-visible failures that a real readiness gate would have prevented.

## Proposed work
- Add a `/readyz` endpoint that returns `200` only when **all** of these pass, and `503` otherwise with a per-check body indicating which failed:
  - PostgreSQL reachable (`SELECT 1` within a short timeout).
  - Object store reachable when `SYNAPSE_BLOB_ENDPOINT` is configured (a cheap head/list on the evidence bucket).
  - Migrations applied: `goose_db_version` count equals the embedded migration count.
- Keep `/healthz` as a pure **liveness** probe (unchanged constant).
- Keep the check timeouts short and the endpoint unauthenticated (same posture as `/healthz`), but ensure it leaks no internal detail beyond check name + pass/fail.

## Acceptance criteria
- [ ] `/readyz` returns `200` only when DB + blob (if configured) + migrations are all ready.
- [ ] `/readyz` returns `503` with the failing check name(s) when any dependency is down.
- [ ] `/healthz` remains a constant liveness probe.
- [ ] Documented; a k8s/compose example uses `/readyz` for readiness and `/healthz` for liveness.
- [ ] Behavioral tests cover ready, DB-down, and migrations-incomplete cases.

## Dependencies
Interacts with B1 (#588): a real readiness gate makes the migration-vs-healthcheck race (`R-02`) observable and safer.

XUanhoa04:
## Summary

- add a public `GET /readyz` endpoint with concurrent, bounded checks for PostgreSQL connectivity, embedded migration state, and the configured evidence bucket
- preserve `GET /healthz` as constant liveness and return only named `ok`/`failed` states without dependency error details
- wire the full Compose healthcheck to readiness and document separate Compose/Kubernetes liveness and readiness probes
- cover ready, database-down, migration-incomplete, timeout, unauthenticated access, missing/unreachable bucket, and real PostgreSQL behavior

## Why

The existing `/healthz` response is intentionally constant, so orchestrators cannot distinguish a live process from an API that cannot serve traffic. This gives Compose, Kubernetes, and load balancers a dependency-aware signal without changing liveness semantics.

## Verification

- `go test ./internal/adapter/httpapi ./internal/infrastructure/blob ./internal/infrastructure/persistence/postgres ./cmd/synapse-api ./api -count=1`
- `SYNAPSE_TEST_DB_DSN=... go test ./internal/infrastructure/persistence/postgres -run TestPostgresReadinessChecks -count=1 -v` (PostgreSQL 18)
- `go vet ./internal/adapter/httpapi ./internal/infrastructure/blob ./internal/infrastructure/persistence/postgres ./cmd/synapse-api ./api`
- `docker compose -f deploy/docker-compose.full.yml config --no-interpolate --quiet`
- `go test ./... -count=1` passes all code packages; the existing Windows-only docs mirror drift check still reports four unchanged repository mirrors

Closes #585

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -13,7 +13,7 @@
 # REQUIRED before the server will start
 # =============================================================================
 # Bootstrap-admin bearer token – the API exits if this is empty. Operational API
-# routes require it; GET /healthz is intentionally public. Generate: openssl rand -hex 32
+# routes require it; GET /healthz and GET /readyz are intentionally public. Generate: openssl rand -hex 32
 SYNAPSE_API_TOKEN=
 
 # =============================================================================
--- a/README.md
+++ b/README.md
@@ -196,7 +196,7 @@ make install                       # Go modules + web deps
 make tools                         # syft + grype into ./bin
 export PATH="$PWD/bin:$PATH"
 
-export SYNAPSE_API_TOKEN="$(openssl rand -hex 32)"   # required for operational API routes; /healthz is public
+export SYNAPSE_API_TOKEN="$(openssl rand -hex 32)"   # required for operational API routes; /healthz and /readyz are public
 make dev                           # API on :8080, dashboard on :5173
 ```
 
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -46,7 +46,7 @@ operator is responsible for holding written permission.
 Hardening highlights operators should preserve:
 
 - Always set `SYNAPSE_API_TOKEN`; operational API routes require authentication. The intentionally
-  public `GET /healthz` endpoint exposes only service health.
+  public `GET /healthz` and `GET /readyz` endpoints expose only liveness and named dependency pass/fail state.
 - Run the execution sandbox and live recon on a Linux host; they **fail closed** without
   the required kernel features rather than running unsandboxed.
 - Keep secrets in the credential vault; never place them in logs or source.
--- a/api/openapi.yaml
+++ b/api/openapi.yaml
@@ -7,7 +7,7 @@ info:
     Authorized security testing only: every scan is gated by an engagement's
     scope and authorization window, enforced server-side.
 
-    Auth: all endpoints except `/healthz` require a bearer token
+    Auth: all endpoints except `/healthz` and `/readyz` require a bearer token
     (`Authorization: Bearer <SYNAPSE_API_TOKEN>`). After authentication, every
     endpoint except the AUP endpoints also requires that the current Acceptable
     Use Policy has been accepted, else `403`.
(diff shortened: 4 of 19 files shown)


nghiadaulau (approved):
**Approving — reviewed strictly (clean-architecture + security), builds/tests clean when merged onto current `main`.**

Verified locally (CI is disabled): merged onto `main` (post-A0) → `go build ./...`, `go vet`, and `go test ./internal/adapter/httpapi/... ./internal/infrastructure/blob/...` all green.

**Architecture — PASS.** Textbook dependency inversion: `readiness.go` imports only stdlib and depends on a consumer-defined `ReadinessCheck func(context.Context) error`; the concrete Postgres/MinIO probes are injected as closures at the composition root (`cmd/synapse-api/main.go`). The adapter never sees infra types. `CheckReady` is deliberately kept OFF `ports.BlobStore` (port stays minimal). Fail-closed (timeout/any error → 503 `not_ready`), bounded 2s timeout, concurrent checks over pooled connections (no fresh dials, no goroutine leak), `/healthz` preserved as a constant liveness probe with a regression test.

**Security — PASS.** No info-disclosure: the response exposes only `status` + per-check `ok`/`failed` (operator-chosen names) — the underlying errors/DSNs/versions are never serialized or logged (there's a negative test asserting a `postgres://secret@…` error does not leak). `/readyz` is public like `/healthz` (added to the public + AUP-exempt sets only for its own key; no other route's auth surface widened), and it reuses pooled connections under the timeout.

Non-blocking follow-ups (fine as later PRs):
- Consider a short-TTL cache of the readiness verdict so an unauthenticated burst collapses to one dependency probe (mild amplification vs the constant `/healthz`); rate-limiting is expected at the reverse proxy per `deployment.md`.
- Optional server-side `slog` of the (safe-to-log-internally) failure reason, kept out of the response body, to aid debugging a 503.
- One-line doc on `SetReadinessChecks` that it must be called before serving (consistent with the other `Set*` setters — safe today since wiring runs before `ListenAndServe`).

Merging.

Issue #606, opened 2026-08-18
A0.1 · Canonical agent identity binding (server-authoritative; fixes D1)


nghiadaulau:
> Phase **A0** (Protocol & Trust Invariants) · child of #594 · **P0** · fixes defect **D1**.

## Goal
Make agent identity **server-authoritative**. `TenantID`, `AgentID`, `AssetID`, `AgentSessionID` come from the authenticated credential + a server-side agent↔asset binding — never from a request body or display name.

## The bug this fixes (D1)
`startDetection` derives identity from the **display name**:
```
host  := shared.ID(r.cfg.name)              // cmd/synapse-agent/detect.go:35
agent := shared.ID("agent:" + r.cfg.name)   // cmd/synapse-agent/detect.go:36
```
and `run()` calls `startDetection(ctx)` **without the credential** (the enrolled canonical AgentID obtained at `main.go:145` is passed only to `cycle`, `main.go:151-153`). So detection AgentID ≠ enrolled AgentID: evidence can be mis-attributed, renaming an agent forges a new data-plane identity, two hosts sharing a name collide, and signing-key lookup by enrolled AgentID can miss.

## Scope
- Pass the enrolled `Credential` (canonical `AgentID`) into `startDetection`; the detection engine + telemetry sensor use the **credential's** AgentID/HostID/AssetID, not `cfg.name`.
- Introduce/confirm a server-side **`AssetBinding`** (agent → canonical asset); resolve `AssetID` from it, never from the agent.
- Telemetry **and** detection ingest must verify the batch's `AgentID`/`HostID` against the authenticated context (today only tenant is checked — `telemetry.Service.Ingest`).
- Add a stable **`AgentSessionID`** (per enrolled session) to the identity set.
- The agent may still send display name / hostname / machine-id / local fingerprint — but only for **reconciliation**, never as security identity.

## Acceptance / exit
- [ ] Renaming an agent (changing `cfg.name`) does **not** change its data-plane identity.
- [ ] Ingest **rejects** a telemetry/detection batch whose `AgentID`/`HostID` ≠ the authenticated agent.
- [ ] Detections and telemetry carry the **enrolled canonical AgentID**, with a test proving detect.go no longer derives identity from the display name.
- [ ] `AssetID` is resolved from the server-side binding.

## Dependencies
None — immediately actionable (real existing bug). Unblocks **A0.2** (key↔AgentID), **A0.4**, **A0.5**.

[...]

Issue #594, opened 2026-08-17
Epic: Security data plane — from fleet agent to governed EDR (observe → detect → ship → correlate → investigate → respond → verify)


nghiadaulau:
## Status

**APPROVE architecture direction · REQUEST CHANGES before freezing the implementation contract and splitting.** v3 folds in a static source review that found the current wiring is missing several **protocol invariants** (not ordinary implementation detail): canonical identity binding, signing-key lifecycle, versioned event/batch protocol, durable delivery/replay/gap semantics, atomic evidence persistence, and distributed idempotent response. The premature Phase-A child issues (#595–#602) were **closed**; Phase A is re-planned around a new **Phase A0 — Protocol & Trust Invariants** and will only be re-split once A0's contract lands.

This epic is now the **contract**, not just a vision. Completing it as written would yield a governed, coverage-honest EDR; being *superior* is decided by detection efficacy, false-positive rate, kernel compatibility, overhead, tamper resistance, event-loss rate, response reliability and investigation UX — so those become first-class workstreams and measurable release gates below, not afterthoughts.

> **v4** adds the non-negotiable **design principles** (led by *zero app instrumentation* — no SDK/library/restart, unlike OTel-SDK), sharpens **portability** (CO-RE/BTF; RHEL/OpenShift/RHEL CoreOS/CRI-O/arm64 + the OpenShift **SCC** requirement), and makes **deployment automation** and **continuous-exposure (CVE) fusion** explicit.
>
> **2026-08-18 update — A0 contract landed; the re-split is done.** All six A0 protocol/trust contracts merged (PRs #613–#618). Phase A (A1–A8), the cross-cutting workstreams, and Phases B/C/D/E are now split into detailed child issues (#622–#640), each carrying its target/tests/acceptance and (where relevant) the specific A0 implementation tail it inherits. See **Tracking** below. The A0 children (#606–#611) stay open until their A2/A3/A4 impl tails land.

## Vision (unchanged)

> **observe** system state → **detect** locally → **ship** (telemetry + detections) → **correlate** into incidents → **investigate** → **respond** → **verify** → **re-score**

Deterministic core runs fully **without AI**; AI is strictly **propose-only** on top (explain / correlate / prioritize), never in the detect/enforce decision.

---

## Design principles (non-negotiable)

[...]

PR #613, opened 2026-08-18
fix(fleet-agent): tag detections with the canonical AgentID, not the display name

nghiadaulau:
## What

Fix defect **D1** (agent-side) from epic #594 / issue #606: the eBPF detection engine tagged its events and detections with an identity derived from the **mutable display name** instead of the **server-issued canonical AgentID**.

Before:
```go
host  := shared.ID(r.cfg.name)              // display name
agent := shared.ID("agent:" + r.cfg.name)   // display name
...
r.startDetection(ctx)                        // credential never passed in
```

After: identity comes from the enrolled credential's canonical `AgentID`, via a pure, testable helper; `startDetection` receives the credential; and it **fails closed** (detection disabled) when no canonical AgentID is present.

## Why (the risk)

With identity taken from `cfg.name`, detections and their **hash-chained evidence** could be mis-attributed:
- renaming an agent (changing `-name` / `SYNAPSE_AGENT_NAME`) forged a **new** data-plane identity;
- two hosts sharing a display name **collided**;
- a signing-key lookup keyed by the enrolled AgentID could **miss**.

The display name is still sent at enrolment/heartbeat — but only for **reconciliation**, never as security identity (unchanged).

## Scope

This is the **agent-side** half of #606. The server-side ingest verification ("reject a batch whose `AgentID`/`HostID` ≠ the authenticated agent") is intentionally **deferred to A3/A4**, when a telemetry/detection ingest endpoint actually exists to enforce it — there is no live ingest path to harden yet. So #606 stays open; this PR closes the identity-forgery vector at the source.

## Tests

- `TestDetectionIdentityUsesCanonicalAgentID` — identity == the canonical enrolled AgentID for both host and agent tags, and is **not** the old `agent:`-prefixed display-name form.
- `TestDetectionIdentityFailsClosedWithoutAgentID` — empty / whitespace-only AgentID → `ok=false` (detection disabled), no display-name fallback.

## Verification (local)

- `go build ./...` · `go vet ./cmd/synapse-agent/...` — clean
- `go test -count=1 ./cmd/synapse-agent/...` — pass
- `golangci-lint run ./cmd/synapse-agent/...` — 0 issues
- `gofmt -l` — clean
- Reviewed for clean-architecture + security (agent-side identity forgery vector closed; no secret logged; fail-closed).

Refs #606 · part of #594.

### DIFF
--- a/cmd/synapse-agent/detect.go
+++ b/cmd/synapse-agent/detect.go
@@ -15,14 +15,33 @@ import (
 	"github.com/KKloudTarus/synapse-ce/internal/domain/shared"
 	"github.com/KKloudTarus/synapse-ce/internal/infrastructure/detectsink"
 	"github.com/KKloudTarus/synapse-ce/internal/infrastructure/ebpf"
+	"github.com/KKloudTarus/synapse-ce/internal/infrastructure/fleetclient"
 	detectuc "github.com/KKloudTarus/synapse-ce/internal/usecase/fleet/detect"
 )
+
+// detectionIdentity derives the canonical (host, agent) identity the detection engine tags its events
+// and detections with, from the ENROLLED credential — never from the mutable display name (cfg.name).
+//
+// This is the D1 fix (#606): the server issues a canonical AgentID at enrolment and resolves the asset
+// binding from it, so renaming an agent must not forge a new data-plane identity, two hosts sharing a
+// display name must not collide, and signing-key lookup by enrolled AgentID must not miss. The agent
+// runs one detection sensor for its own host, so that single canonical AgentID is the identity for both
+// the host and agent tags here; the control plane remains authoritative and binds the asset on ingest.
+// Returns ok=false when the credential carries no AgentID, so detection fails closed rather than
+// starting under an empty or attacker-influenced identity.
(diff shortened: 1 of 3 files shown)


nghiadaulau:
### Verification (repo Actions is disabled → CI replicated locally)

- `go build ./...` (host) + **cross-build** `GOOS=windows` and `GOOS=linux` for `./cmd/synapse-agent/...` — all OK (the change is platform-neutral; the Windows/distroless CI jobs would pass).
- `go vet ./cmd/synapse-agent/...` — clean · `gofmt -l` — clean.
- `golangci-lint run` (v2.12.2, the CI-pinned version) — **0 issues**.
- `go test -count=1 ./cmd/synapse-agent/...` — pass, incl. the two new identity tests.
- `mergeable_state = clean`; GitGuardian secret scan — success.

### Review

- **go-arch-reviewer: PASS** — dependency rule + Go idioms clean; single caller updated; tests prove the fix (identity = canonical AgentID, not the display name).
- **security-auditor: APPROVE** — the D1 identity-forgery vector is closed on the agent side; fail-closed on empty AgentID; no secret (`Token`) logged; deferring server-side ingest verification to A3/A4 is safe and adds no new surface.

Merging.

PR #619, opened 2026-08-18
feat: add Prometheus metrics and access logging

### LINKED ISSUE #584: Observability: opt-in Prometheus /metrics endpoint (+ optional OTel exporter)
> Part of #583 (production-readiness epic). Item **A1** · Priority **P0** · Group A (Operations & Observability).

## Problem
There is no application-level metrics surface. The server exposes no `/metrics` endpoint, imports no `promhttp`, and does not initialize OpenTelemetry at any composition root — the `go.opentelemetry.io` modules in `go.mod` are all `// indirect` (transitive, unused by us). No environment variable configures telemetry. The fleet blue-team guide notes a telemetry-store boundary exists, but the operator-facing wiring is not connected.

## Evidence
- `E-OPS-03` — `go.mod` (OTel entries all `// indirect`), `docs/guide/fleet-blue-team.md`.
- Verified on `main`: `grep -rn 'promhttp\|"/metrics"' internal/ cmd/` returns nothing.

## Impact
Latency, error rate, request throughput, job-queue depth, and resource usage cannot be observed from the application. Every operational signal must be reconstructed from a reverse proxy, Docker stats, or PostgreSQL. Incidents are detected late (risk `R-07`), and there is no data to drive SLOs (see C3 / #593).

## Proposed work
- Add an **opt-in** Prometheus `/metrics` endpoint exposing at minimum:
  - HTTP RED metrics: request count, latency histogram, and status class, labelled by **route pattern class** (not raw path — avoid high cardinality).
  - Job-queue metrics: queued depth, in-flight count, oldest-claim age.
  - Scan metrics: scan duration histogram and outcome counter (success / failed / blocked).
- Optionally wire an OTel exporter behind `SYNAPSE_OTEL_*` env (endpoint, headers, sampling), **default off**.
- Never emit secrets, tokens, tenant identifiers, or other high-cardinality/PII values as label values.
- Do **not** place `/metrics` behind the bearer-auth gate by default; bind it to a separate listener addr or leave protection to the proxy. Document the exposure decision.

## Acceptance criteria
- [ ] `/metrics` (or OTLP export) exposes HTTP RED + job-queue-depth + scan-outcome series.
- [ ] Metric names/labels documented in `docs/guide/configuration.md`; a Prometheus scrape-config example ships.
- [ ] No secret, token, or PII appears in any label (reviewed).
- [ ] The endpoint is opt-in (env-gated) and its exposure surface (separate addr / proxy protection) is documented.
- [ ] Behavioral test covers that the endpoint serves and that a representative metric increments.

## Dependencies
None. Complements A2 (#585 readiness) and feeds C3 (#593 benchmarks/SLO).

### LINKED ISSUE #586: Observability: structured request-logging (access-log) middleware
> Part of #583 (production-readiness epic). Item **A3** · Priority **P0** · Group A (Operations & Observability).

## Problem
The router middleware chain — path-normalize → authentication → acceptable-use-policy → route — has **no request-logging middleware**. There is no application-level access log (method, path, status, latency), so per-request observability depends entirely on a reverse proxy that may not exist in every topology.

## Evidence
- `E-OPS-02` — `internal/adapter/httpapi/router.go:580-595` (middleware chain, no access-log stage).

## Impact
No per-request audit or trace at the application tier. Incident timelines and abuse investigations depend on an external proxy; in the native and full-stack-Compose topologies there is often nothing in front of the API, so requests are effectively unlogged.

## Proposed work
- Add an access-log middleware that emits one structured `log/slog` JSON line per request with: HTTP method, **normalized route pattern** (not the raw path with ids), status code, latency, a request id, and the authenticated principal id when present.
- Reuse the existing multi-encoding redaction (`internal/platform/redact`) so no token/secret can appear (e.g. in a query string or `Authorization` echo). Do **not** log request or response bodies.
- Make it configurable: enable/disable and level, consistent with `SYNAPSE_LOG_LEVEL`.
- Generate/propagate a request id so app logs and access logs correlate.

## Acceptance criteria
- [ ] Every request emits exactly one structured access-log line (method, route pattern, status, latency, request id, principal id when authed).
- [ ] No secret, token, `Authorization` value, or body content ever appears in the line (verified via test using a known secret).
- [ ] Request id is present and correlatable with other app log lines.
- [ ] Enable/level is configurable and documented in the ops guide.

## Dependencies
None. Pairs with A1 (#584 metrics) for a complete observability baseline.

pho-veteran:
## Summary

- add an opt-in, separate Prometheus metrics listener with HTTP RED, aggregate durable-job queue, and SCA scan metrics
- add structured HTTP access logging with server-generated request IDs, bounded route/method fields, principal correlation, and streaming-safe response recording
- coordinate API and metrics listener shutdown and make publication mirror drift checks line-ending-neutral on Windows

## Security and privacy

- metrics use a private registry and bounded labels; no tenant, target, raw path, query, headers, body, or error text is exported
- PostgreSQL queue aggregates preserve forced tenant RLS and expose no tenant labels
- the unauthenticated metrics listener is disabled by default and bound to `127.0.0.1:9090` by default; deployment docs require loopback or a private scrape network
- access logs never record bearer tokens or request bodies

## Verification

- `go build ./...`
- `go vet ./...`
- `make test`
- focused race tests for HTTP observability, Prometheus collectors, and SCA lifecycle
- `make typecheck`
- `cd web && pnpm build`
- runtime smoke test of API, `/metrics`, request IDs, and token non-disclosure
- Go architecture and security reviews with focused confirmation passes

Closes #584
Closes #586

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -26,6 +26,16 @@ SYNAPSE_HTTP_ADDR=:8080
 SYNAPSE_ENV=development
 SYNAPSE_LOG_LEVEL=info
 SYNAPSE_SINGLE_TENANT=true
+# Prometheus /metrics on a SEPARATE, uninstrumented, non-bearer-protected listener.
+# Off by default; when enabled the listener stays loopback-only unless you widen it –
+# do not expose it beyond a private scrape network. See docs/guide/configuration.md
+# for the metric names and the label/privacy policy.
+SYNAPSE_METRICS_ENABLED=false
+SYNAPSE_METRICS_ADDR=127.0.0.1:9090
+# One structured "http access" log event per request (method, matched route, status,
+# latency, request id, and the resolved principal id when authenticated). Never logs
+# raw paths, queries, headers, bodies, or secrets. On by default.
+SYNAPSE_ACCESS_LOG_ENABLED=true
 # Acceptable-Use Policy version operators must accept on first run.
 SYNAPSE_AUP_VERSION=1.0
 # File-backed paths used only in in-memory mode (no SYNAPSE_DB_DSN).
--- a/cmd/synapse-api/main.go
+++ b/cmd/synapse-api/main.go
@@ -13,6 +13,7 @@ import (
 	"crypto/rand"
 	"fmt"
 	"log/slog"
+	"net"
 	"net/http"
 	"os"
 	"os/signal"
@@ -194,6 +195,29 @@ func requireJudgmentsOrSkip(log *slog.Logger, hasJudgment bool, envKey, name str
 	}
 	log.Warn(name + " auto-skipped: SYNAPSE_JUDGMENTS_ENABLED is off (it mints judgments)")
 	return false
+}
+
+// metricsAddrIsLoopback reports whether addr binds only to a loopback interface. The
+// metrics listener is intentionally unauthenticated, so a non-loopback bind exposes
+// aggregate operational metrics to anything that can reach it; callers use this to
+// decide whether to warn. It fails loud (returns false, i.e. "warn") on anything it
+// cannot confidently classify as loopback-only.
+func metricsAddrIsLoopback(addr string) bool {
+	host, _, err := net.SplitHostPort(addr)
+	if err != nil {
+		return false
+	}
+	if host == "" {
+		return false // empty host binds all interfaces
+	}
+	if host == "localhost" {
+		return true
+	}
+	ip := net.ParseIP(host)
+	if ip == nil {
+		return false
+	}
+	return ip.IsLoopback()
 }
 
 func migrationDSNForStartup(cfg config.Config) (string, error) {
@@ -1177,6 +1201,9 @@ func main() {
 		metrics = observability.New(queueReader)
 		httpObserver = metrics
 		scaService.SetObserver(metrics)
+		if !metricsAddrIsLoopback(cfg.MetricsAddr) {
+			log.Warn("metrics listener is bound to a non-loopback address; it is unauthenticated and exposes aggregate operational metrics to anything that can reach it", "addr", cfg.MetricsAddr)
+		}
 	}
 	router.SetObservability(cfg.AccessLogEnabled, httpObserver)
 	vulnerabilityRollout, err := vulnerabilityrollout.New(vulnerabilityrollout.Config{
--- a/cmd/synapse-api/main.go
+++ b/cmd/synapse-api/main.go
@@ -13,6 +13,7 @@ import (
 	"crypto/rand"
 	"fmt"
 	"log/slog"
+	"net/http"
 	"os"
 	"os/signal"
 	"path/filepath"
@@ -21,6 +22,7 @@ import (
 	"time"
 
 	"github.com/KKloudTarus/synapse-ce/internal/adapter/httpapi"
+	"github.com/KKloudTarus/synapse-ce/internal/adapter/observability"
 	"github.com/KKloudTarus/synapse-ce/internal/domain/agent"
 	ap "github.com/KKloudTarus/synapse-ce/internal/domain/attackpath"
 	"github.com/KKloudTarus/synapse-ce/internal/domain/cloudposture"
@@ -1162,6 +1164,21 @@ func main() {
 	if slaService != nil {
 		router.SetSLA(slaService)
 	}
+	// Metrics stay off by default and, when enabled, are exposed only on the separate
+	// loopback-by-default listener (never bearer-protected, never instrumented itself).
+	var metrics *observability.Collectors
+	var httpObserver httpapi.HTTPObserver // kept as a nil INTERFACE unless metrics is built
+	if cfg.MetricsEnabled {
+		queueReader, ok := vulnerabilityQueue.(ports.AggregateJobQueueStatsReader)
+		if !ok {
+			log.Error("metrics enabled but the configured job queue does not support aggregate stats")
+			os.Exit(1)
+		}
+		metrics = observability.New(queueReader)
+		httpObserver = metrics
+		scaService.SetObserver(metrics)
+	}
+	router.SetObservability(cfg.AccessLogEnabled, httpObserver)
 	vulnerabilityRollout, err := vulnerabilityrollout.New(vulnerabilityrollout.Config{
 		ProviderSync: cfg.VulnerabilityProviderSyncEnabled, OccurrenceWrites: cfg.VulnerabilityOccurrenceWritesEnabled,
 		FindingProjection: cfg.VulnerabilityFindingProjectionEnabled, Actions: cfg.VulnerabilityActionsEnabled,
@@ -2128,7 +2145,11 @@ func main() {
 		)
 	}
 
-	if err := httpserver.Run(ctx, cfg.HTTPAddr, router.Handler(), log); err != nil {
+	var metricsHandler http.Handler
+	if metrics != nil {
+		metricsHandler = metrics.Handler()
+	}
+	if err := httpserver.RunPair(ctx, cfg.HTTPAddr, router.Handler(), cfg.MetricsAddr, metricsHandler, log); err != nil {
 		log.Error("server error", "err", err)
 		os.Exit(1)
 	}
--- a/cmd/synapse-api/main_test.go
+++ b/cmd/synapse-api/main_test.go
@@ -49,3 +49,26 @@ func TestMigrationDSNForStartup(t *testing.T) {
 		})
 	}
 }
+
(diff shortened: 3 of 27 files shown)


nghiadaulau (changes requested):
Thanks — this is a strong, privacy-conscious contribution. Reviewed strictly (clean-architecture + security). The observability design itself is excellent, but there is **one blocking issue** that must be fixed before merge because it touches a golden-rule invariant, plus a few minor follow-ups.

## 🔴 Blocking — evidence double-seal / no-double-exec regression (and it's scope creep)

`internal/usecase/sca/service.go` — `runScanJob` now **returns** the terminal `s.jobs.Save` error and `RunScanJob` propagates it. A propagated error makes the durable queue `Fail` and **redeliver** the job. On redelivery the idempotency guard checks `latest.Status`, but the failed `Save` left the status `Running`, so the guard does **not** skip — the pipeline re-runs and seals a **duplicate `scan` evidence link + a phantom `ScanRun` row**. That's exactly the hazard the guard's own comment warns about, and it violates Golden Rule 5/6 (hash-chained evidence, no-double-exec). Previously the terminal-Save error was non-fatal (`RunScanJob` returned nil; evidence sealed once, the stale row was reconciled later by `SweepStaleScans`).

This is also unrelated to the PR's stated purpose (metrics + access logging), so it shouldn't ride along silently. Please either:
- keep the terminal `Save` **non-fatal** for redelivery (restore prior behavior), OR
- make evidence sealing **idempotent per `job.ID`** before enabling retry,

and please confirm the change against the `make load` "no-double-exec + chain-intact" SLO.

## 🟡 Minor (non-blocking, but worth addressing)

- **Metrics listener on a non-loopback addr is silently unauthenticated.** Default is safe (`127.0.0.1:9090`, off by default) and the `/metrics` registry is private (no PII/secrets/tenant ids, bounded labels — all verified). But if an operator sets `SYNAPSE_METRICS_ADDR` to a non-loopback interface, aggregate operational metrics are exposed with no auth. Emit a startup WARN when `MetricsEnabled && MetricsAddr` is not loopback.
- **`queueCollector.Collect` drops all three `synapse_job_queue_*` gauges on the first `AggregateJobQueueStats` error, with no signal** — queue depth silently vanishes from a scrape exactly when it may matter. Log at debug or emit a `scrape_error` counter.
- **`AggregateJobQueueStats` is O(tenants) per scrape** (`SELECT id FROM tenants` + one RLS-scoped `Stats` tx per tenant, under a 1s cap). Correct for tenant isolation and fine for CE's single-tenant default, but it will time out (and drop the gauges, per above) as tenant count grows — add a comment acknowledging the ceiling, or a single aggregate under a system role / a scrape-interval cache.
- **`observeGateOutcome(started time.Time, err error)`** never uses `started` — drop the dead parameter.

## ✅ What's genuinely good (verified)

- **Dependency rule PASS**: `usecase/sca` depends only on the new `ports.SCAObserver` interface — no prometheus/adapter import leaks into usecase or domain; the concrete `client_golang` is confined to `internal/adapter/observability`; all wiring is at the composition root.
- **Label hygiene / no cardinality blow-up**: `route` uses the matched ServeMux pattern (or `unmatched`), `method` normalized to a fixed set, `status_class` collapsed to `Nxx` — with dedicated tests. No tenant/engagement/target ids in labels.
- **Access-log privacy**: only method/route/status/duration/request_id/principal_id — no raw path, query, headers, or body; negative tests assert the bearer token / query token / body don't appear; request-id is server-minted (no CRLF/log-injection or spoof surface).
- **Private registry** (no default Go/process collectors), separate loopback listener via `RunPair` with sibling-cancel-on-failure, `prometheus/client_golang v1.23.2` pinned (reputable). Thorough tests throughout.

Happy to re-review as soon as the blocking `runScanJob` change is addressed. (Note: CI/Actions is disabled on this repo, so this was validated locally — `go build`/`go vet` pass on the changed packages.)

pho-veteran:
Thanks for the review. Rebased onto `main` (`ce7560e`) and addressed all four findings. Two commits: the original feature, then `d83849c` with the fixes.

**1. Terminal `Save` failure must not redeliver (blocking)** — you're right, and the fix goes further than restoring the old behavior. `runScanJob` now logs and returns `nil` on a terminal `s.jobs.Save` failure. Returning the error made the queue redeliver a scan that had already executed, and because the failed save left the job at `Status=Running`, the idempotency guard did not skip — duplicate `scan` evidence link plus a phantom `ScanRun` row.

Worth noting why the branch also does *not* observe an outcome: the job is left stranded at `running`, and `SweepStaleScans` is what finalizes it and counts the eventual `failed`. Observing here as well would double-count. So the stranded-then-swept path yields exactly one outcome.

The existing test was inverted rather than deleted — `TestRunScanJobDoesNotRedeliverOnTerminalSaveFailure` now asserts `RunScanJob` returns nil, observes nothing, and leaves the job non-terminal for the sweeper.

**2. Non-loopback metrics bind** — startup now emits a WARN when `SYNAPSE_METRICS_ENABLED` is set and `SYNAPSE_METRICS_ADDR` is not loopback-bound, via a new `metricsAddrIsLoopback` helper that fails loud: unparseable address, empty host (`:9090`), and `0.0.0.0` all warn. Kept as a warning rather than a rejection since a non-loopback bind is a legitimate choice for a private container scrape network. Table-driven test covers v4/v6 loopback, `localhost`, all-interfaces, a routable IP, and a malformed address.

**3. Silent queue-gauge drop** — added `synapse_job_queue_scrape_errors_total` (counter, no labels) on the same private registry, incremented in the error branch and always emitted so it is present from `0`. The three gauges are still omitted on failure rather than reported stale — a wrong queue depth is worse than an absent one, and the counter now makes the absence explainable. New test asserts gauges absent and counter `1` on a failed read; the existing one-read-per-gather, bounded-deadline, and no-tenant-label tests are unchanged.

**4. Dead parameter** — dropped from `observeGateOutcome`; this orphaned a `started := s.clock.Now()` in `StartScanWithOptions`, removed too. `observeSyncTerminal` still uses its `started`.

Also documented `synapse_job_queue_scrape_errors_total` and the non-loopback warning in `docs/guide/configuration.md`.

Rebase resolutions, for the record: took `main`'s dependency bumps in `go.mod`/`go.sum` and re-pinned `prometheus/client_golang` to v1.23.2 as a direct requirement (a plain `--ours` resolution silently downgraded it to the v1.12.1 minimum); in `docs/guide/deployment.md` kept your "Liveness and readiness probes" heading from #612 with the new metrics section above it.

Verification on the rebased tree: `go build ./...`, `go vet ./...`, full `go test ./...` (zero failures), `-race` on sca/observability/httpapi/synapse-api, and `pnpm typecheck` + `pnpm build` all green.

nghiadaulau (approved):
**Approving — the blocking issue is fully resolved and all follow-ups landed. Re-reviewed strictly + validated locally (CI is disabled).**

The follow-up commit `d83849cc` closes the blocker exactly right:

- **Evidence double-seal / no-double-exec — CLOSED.** `runScanJob` no longer propagates the terminal `ScanJob.Save` error; it logs and returns `nil`, so a save failure can never make the durable queue redeliver. The stranded `Status=Running` job is finalized solely by `SweepStaleScans`, and the `"failed"` outcome is counted **exactly once** (the save-failure branch returns before `observeSyncTerminal`, and `SweepStaleScans` now records only after its own `Save` succeeds). Locked in by `TestRunScanJobDoesNotRedeliverOnTerminalSaveFailure`, `TestRunScanJobIdempotentSkipDoesNotDoubleObserve`, `TestFailStrandedScanJobObservesFailedOnce`, `TestStartScanEnqueueFailureObservesFailedOnce` — all green under `-race`.

Minors all addressed: startup WARN when the metrics listener is enabled on a non-loopback address (`metricsAddrIsLoopback` + table test), `synapse_job_queue_scrape_errors_total` so a failed aggregate read is observable instead of the gauges silently vanishing, and the dead `started` param dropped from `observeGateOutcome`.

Verified merged onto current `main` (post-A0): `go build ./...`, `go vet`, and the touched-package tests (`observability`, `httpapi`, `sca`, `httpserver`, `config`, `cmd/synapse-api`) all pass. Dependency rule remains clean (no prometheus/infra leak into usecase/domain; private registry; loopback-default separate listener); label hygiene + access-log privacy verified in the first pass.

Two optional, non-gating notes for a future follow-up: `RunPair` couples the listeners (a metrics-port bind conflict takes down the API — intentional/documented, but debatable for an opt-in listener); and `SweepStaleScans` now returns a partial-`n`+error on a per-job save failure (more correct — counts only on persistence).

Nice work addressing the review thoroughly. Merging.

PR #620, opened 2026-08-18
fix(httpapi,postgres): harden readiness concurrency and telemetry gap keying

XUanhoa04:
### Summary

This PR introduces two clean, self-contained robustness improvements:

1. **`httpapi` Concurrency Safety**: Protects `Router.readiness.checks` map with a `sync.RWMutex`. `SetReadinessChecks` takes a write lock while `ready()` snapshots the configuration under a brief read lock before dispatching probe checks concurrently outside the critical section.
2. **`postgres/telemetry_repo` Typed Composite Key**: Replaces the string concatenation delimiter pattern (`host + "\x00" + class`) with a typed `telemetryKey` struct for grouping sequence streams during retro-hunt queries. This eliminates unnecessary per-row string allocations and removes `splitTelemetryKey`.

### Changes

- `internal/adapter/httpapi/readiness.go`: add `sync.RWMutex` to `readinessConfig`, guard `SetReadinessChecks` (write lock) and `ready()` (read lock snapshot).
- `internal/adapter/httpapi/readiness_test.go`: add `TestReadinessSetChecksConcurrentRace`.
- `internal/infrastructure/persistence/postgres/telemetry_repo.go`: define `telemetryKey struct`, update `Query` and `telemetryGaps` to use struct keys in map, remove `splitTelemetryKey`.
- `internal/infrastructure/persistence/postgres/telemetry_gaps_test.go`: add unit tests for `telemetryGaps`, `dedupSorted`, and multi-class key isolation.

### Verification

- Unit tests pass with `-count=1`
- Full repository build succeeds (`go build ./...`)

### DIFF
--- a/internal/adapter/httpapi/readiness.go
+++ b/internal/adapter/httpapi/readiness.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"net/http"
 	"strings"
+	"sync"
 	"time"
 )
 
@@ -14,6 +15,7 @@ const defaultReadinessTimeout = 2 * time.Second
 type ReadinessCheck func(context.Context) error
 
 type readinessConfig struct {
+	mu      sync.RWMutex
 	checks  map[string]ReadinessCheck
 	timeout time.Duration
 }
@@ -25,7 +27,10 @@ type readinessResponse struct {
 
 // SetReadinessChecks replaces the dependency checks used by GET /readyz. It copies the map so
 // startup wiring cannot mutate the live probe configuration after the server begins serving.
+// Protected by write-lock against concurrent probe execution (race safety).
 func (rt *Router) SetReadinessChecks(checks map[string]ReadinessCheck) {
+	rt.readiness.mu.Lock()
+	defer rt.readiness.mu.Unlock()
 	rt.readiness.checks = make(map[string]ReadinessCheck, len(checks))
 	for name, check := range checks {
 		if name = strings.TrimSpace(name); name != "" && check != nil {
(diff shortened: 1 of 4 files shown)


nghiadaulau (approved):
✅ **Approve + merge.** Strict review (go-arch + security subagents, local verify; CI intentionally off).

**Correctness / architecture (go-arch): PASS**
- `readinessConfig` gains a `sync.RWMutex`; `SetReadinessChecks` write-locks the rebuild, `ready()` read-locks then **snapshots the map + timeout and releases before running any probe** — no lock held across probe execution, no lock-copy bug (`Router` is always `*Router`, pointer receivers, `go vet` copylocks-clean).
- Typed `telemetryKey{host, class}` map key replaces the `host+"\x00"+class` string concat — infra→domain import is inward (allowed), `splitTelemetryKey` removed with no dangling imports.

**Security: CLEAN (net positive)**
- Fixes a genuine data race on `rt.readiness.checks` (bare header read vs concurrent replace).
- The snapshot-then-unlock pattern means a slow/hanging check can't wedge the probe → avoids a lock-contention DoS rather than adding one; `/readyz` still exposes only check-name + `ok`/`failed` (never the error string), `Cache-Control: no-store`.
- The typed key **removes** the old NUL-delimiter collision (`host="a\x00b",class="c"` vs `host="a",class="b\x00c"` both collapsed to one bucket); struct equality can never collide across content. Tenant isolation (`WithContextTenant`/RLS) unchanged; no secrets logged.

**Local verification:** `go build` + `go vet` clean; `-race` readiness test green; postgres gap/dedup/key unit tests green.

**One fixup applied on the branch** (`56f0266`): `gofmt` removed a trailing blank line in `readiness_test.go` and aligned a map-entry comment in `telemetry_gaps_test.go` so `make lint` is clean — pure whitespace, no behavior change.

Tests are well-targeted (race regression + gap/dedup/class-isolation). Merging.

Issue #591, opened 2026-08-17
Security: enforce sandbox-on in production and prove startup fail-closed


nghiadaulau:
> Part of #583 (production-readiness epic). Item **C1** · Priority **P0** · Group C (Security Posture & Validation).

## Problem
The full-stack demo runs with the hardened sandbox **off**: `deploy/docker-compose.full.yml` sets `SYNAPSE_SANDBOX_ENABLED: "false"`, and the guide states the profile is not for production. The hardened sandbox is implemented and fail-closed **by design**, but it has not been exercised on a Linux host with bubblewrap/seccomp/cgroups in the review — so real isolation and fail-closed behavior are graded `P` (partial), not proven.

## Evidence
- `E-DEP-03` — `deploy/docker-compose.full.yml:147-148`, `docs/guide/deployment.md`.
- `E-REQ-02` (sandbox is Linux-only and fail-closed by design); risk `R-11`; review §15.8.

## Impact
Running production with the sandbox off means analyzers execute on **untrusted source** with no hardened isolation (`R-11`) — the exact threat the sandbox exists to contain.

## Proposed work
- As part of the production manifest (B3, #590), require the sandbox on (`SYNAPSE_SANDBOX_ENABLED=true`) on a Linux host.
- Add a verification step / test that proves startup **fail-closed**: on a Linux host with sandbox enabled, removing `bwrap` must make the API **refuse to start** (not silently downgrade to unsandboxed execution).
- Capture a log/artifact demonstrating the sandbox actually blocks out-of-policy filesystem and network access during a real run.

## Acceptance criteria
- [ ] The production manifest enforces sandbox on (Linux).
- [ ] A documented run shows the API fails closed when `bwrap` is absent (does not run unsandboxed).
- [ ] An isolation test demonstrates filesystem and network confinement holds for a tool run under the sandbox.

## Dependencies
Realized through **B3** (#590) production manifest. Corresponds to review acceptance item P0-1.

pho-veteran:
I'm working on this!

pho-veteran:
## Production sandbox finding from the EPIC #583 staging run

PR #648 implements the production image/chart posture and preserves fail-closed startup, but the **positive Bubblewrap execution gate remains open**.

### Current failure

The production sandbox probe runs `/bin/true` through the same isolation shape required for tools:

- `bwrap --unshare-all`
- `--die-with-parent`
- `--new-session`
- read-only root binding
- seccomp supplied on FD 3
[...]

Issue #590, opened 2026-08-17
Deploy: production deployment manifest (TLS, sandbox on, hardened, gated migration)


nghiadaulau:
> Part of #583 (production-readiness epic). Item **B3** · Priority **P0** · Group B (Scale & Deployment Architecture).

## Problem
The `production` topology exists only as prose and a checklist in `docs/guide/deployment.md` — there is no deployable manifest. Concretely:
- No reverse proxy / ingress / TLS terminator / load balancer in any manifest; the application does not open an HTTPS listener (`http.Server` + `ListenAndServe`, no `ServeTLS`).
- The k8s manifest ships only the read-only cluster-agent (Deployment + ClusterRole + PVC) — no API Service/Ingress.
- Both Compose files publish data-service ports (`5432`, `9000`, `9001`) to the host and declare **no restart policy**.
- Two external images (`minio/minio`, `minio/mc`) use `latest` (no pinned digest).

## Evidence
- `E-DEP-05` — `internal/platform/httpserver/server.go:14-24` (no TLS), `docs/guide/deployment.md`.
- `E-DEP-02` (Compose ports/no-restart/latest), `E-DEP-04` (k8s = cluster-agent only).
- Risks `R-03` (unpinned images), `R-04` (no self-recovery), `R-12` (token in cleartext without TLS), `R-14` (data services exposed to host).

## Impact
There is no supported, reproducible production install. If `8080` is exposed without TLS, bearer tokens travel in cleartext (`R-12`); data services are reachable from the host (`R-14`); nothing self-recovers after a host reboot (`R-04`); and `latest` tags make builds non-reproducible and silently drift (`R-03`).

## Proposed work
Ship a production-grade deployment manifest (Helm chart, or hardened k8s manifests / a `docker-compose.prod.yml`) that:
- Terminates TLS in front of the API and exposes **only** the API behind the proxy — **no** data-service host ports on an untrusted network.
- Sets `SYNAPSE_SANDBOX_ENABLED=true` on a Linux host (see C1, #591).
- Sets a restart policy / relies on an orchestrator with self-healing.
- Pins external images by **digest**, and sets `SYNAPSE_TOOL_HASHES` to concrete hashes (no trust-on-first-use).
- Sources the four durable secrets from a secret store, not env files baked into an image.
- Runs migration as the gated pre-step from **B1** (#588), not as a per-instance startup side-effect.

## Acceptance criteria
[...]

PR #621, opened 2026-08-18
fix: gate migrations and production sandbox startup

### LINKED ISSUE #588: Scale: run migrations as a gated single-migrator pre-step (lock before migrate)
> Part of #583 (production-readiness epic). Item **B1** · Priority **P0** · Group B (Scale & Deployment Architecture).

## Problem
`synapse-api`, `synapse-worker`, and `synapse-mcp` all run database migrations automatically at startup, and each takes a per-role **session advisory lock** to block a second instance of the same role. But the lock is acquired **after** migrations have already run — so the migration step itself is unprotected. Two instances starting concurrently can both enter migration. A code comment already asks operators to run a single instance, or split migration into a separate job, until horizontal scaling is supported.

## Evidence
- `E-ARCH-03` — `cmd/synapse-api/main.go:332-347` and `:363-377`, `internal/infrastructure/persistence/postgres/db.go:99-120`; also `cmd/synapse-worker/main.go:83-109`, `cmd/synapse-mcp/main.go:55-76`.
- Risks `R-01` (concurrent migration → undefined schema), `R-02` (migration exceeding the healthcheck window → mid-migration restart).

## Impact
- **R-01:** concurrent startup can leave the schema in an undefined state.
- **R-02:** 103 migrations on a data-heavy DB may exceed the API healthcheck window (~120s: interval `10s` × retry `12`), so the orchestrator can restart the container mid-migration. Migration time on real data is unmeasured (`U`).

## Proposed work
- Extract migrations into a dedicated one-shot step that holds its **own** advisory lock and runs to completion **before** any app instance starts — e.g. a `synapse-migrate` command or an init-job/init-container. App instances then either skip auto-migrate, or take the lock **before** migrating (closing the ordering gap in-process).
- Measure migration time on a representative data volume; either widen the API healthcheck window accordingly or rely on the gated pre-step so healthcheck never races migration.

## Acceptance criteria
- [ ] Two instances starting concurrently cannot both migrate (proven by test or by design: lock precedes migrate).
- [ ] A gated migration pre-step (command / init-job) exists and is wired into the production manifest (B3, #590).
- [ ] Migration time on a representative data volume is measured and recorded.
- [ ] The API healthcheck no longer races migration (documented interaction with A2, #585).

## Dependencies
Precedes **B2** (#589): removing the API single-instance cap is only safe once migration is no longer a per-instance startup side-effect. Feeds **B3** (#590) production manifest and interacts with **A2** (#585) readiness.

pho-veteran:
## Summary

Make database migrations a gated, single-owner startup operation and fail closed when production sandboxing is disabled. This closes #588 and implements the code/startup portion of #591; deployment manifests and Linux isolation evidence remain tracked by #590 and #591.

## Changes

- Add the one-shot `synapse-migrate` command for production migrations using a dedicated blocking PostgreSQL advisory lock.
- Require production long-running services to set `SYNAPSE_DB_AUTO_MIGRATE=false`, keeping owner/DDL credentials out of API, worker, MCP, and advisory-sync startup paths.
- Validate the exact latest state of every embedded Goose migration; API exposes stale state through `/readyz`, while worker and MCP refuse startup.
- Require production API and worker startup to enable sandboxing, preserving fail-closed behavior when bubblewrap is unavailable.
- Document migration ownership, runtime-role separation, readiness behavior, and production sandbox requirements.
- Add unit and PostgreSQL integration coverage for migration-state validation, role separation, and lock contention.

## Checklist

### Documentation

Drift guards run in `make test`, so these usually fail loudly rather than silently. Confirm anyway:

- [x] New or changed `SYNAPSE_*` variables are in `docs/guide/configuration.md`, and
      security-relevant ones state the risk of a careless value
- [x] New or changed CLI flags, subcommands, and exit codes are in `docs/guide/cli.md` (N/A: the new composition-root command is documented in the architecture, configuration, and quickstart guides; no existing CLI subcommand or exit code changed)
- [x] New HTTP routes are described in `api/openapi.yaml` (do not grow
      `api/testdata/openapi-coverage-debt.txt`) (N/A: no HTTP routes changed)
- [x] New guides are in `mkdocs.yml` nav, so the site actually publishes them (N/A: no new guide added)
- [x] Claims match shipped behavior: no unbuilt artifact, unimplemented setting, or
      ingest-only format described as an export
- [x] Screenshots retaken if the documented UI changed, with no tokens, keys, or real
      customer data captured (N/A: no UI or screenshots changed)

### Validation

- `cd web && pnpm build`
- Fresh PostgreSQL 17 integration run: all 106 migrations applied successfully.
- `TestMigrateLockedWaitsForMigrationLock`: verified the contender waits in `pg_locks` and completes only after lock release.
- `TestPostgresReadinessChecks`: verified exact migration readiness and rejection of a latest `down` state.
- Architecture and security reviews completed, followed by one remediation batch and focused confirmation.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -61,6 +61,9 @@ SYNAPSE_DB_DSN=postgres://localhost:5432/synapse?sslmode=disable
 # Optional owner-level DSN used only for embedded migrations; keeps the runtime
 # identity least-privileged. Empty falls back to SYNAPSE_DB_DSN.
 # SYNAPSE_DB_MIGRATION_DSN=
+# Set false in production: synapse-migrate owns migrations. API exposes /readyz;
+# worker and MCP refuse startup until migrations are current.
+SYNAPSE_DB_AUTO_MIGRATE=true
 
 # =============================================================================
 # Evidence blob store – MinIO / S3 (empty endpoint ⇒ in-memory dev store)
@@ -175,8 +178,8 @@ SYNAPSE_SCAN_CACHE_ENABLED=true
 # =============================================================================
 # Recon & execution sandbox  (sandbox is REQUIRED in production)
 # =============================================================================
-# Run tool execution + acquisition in the bubblewrap sandbox. If set but
-# bubblewrap is unavailable, startup FAILS CLOSED (never runs unsandboxed).
+# Run tool execution + acquisition in the bubblewrap sandbox. Production requires
+# this to be true; if bubblewrap is unavailable, startup FAILS CLOSED.
 SYNAPSE_SANDBOX_ENABLED=false
 # Per-run cgroup limits (bytes / pid count).
 # SYNAPSE_SANDBOX_MEM_MAX=536870912    # 512 MiB
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -73,6 +73,10 @@ goose**. Two rules avoid the duplicate-version startup crash goose raises on col
   the later one rebases and renumbers. goose keys on the leading integer, so a duplicate is a
   hard startup panic, not a git merge conflict (the filenames differ), and a clean merge will not
   catch it.
+- **Use backward-compatible, phased schema changes.** Production rollouts migrate first, then
+  deploy application binaries. During that overlap an older API may serve only a database whose
+  additional migration versions are applied and strictly newer than its embedded maximum; it must
+  not depend on columns or constraints removed by the migration.
 - **Add a Postgres integration test** (`migration_00NN_test.go`) that calls `postgres.Migrate`,
   so a broken or duplicate version is caught locally.
 
--- a/cmd/synapse-api/main.go
+++ b/cmd/synapse-api/main.go
@@ -220,22 +220,6 @@ func metricsAddrIsLoopback(addr string) bool {
 	return ip.IsLoopback()
 }
 
-func migrationDSNForStartup(cfg config.Config) (string, error) {
-	migrationDSN := cfg.DBMigrationDSN
-	if migrationDSN == "" {
-		if cfg.IsProduction() {
-			return "", fmt.Errorf("SYNAPSE_DB_MIGRATION_DSN is required with SYNAPSE_DB_DSN outside development")
-		}
-		return cfg.DBDSN, nil
-	}
-	if cfg.IsProduction() {
-		if err := postgres.ValidateMigrationRoleSeparation(migrationDSN, cfg.DBDSN); err != nil {
-			return "", fmt.Errorf("validate migration and runtime database roles: %w", err)
-		}
-	}
-	return migrationDSN, nil
-}
-
 func main() {
 	cfg := config.Load()
 	if cfg.CSPMEnabled && !cfg.FleetAssetsEnabled {
@@ -248,6 +232,14 @@ func main() {
 	}
 	log := logging.New(cfg.LogLevel)
 	log.Info("starting synapse-api", "env", cfg.Environment, "single_tenant", cfg.SingleTenant)
+	if err := cfg.ValidateSandboxPosture(); err != nil {
+		log.Error("sandbox posture invalid", "err", err)
+		os.Exit(1)
+	}
+	if err := cfg.ValidateMigrationPosture(); err != nil {
+		log.Error("database migration posture invalid", "err", err)
+		os.Exit(1)
+	}
 
 	// Fail closed: no anonymous access. The token is never logged.
 	if cfg.APIToken == "" {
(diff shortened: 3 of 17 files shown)


nghiadaulau (changes requested):
Strict review — go-arch + security subagents, local verify on a fresh Postgres 17 (CI intentionally off). The design is excellent and **security is CLEAN**, but there is **one blocker** (a real, `-race`-provable data race that this PR introduces) plus two minor notes. Requesting changes on the blocker.

---

## 🔴 Blocker — data race on goose's package-global `baseFS`, newly introduced into the concurrent `/readyz` path

`embeddedMigrationVersions()` (`internal/infrastructure/persistence/postgres/db.go`) calls `goose.SetBaseFS(migrations.FS)` + `goose.CollectMigrations(...)` on **every readiness probe**:

- `cmd/synapse-api/main.go:385` registers `readinessChecks["migrations"] = func(ctx){ return postgres.CheckMigrationsReady(ctx, pool) }`
- `CheckMigrationsReady → checkMigrationsReady → embeddedMigrationVersions()` → `goose.SetBaseFS(...)` (mutates the unguarded goose package global `baseFS`)
- `internal/adapter/httpapi/readiness.go` runs **each check in its own goroutine**

So two overlapping `GET /readyz` requests run two `embeddedMigrationVersions` goroutines that race write/write (and read/write vs `CollectMigrations`) on the goose global. This is a regression: the previous `checkMigrationsReady` used `migrations.FS.ReadDir(".")` and touched no goose global.

Proof — a temporary 8-goroutine repro calling `embeddedMigrationVersions()` under `-race`:

```
WARNING: DATA RACE
  github.com/pressly/goose/v3.SetBaseFS()      goose.go:38
  github.com/pressly/goose/v3.SetBaseFS()      goose.go:38
WARNING: DATA RACE
  github.com/pressly/goose/v3.CollectMigrations()   migrate.go:176
  github.com/pressly/goose/v3.SetBaseFS()           goose.go:38
--- FAIL: TestZZZGooseBaseFSRaceRepro
```

The written value is always `migrations.FS` so the effect is benign, but it is a true data race per the Go memory model and `make load -race` (concurrent `/readyz`) will flag it. The integration tests don't hit `/readyz` concurrently, which is why they're green.

**Suggested fix** (either):
1. **Preferred** — derive the embedded version set from `migrations.FS` filenames (parse the `NNNN_` prefix) so the request path never touches a goose global; or
2. Cache it once behind a `sync.Once` (package-level `[]int64` computed on first call) so `SetBaseFS`/`CollectMigrations` run exactly once.

Startup `MigrateLocked`/`migrate()` may keep using goose directly — it is serial before the server serves traffic.

---

## 🟡 Minor 1 — `compareMigrationStates` "unexpected migration" rejects a forward schema during a rolling deploy
`db.go` final loop ("database contains unexpected migration %d"). The exact-set match now fails readiness if the DB has any version the running binary lacks. With the new model (a `synapse-migrate` job applies schema *before* the new binaries roll), an older replica still serving during the window sees the newer migration and flips `/readyz` → not-ready for the whole window. Moot pre-P5 single-instance, but stricter than the count-based check it replaces — worth a comment or a "DB superset is acceptable" relaxation before horizontal scale (E23). The missing/not-applied/duplicate checks are strict improvements — no concern.

## 🟡 Minor 2 — api vs worker/mcp asymmetry on stale schema when `DBAutoMigrate=false`
Worker (`cmd/synapse-worker/main.go`) and mcp (`cmd/synapse-mcp/main.go`) add a hard `CheckMigrationsReady` gate that exits on a stale schema; the api relies solely on `/readyz`. Defensible (api has a readiness endpoint, worker/mcp don't) — just confirm the "api comes up live-but-not-ready vs worker/mcp fail-fast" asymmetry is intended.

---

## ✅ Verified correct (no action)
- **Security: CLEAN.** No DSN/credential ever logged (pgx v5 `ParseConfigError` redacts passwords; all startup `log.Error`s carry static messages or DSN-free error chains). `ValidateMigrationPosture` forces `AUTO_MIGRATE=false` in prod so DDL/owner creds stay out of long-running services; role-separation validation is preserved on the sole prod-migration path (`synapse-migrate`). `ValidateSandboxPosture` **strengthens** the sandbox fail-closed (now an early unconditional check in api+worker, built on `IsProduction()` which fails closed on unknown env); mcp correctly omits it (read/propose-only). No bypass.
- **Advisory lock correct.** `SetMaxOpenConns(1)` keeps lock+goose+unlock on one session; acquisition bounded by the 2-min startup ctx; defer order unlock→Close, fresh 5s release ctx, and `db.Close()` releases the session lock if best-effort unlock fails. Dedicated lock key distinct from role singleton keys.
- **Dependency rule: PASS.** config methods add only stdlib `errors`; migration/lock logic in infra/postgres; `synapse-migrate` is a pure composition root.
- **Local verify green:** `go build ./...`, `go vet`, `gofmt -l` (clean), all config + postgres unit tests, and **both** PG integration tests on a fresh DB — `TestMigrateLockedWaitsForMigrationLock` (advisory-lock contention) and `TestPostgresReadinessChecks` (106 migrations, exact readiness) — PASS.

Fix the blocker (readiness-path goose global) and re-request; the rest is ready.

pho-veteran:
Addressed all requested changes in commit `9e118cd`.

- Removed Goose package-global state from migration readiness. Embedded versions are now derived directly from the immutable `migrations.FS`; parsing reuses Goose's pure `NumericComponent` semantics, adds deterministic sorting and duplicate/invalid-version checks, and has concurrent race coverage.
- Narrowly enabled migrate-first rolling deployments: an older binary accepts only applied database migrations strictly newer than its embedded maximum. Missing/down required migrations, divergent versions within known history, duplicate states, and newer down states still fail readiness.
- Confirmed and documented the intentional service behavior: API remains live but returns `503` from `/readyz` when migration readiness fails; worker and MCP refuse startup because they do not have a readiness endpoint that can gate work. Migration guidance now requires backward-compatible, phased schema changes.

Validation:
- Focused migration readiness/parser tests passed.
- `go test -race ./internal/infrastructure/persistence/postgres -run '^TestEmbeddedMigrationVersionsConcurrent$' -count=1` passed.
- `make build vet test typecheck` passed.
- `cd web && pnpm build` passed.
- Opus architecture review and Sonnet security review completed; the one confirmed implementation finding (filename semantics stricter than Goose) was fixed, and both confirmation passes succeeded.
- A fresh PostgreSQL integration rerun was attempted, but Docker Desktop is not currently running and `SYNAPSE_TEST_DB_DSN` is unset. The earlier PostgreSQL 17 run on this PR remains successful for `TestMigrateLockedWaitsForMigrationLock` and `TestPostgresReadinessChecks`; this remediation adds forward-schema assertions to the latter, while the non-integration comparison tests cover the same policy locally.

@nghiadaulau please re-review when convenient.

nghiadaulau (approved):
✅ **Approve + merge.** Both findings from the previous review are resolved; re-verified locally (CI intentionally off).

**Blocker (goose `baseFS` data race in the concurrent `/readyz` path) — FIXED.** `embeddedMigrationVersions` no longer calls `goose.SetBaseFS`/`CollectMigrations`; it derives the version set from `migrations.FS` filenames via `migrationVersions` (`fs.ReadDir` + `goose.NumericComponent`, a pure parse) with duplicate detection. The only remaining `goose.SetBaseFS` is in the serial startup `migrate()` path. My 8-goroutine `-race` repro that previously reported `DATA RACE` on `SetBaseFS`/`CollectMigrations` now passes clean.

**Minor (forward-schema readiness during a rolling deploy) — FIXED, well-reasoned.** `compareMigrationStates` now distinguishes a DB version that is `<= latestExpected` (a genuine hole/divergence → still rejected as "divergent") from one `> latestExpected` (a newer migration applied by a rolling-deploy peer that is ahead → tolerated when applied, rejected when present-but-not-applied). An older replica no longer flips to not-ready for the whole deploy window, while true divergence is still caught. Missing/unapplied/duplicate checks remain strict.

**Verified locally:** `go build ./...` + `go vet` + `gofmt -l` clean; the `-race` readiness repro passes; unit tests (`compareMigrationStates` incl. the new rollout cases, `migrationVersions`, lock-key) green; and on a fresh Postgres 17 DB both integration tests pass — `TestMigrateLockedWaitsForMigrationLock` (advisory-lock contention; the lock path is unchanged by the rollout-safe commit) and `TestPostgresReadinessChecks` (exact-state readiness through the new logic).

Security posture is unchanged from the prior clean audit — the delta is pure version-comparison logic + docs, no new secret/exec/DSN/authz surface. Merging.

Issue #633, opened 2026-08-18
Cross-cutting · Deployment & fleet automation (VM + DaemonSet, desired-vs-observed)


nghiadaulau:
> **Cross-cutting workstream (#633)** · child of #594 · **P1**. Related: X2 (#631, secure update), X3 (#632, DaemonSet), A8 (#629, inventory).

## Goal
**Zero-touch deployment & fleet automation**: one agent per VM (installed once, updated via the signed fleet-update channel) and one DaemonSet per cluster — **no per-app, per-workload, or per-machine manual instrumentation** (the anti-OTel-SDK promise). Central **desired-state** reconciled against **observed-state** so a missing/broken agent is a coverage gap, not silence.

## Scope
- One-agent-per-VM install/update flow over the signed fleet-update channel (coordinate with X2 secure update); one DaemonSet per cluster (#632).
- **New workloads on a covered host are observed automatically** — no manual step.
- **Desired-state controller**: which agents/classes should run per host/cluster; reconcile against observed-state (heartbeat/coverage). An agent that should be running detection but isn't → a **coverage gap** (first-class), not silence.
- Fleet views: enrolled vs healthy vs stale vs decommissioned; per-class coverage per host.

## Target (files/interfaces)
- `internal/usecase/fleet/*` (fleet agent registry, coverage) + `internal/usecase/fleet/coverage`; reuse `fleetagent` state machine (active/stale/revoked/decommissioned) + fleet-update channel.
- Desired-vs-observed reconciler; emits `CoverageWindow`/gap on divergence.
- Install/update artifacts (RPM/DEB/DaemonSet) via the existing packaging + fleet-update.

## Tests
- Install once on a VM → new workloads observed automatically (no per-workload step).
- Desired-state says class X should run; agent not reporting X → coverage gap raised (not silence).
- Update via signed channel succeeds; a stale/failed agent shows as such in the fleet view.

## Acceptance / exit
- [ ] One-agent-per-VM + one-DaemonSet-per-cluster; new workloads auto-covered.
- [ ] Desired-vs-observed reconciliation; missing/broken agent → queryable coverage gap.
- [ ] Signed fleet update path exercised; fleet health/coverage views.

## Non-goals
The secure-update crypto itself (#631); the K8s 3-way split (#632); inventory sweep content (#629).

Part of #594.

VietSory:
I'm working on this

nghiadaulau:
[...]

Issue #632, opened 2026-08-18
Cross-cutting · Kubernetes runtime topology (3-way split + K8s context/audit + OpenShift SCC)


nghiadaulau:
> **Cross-cutting workstream (#632)** · child of #594 · **P0** · **depends on** A1 (#622 — the `ResourceContext`/envelope this phase populates the full K8s field-set on). The **Response Actuator** sub-component **depends on** C's response saga + execution journal (#638) — it consumes, never redefines, them. Related: X2 (#631, node sensor portability), X4 (#633). (The inventory-controller + node-sensor components need only A1 #622.)

## Goal
Split the single read-only cluster agent into **three least-privilege components** so read-all + privileged host sensor + workload mutation are never fused in one pod, add full **K8s runtime context + audit/RBAC events**, and support the **OpenShift SCC** security model.

## Scope
- **Three components**:
  1. **Cluster Inventory Controller** — Deployment, read-only K8s API, low privilege.
  2. **Node Runtime Sensor** — DaemonSet, eBPF/cgroup/container context, minimal privilege.
  3. **Response Actuator** — separate identity, narrow RBAC; **consumes** the C-owned response saga (approval + work-order + execution journal, #638) — it executes governed work-orders and appends to that journal, it does **not** define its own approval/journal model.
- **K8s telemetry context** on every envelope: `ClusterID/NodeUID/ContainerID/CgroupID/PodUID/WorkloadUID/Namespace/ServiceAccount/ImageDigest/Runtime` (pod name is **not** identity).
- **K8s audit/RBAC events**: exec/attach, SA-token use, rolebinding changes, privileged/hostPath/hostPID pods, image changes, secret access.
- **OpenShift**: Node Runtime Sensor needs a granted **SCC** — minimal caps (`CAP_BPF`/`CAP_PERFMON`), `hostPID`, hostPath to BTF (`/sys/kernel/btf`) + tracefs, bound to its own service account; the inventory controller needs only a minimal SCC. The 3-way split exists so a cluster admin approves a **narrow** grant, not one all-powerful pod.
- New pods/nodes auto-covered by the DaemonSet.

## Target (files/interfaces)
- `internal/usecase/fleet/clusterinventory` (exists) → the read-only controller; a new node-sensor deployment surface; response-actuator identity/RBAC.
[...]

Issue #631, opened 2026-08-18
Cross-cutting · Agent integrity & compatibility matrix (arm64/CO-RE-BTF/tamper/secure-update; fixes D5)


nghiadaulau:
> **Cross-cutting workstream (#631)** · child of #594 · **P0** · fixes **D5** (arm64). Related: X3 (#632, node sensor), X4 (#633, fleet update).

## Goal
Make the agent **tamper-resistant, securely updatable, portable, and proven across the real OS/arch/runtime matrix** — including **arm64**, which is a hard gap today (BPF objects are compiled `-D__TARGET_ARCH_x86`, **D5**). Portability is by **CO-RE/BTF capability test per host**, never by kernel-version gating (Red Hat backports eBPF into 4.18-based RHEL 8 — version gating is wrong).

## Scope
- **arm64 (fix D5)**: build the eBPF objects for arm64 as well as x86_64; CO-RE relocation at load; prove attach on arm64.
- **Tamper resistance**: detect service-stop / eBPF-detach / config-tamper; watchdog; secure decommission; **disabling detection emits an auditable coverage gap**; central desired-state vs observed-state.
- **Privilege separation**: small privileged sensor ↔ authenticated IPC ↔ unprivileged normalizer/shipper.
- **Secure update channel (X2 owns the crypto/transport)**: verify signatures + enforce pinning, canary, rollback, and **anti-downgrade on the wire** for binaries, **rulepacks** (signed as content by X1 #630 — X2 verifies + delivers, does not author them), and config alike; run the compat gate at install. X4 (#633) drives the fleet rollout orchestration on top of this channel.
- **Key protection**: 0600 baseline; optional TPM/keyring/PKCS#11; tamper-evident spool (coordinate with A2/A0.2).
- **Compatibility matrix with REAL tests**: Ubuntu · RHEL 8/9 · Rocky · Amazon Linux · **RHEL CoreOS (OpenShift node OS)** · Debian · COS × **x86_64 and arm64** × kernel/BTF variants × **containerd and CRI-O**.

## Target (files/interfaces)
- `internal/infrastructure/ebpf/` build: add `-D__TARGET_ARCH_arm64` objects + per-arch build tags; CO-RE/BTF load-time capability probe (per-host, not version check).
- Agent watchdog + tamper-detection in `cmd/synapse-agent`; secure decommission path; coverage-gap emission on self-disable.
[...]

PR #644, opened 2026-08-19
feat(fleet): add asset-scoped desired reconciliation foundation

VietSory:
## Summary

Adds the AssetID-scoped desired-state reconciliation foundation for #633.

Desired fleet capabilities are persisted independently from agent-observed
capabilities and reconciled through the server-authoritative agent -> asset
binding contract.

Policies follow the canonical host/cluster `AssetID` rather than an
enrolment-scoped `AgentID`, so agent reinstall or re-enrolment does not orphan
the desired intent for the underlying asset.

## What changed

- Add a canonical desired fleet policy domain and memory/PostgreSQL stores.
- Add migration `0107_fleet_desired_state.sql` with tenant RLS and DB-level
  canonical capability constraints.
- Add lifecycle-aware `PolicyID + Version` CAS to prevent lost updates and
  clear/recreate ABA races.
- Make clearing desired intent an explicit operation rather than persisting an
  empty policy.
- Add desired-vs-observed reconciliation under
  `internal/usecase/fleet/desired`.
- Emit explicit gaps for:
  - missing agents;
  - stale agents;
  - revoked agents;
  - decommissioned agents;
  - missing capabilities.
- Support multiple current agents serving the same canonical asset. A desired
  capability is covered when any healthy bound agent advertises it.
- Keep the reverse binding assumption intentionally loose: one asset may have
  multiple agents, while one `AgentID` cannot resolve to multiple canonical
  assets.
- Add narrow canonical AssetID readers without expanding the broad asset
  repository port.
- Add migration, RLS, lifecycle, concurrency, ABA, malformed-snapshot,
  replacement-agent, multi-agent and reconciliation regression coverage.

## Desired-state semantics

`fleet_agents.capabilities` remains observed state only.

The desired policy is stored separately and is keyed by canonical `AssetID`.
This prevents a missing, compromised, reinstalled, or re-enrolled agent from
silently erasing the operator's expected coverage.

Desired capability sets are canonicalized and bounded:

- whitespace trimmed;
- duplicates removed;
- deterministic ordering;
- valid UTF-8 only;
- control characters rejected;
- bounded raw input and stored cardinality;
- bounded capability length;
- empty persisted policies rejected.

An identical declarative re-apply is a true no-op: it performs no asset lookup,
ID generation, clock read, persistence write, or audit churn.

## Reconciliation

The reconciler consumes:

1. desired state by canonical asset;
2. the server-authoritative current agent -> asset binding projection;
3. observed fleet-agent state.

For each desired capability, coverage is satisfied when at least one healthy
bound agent currently advertises it.

This allows a replacement/re-enrolled agent to satisfy an existing asset policy
as soon as the authoritative binding points the new agent at the same
`AssetID`, without rewriting desired state.

Malformed, cross-tenant, or internally contradictory snapshots fail closed.

## Concurrency and persistence safety

Desired mutations use `PolicyID + Version` lifecycle-aware CAS.

Coverage includes real competing operations for both memory and PostgreSQL:

- concurrent create -> exactly one version-1 policy wins;
- concurrent updates from the same version -> exactly one wins;
- update racing clear -> exactly one operation wins;
- clear -> recreate mints a new `PolicyID`, preventing stale-delete ABA;
- duplicate `PolicyID` within a tenant conflicts consistently.

PostgreSQL persistence also includes:

- tenant RLS with `FORCE ROW LEVEL SECURITY`;
- canonical asset foreign-key integrity;
- tenant-scoped unique policy IDs;
- DB-level capability canonicalization checks;
- timestamp/version invariants.

The RLS integration test switches to a real
`NOSUPERUSER NOBYPASSRLS` role when the test DSN uses an administrative
PostgreSQL account, so it tests the policy itself rather than relying on
repository query predicates.

## Validation

The final branch is a single commit on the current upstream base.

The final squashed commit contains the exact same Git tree as the fully
hardened pre-squash head, so squashing changed history only.

Feature-specific validation includes:

- Linux Go build: pass
- Linux Go vet: pass
- Windows Go build/vet/test: pass
- frontend test/typecheck/build: pass
- Agent Package Matrix: pass
- fleet desired domain tests: pass
- memory persistence tests: pass
- `internal/usecase/fleet/desired` tests: pass
- PostgreSQL 17 migration `0107`: pass
- PostgreSQL desired repository tests: pass
- RLS tests with a real non-bypass role: pass
- CAS/ABA/concurrency tests: pass
- 10k-asset reconciliation/gap benchmarks exercised during hardening

The repository-wide Linux test job still encounters an existing unrelated
JobQueue fixture that inserts a tenant without the now-required `name`.

`golangci-lint` likewise reports an existing unrelated SA4023 in
`internal/usecase/sca/service.go`.

Neither file is modified by this PR, and no #633-specific failure appears in
those jobs.

## Deliberate boundary

This is a foundation slice of #633, not full completion of the cross-cutting
workstream.

It intentionally does not implement `AssetBinding` or derive canonical identity
from hostname/work-order heuristics. The reconciler consumes the
server-authoritative binding contract instead of creating a competing identity
model.

Final operator HTTP/query wiring should be composed once the authoritative
binding adapter exists.

This PR also intentionally leaves:

- VM RPM/DEB install/update lifecycle to the corresponding deployment work;
- Kubernetes DaemonSet/topology semantics to #632;
- signed fleet-update mechanics to #631;
- final shared `CoverageWindow` materialization/integration to the appropriate
  coverage path.

Refs #633

### DIFF
--- a/internal/domain/fleetdesired/desired.go
+++ b/internal/domain/fleetdesired/desired.go
@@ -1,0 +1,139 @@
+// Package fleetdesired defines control-plane-owned fleet intent for canonical technical assets.
+//
+// Desired state is deliberately separate from fleetagent.Agent.Capabilities. Capabilities are an
+// OBSERVATION reported by an untrusted agent; desired capabilities are an operator decision. Folding
+// the two together would let an agent erase a missing sensor by ceasing to advertise it. The durable
+// subject is the canonical host/cluster AssetID, not an enrolment-scoped AgentID, so replacing or
+// re-enrolling an agent does not orphan the policy it is expected to satisfy.
+package fleetdesired
+
+import (
+	"fmt"
+	"sort"
+	"strings"
+	"unicode"
+	"unicode/utf8"
+
+	"github.com/KKloudTarus/synapse-ce/internal/domain/asset"
+	"github.com/KKloudTarus/synapse-ce/internal/domain/shared"
+)
+
+const (
+	MaxCapabilities     = 64
+	MaxCapabilityInputs = 256
+	MaxCapabilityLen    = 128
+)
+
+// State is the operator-owned desired capability set for one canonical host/cluster AssetID.
+// PolicyID identifies one policy lifecycle and never changes during updates; clearing and recreating
+// the policy mints a new PolicyID. Version starts at 1 within that lifecycle and increments once per
+// semantic replacement. The pair prevents both lost updates and delete/recreate ABA races.
+type State struct {
+	TenantID     shared.ID
+	AssetID      shared.ID
+	PolicyID     shared.ID
+	Capabilities []string
+	UpdatedBy    shared.ID
+	Version      int64
+	Audit        shared.Audit
+}
+
+type GapReason string
+
+const (
+	GapAgentMissing        GapReason = "agent_missing"
+	GapAgentStale          GapReason = "agent_stale"
+	GapAgentRevoked        GapReason = "agent_revoked"
+	GapAgentDecommissioned GapReason = "agent_decommissioned"
+	GapCapabilityMissing   GapReason = "capability_missing"
+)
+
+func (r GapReason) Valid() bool {
+	switch r {
+	case GapAgentMissing, GapAgentStale, GapAgentRevoked, GapAgentDecommissioned, GapCapabilityMissing:
+		return true
+	default:
+		return false
+	}
+}
+
+func SupportedAssetKind(kind asset.Kind) bool {
+	return kind == asset.KindHost || kind == asset.KindCluster
+}
+
+// NormalizeCapabilities returns the canonical persistence/comparison representation: trimmed,
+// de-duplicated, sorted, and bounded. Empty raw slots are ignored; an empty canonical result is handled
+// by the mutation API (which requires an explicit Clear rather than persisting a meaningless empty row).
+func NormalizeCapabilities(in []string) ([]string, error) {
+	if len(in) > MaxCapabilityInputs {
+		return nil, fmt.Errorf("%w: desired state received %d capability inputs, over the %d input bound",
+			shared.ErrValidation, len(in), MaxCapabilityInputs)
+	}
+	seen := make(map[string]struct{}, min(len(in), MaxCapabilities))
+	out := make([]string, 0, min(len(in), MaxCapabilities))
+	for _, raw := range in {
+		capability := strings.TrimSpace(raw)
+		if capability == "" {
+			continue
+		}
+		if !utf8.ValidString(capability) {
+			return nil, fmt.Errorf("%w: capability %q is not valid UTF-8", shared.ErrValidation, capability)
+		}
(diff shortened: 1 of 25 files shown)


nghiadaulau (approved):
✅ **Approve + merge.** Strict dual review (go-arch + security subagents) + local verification on a fresh Postgres 17 (CI intentionally off). Unusually careful implementation.

**go-arch: PASS.** Dependency rule clean — `domain/fleetdesired` imports only domain+stdlib; the port is domain-only; `usecase/fleet/desired` imports only domain + `usecase/ports` with locally-defined narrow readers (the broad asset repo never leaks inward, mirroring `riskstoryuc`); both stores carry compile-time port assertions; no cycle. **CAS/ABA/concurrency sound** — create is `INSERT … ON CONFLICT (tenant,asset) DO NOTHING RETURNING`; update is `UPDATE … WHERE policy_id=$ AND version=N-1` with a `RowsAffected!=1 → ErrConflict` guard; delete is version-matched and distinguishes ABA-conflict from idempotent already-cleared; memory mirrors the same semantics; the service Get→Put/Delete TOCTOU is resolved by the authoritative store CAS. **Idempotent re-apply is a true no-op** — verified it short-circuits before asset lookup / id-gen / clock / put / audit.

**security: essentially clean, no critical/high.** RLS via `synapse_enable_tenant_rls` (ENABLE + FORCE) with the standard tenant policy; composite `FK (tenant_id, asset_id)` makes a cross-tenant asset reference structurally impossible; every query runs under `WithTenant`; fully parameterized SQL; input bounds enforced twice (domain `NormalizeCapabilities` + a SQL CHECK that byte-matches Go's ordering); reconciliation is fail-closed (only a *healthy* advertising agent covers; revoked/stale/decommissioned/missing → gaps, never covered); mutations audited with an attributable actor, tenant rebound from the durable policy.

**Applied from review (fixups pushed as `c00e2c5`):**
- **Fail-closed hardening**: `NewService` now rejects a non-positive `staleAfter` (which `AgentStateFrom` treats as "never stale" → a coverage fail-open) + a regression test.
- Doc comments on the three exported domain symbols that lacked them (`GapReason`, `GapReason.Valid`, `SupportedAssetKind`).
- `gofmt` (struct-tag alignment + a one-line func body).

**Local verify:** `go build ./...` + `go vet` + `gofmt -l` clean; `-race` green across domain/usecase/memory; and on a fresh Postgres 17 DB the full repo suite passes — concurrent-CAS (exactly one winner), duplicate-policy-id conflict, delete-recreate stale-lifecycle (ABA), missing-subject→conflict, migration 0107 (RLS FORCE + capability/time CHECKs), and the RLS test under a real `NOSUPERUSER/NOBYPASSRLS` role.

**Scope (as stated):** foundation only — domain + stores + usecase + port + migration. No composition-root wiring or HTTP handler yet, so request-edge RBAC/authz must be verified when the service is wired behind the `authz` chokepoint (follow-up on #633). Merging the foundation.

Issue #609, opened 2026-08-18
A0.4 · Delivery/replay/gap contract (at-least-once + idempotent + durable WAL)


nghiadaulau:
> Phase **A0** · child of #594 · **P0**.

## Goal
Define **and** implement the telemetry/detection **delivery contract** so no security signal is silently lost and replays are safe.

## Contract (no exactly-once claim)
**at-least-once delivery + idempotent ingest + durable agent spool + monotonic sequence per AgentSession/Stream + highest-contiguous ACK + explicit gaps.**

## Scope
- **Idempotent ingest** keyed on `(AgentSession/Stream, Sequence, event index)` — a resend is a no-op, not a duplicate.
- **Durable agent spool (WAL)** with priority classes:
  - P0 response-verification / coverage / sensor-health · P1 detections · P2 privilege / critical-file · P3 background process/network.
  - disk quota, eviction policy, fsync policy, corruption recovery, retry backoff+jitter, resume-after-restart, disk-full behavior, 429/5xx behavior, agent-side metrics.
- **Session / Stream / Boot ids** so a reboot / reinstall / sequence-reset-to-1 is distinguishable from a replay (host/class sequence alone is not enough).
- **Highest-contiguous ACK** + **explicit persisted gaps** for anything not contiguous.

## Acceptance / exit
- [ ] Agent offline → events durably spooled → resumed after restart → delivered → ACK → local batch deleted; nothing on the `event → HTTP failure → lost` path.
- [ ] A reboot with a reset sequence is **not** treated as a replay.
- [ ] A missing batch surfaces as an **explicit, queryable gap** (no silent loss).
- [ ] Idempotent re-delivery produces no duplicate stored events.

## Dependencies
Needs **A0.1** (session/agent identity) + **A0.3** (schema). Feeds A2 (WAL implementation) / A3 (transport).

nghiadaulau:
Landed the delivery **contract** in #617 (squash-merged to `main` as `76626a9e`) — the pure-domain primitives both sides agree on (`internal/domain/fleetagent/delivery.go`).

**Shipped (contract + exhaustive tests):**
- **Priority ladder** `DeliveryPriority` P0..P3 (drain/eviction order): P0 response-verification/coverage/sensor-health · P1 detections · P2 privilege/critical-file · P3 background process/network.
[...]

PR #646, opened 2026-08-20
feat(agent): add durable priority telemetry spool

### LINKED ISSUE #623: A2 · Priority durable agent WAL/spool (carries the A0.4 WAL tail)
> **Phase A2** · child of #594 · **P0** · depends on: A1 (#622), A0.4 delivery contract (#609). Blocks: A3 (#624).
> **Carries the deferred A0.4 tail:** the durable agent write-ahead spool that #617 (delivery contract) explicitly left to A2.

## Goal
Implement the **priority durable agent spool (WAL)** so no security signal is lost across agent offline / restart / disk pressure. #617 landed the pure-domain delivery contract (`StreamPosition`/incarnation, `AckLedger` highest-contiguous ACK + explicit gaps, `DeliveryKey`); the never-shed classification (`MustNotShed`, in `internal/domain/telemetry/loss.go`) came with the A0.6 loss half (#618). A2 builds the on-disk implementation that produces and drains them.

## Scope
- Per-priority durable queues on disk keyed by the A0.4 `DeliveryPriority` ladder (P0 response-verify/coverage/sensor-health · P1 detections · P2 privilege/critical-file · P3 background process/network).
- **fsync policy** (durability vs throughput per class), **disk quota** + **eviction policy** (shed lowest priority first; a `MustNotShed` class over quota → back-pressure, never silent drop — record a `TelemetryGap`).
- **Corruption recovery** (torn-write detection, checksummed records, skip-and-report bad frames).
- **Resume-after-restart**: reload the spool, continue the monotonic `Sequence` within the incarnation, or start a new incarnation (Epoch++) on spool loss/reboot — per the A0.4 contract.
- **Retry backoff + jitter**; disk-full behavior; 429/5xx handling from the (future A3) transport.
- Agent-side metrics: spool depth per priority, evicted count, fsync latency, corruption events, oldest-unacked age.

## Target (files/interfaces)
- `internal/infrastructure/spool/` (new, agent-side) — `Spool` implementing a `ports.TelemetrySpool` port; per-priority segment files, index, checksums.
- Consumes `internal/domain/fleetagent/delivery.go` (`StreamPosition`, `AckLedger`, `DeliveryPriority`, `DeliveryKey`) + `internal/domain/telemetry/loss.go` (`MustNotShed`).
- Wired in `cmd/synapse-agent` (the agent binary) — enqueue on normalize (#622), drain on transport (#624).
- Emits first-class `TelemetryGap` (domain) on eviction/quota back-pressure; records `SensorState` for the P0 sensor-health stream.
- **Emits/persists** `TelemetryGap`, `CoverageWindow`, `SensorState` — the domain types are **defined/owned by A0.6 (#611)** as first-class persisted objects (not best-effort audit lines); A2 is a producer of them, it does not re-reserve them.

## Tests
- **Resume-after-restart**: enqueue → kill → reopen → the exact set (order + membership) survives; sequence continues; a genuine reset advances Epoch (not classed as replay).
- **Disk-full / quota**: eviction sheds P3 first; a P2/`MustNotShed` overflow back-pressures + records a gap; never silent loss.
- **Corruption recovery**: inject a torn/hand-corrupted frame → it is skipped, reported, and the rest drains.
- **fsync durability**: a crash after fsync keeps the record; before fsync is bounded loss represented as a gap.
- Backoff+jitter unit tests; oldest-unacked-age metric; property test: everything enqueued is either drained-and-ACKed or represented as a durable gap (coverage-honesty invariant).

## Acceptance / exit
- [ ] Agent offline → events durably spooled → resumed after restart → nothing on the `event → failure → lost` path.
- [ ] Never-sample classes are never silently dropped under disk pressure (back-pressure + queryable gap instead).
- [ ] Corruption is recovered from and reported, not fatal.
- [ ] Priority draining + eviction proven; agent-side spool metrics exported.
- [ ] `-race` clean; property test for no-silent-loss passes.

## Non-goals
The wire batching/transport + ACK exchange (#624); the columnar server ingest (already exists, #424).

Part of #594.

XUanhoa04:
## Summary

Closes #623.

Implements the A2 priority durable agent WAL and wires it into the live host-agent detection path. Raw eBPF observations are normalized to the canonical A1 telemetry envelope and committed before rule evaluation; confirmed detections, coverage snapshots, and per-class sensor state enter the same WAL at their contracted priorities.

This preserves the delivery claim from #609: at-least-once delivery, incarnation-aware monotonic positions, highest-contiguous exact-epoch ACKs, and explicit durable gaps. It does not claim exactly-once delivery.

## Changes

- Add ports.TelemetrySpool with immutable enqueue/read records, exact priority+epoch ACKs, stats, and queryable loss evidence.
- Add checksummed per-priority segment files with a fixed bounded frame format, payload CRC32C, exclusive process locking, owner-only Unix modes, bounded recovery allocation, and atomic dual-generation state.
- Resume sequence after process restart; advance Epoch on boot/enrolment change or state loss; never reuse an incarnation already present in WAL or gap history.
- Detect torn/hand-corrupted frames, resynchronize to later valid frames, commit exact or explicitly unknown gaps, and atomically rewrite repaired segments so corruption is not rediscovered every restart.
- Enforce a logical disk quota by deleting P3 segments only after exact eviction gaps are fsynced. P0-P2 return the shared saturation/backpressure error and record a durable quota-backpressure gap; repeated retries of the same event do not flood the gap journal.
- Roll partial WAL and gap-journal writes back to their last known-good offsets. ENOSPC becomes observable backpressure rather than a malformed tail.
- Persist ACK state before deleting bytes; reject ahead/stale-incarnation ACKs and keep equal/regressive ACKs idempotent.
- Add configurable P3 batch sync, mandatory always-sync for P0-P2, flush/close semantics, and fsync metrics.
- Add capped exponential full-jitter retry policy for network errors, 408, 429 with bounded Retry-After, and 5xx; other 4xx are permanent.
- Add a private Prometheus collector/listener with bounded priority-only labels for depth, bytes, oldest-unacked age, sequence/ACK, gaps, eviction, corruption, and fsync latency.
- Replace the host agent detection file sink with the durable P1 sink and add a sensor adapter which commits normalized raw telemetry before forwarding it to rule evaluation. Legacy userspace-only event timestamps are flagged as kernel-timestamp-unavailable rather than presented as kernel truth.
- Persist aggregate coverage and per-class sensor-state snapshots as never-shed P0 records.
- Document the on-disk format, crash ordering, recovery invariants, quota behavior, A3 handoff, settings, and metrics.

## Safety invariants covered by tests

- enqueue -> close/restart -> exact membership and lane ordering survive; sequence continues
- boot/session change -> Epoch increments and Sequence restarts at 1
- accepted record -> readable, ACKed, or covered by durable gap evidence
- P3 is evicted first; P0-P2 are never deleted under pressure
- gap fsync precedes P3 deletion; ACK persistence precedes segment deletion
- a corrupt middle frame is skipped while later frames remain readable
- a torn or unsynced tail becomes exact loss evidence without coordinate reuse
- both corrupt state generations recover into a new non-reused incarnation
- concurrent producers receive unique monotonically increasing coordinates
- one process owns a spool directory at a time on Unix and Windows
- retry classification, jitter bounds, metric values, and deterministic record IDs

## Non-goals

A3 still owns wire batching, signing, HTTP delivery, server ACK exchange, and mapping local gaps into the control-plane persisted gap model. This PR exposes Peek, Ack, Gaps, and RetryPolicy for that work; it does not add a competing transport.

## Validation

- [x] go build ./cmd/...
- [x] go vet ./...
- [x] go test ./... -count=1
- [x] pnpm run typecheck in web
- [x] Linux amd64 cross-compilation for spool, agentspool adapter, and synapse-agent
- [x] focused frame fuzz run: 209,662 executions, no crash
- [x] tests added for restart, corruption, torn writes, quota/backpressure, property invariant, concurrency, metrics, retry, agent wiring, and platform locks
- [x] documentation/configuration drift guards pass
- [x] no dashboard changes; web build/screenshots not applicable
- [x] no new HTTP routes or OpenAPI changes

The local Windows Go toolchain has no C compiler, so the race-enabled run is left to the repository Linux CI job; the concurrency suite itself passes locally and the Linux targets compile.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -304,6 +304,8 @@ SYNAPSE_GOMODGRAPH_ENABLED=true      # transitive Go dep edges via `go mod graph
 # SYNAPSE_FLEET_URL=https://synapse.example.com
 # SYNAPSE_FLEET_ENROL_TOKEN_FILE=/run/secrets/synapse-enrol-token
 # SYNAPSE_AGENT_STATE_DIR=/var/lib/synapse-agent
+# SYNAPSE_TELEMETRY_SPOOL_BYTES=536870912             # P0-P3 durable WAL quota (512 MiB)
+# SYNAPSE_AGENT_METRICS_ADDR=127.0.0.1:9465           # optional, unauthenticated private scrape listener
 
 # =============================================================================
 # Leader election (horizontal scale) – fence scheduled work to one node
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,13 @@ and this project aims to adhere to [Semantic Versioning](https://semver.org/spec
 
 ### Added
 
+- **Crash-recoverable priority telemetry spool for the host agent.** Normalized eBPF events and
+  confirmed detections now enter checksummed, per-priority WAL segments before downstream use. The
+  spool resumes sequence incarnations after restart, commits exact-epoch ACKs, evicts only P3 under
+  quota pressure, persists queryable gap evidence before any loss, repairs torn/corrupt frames, and
+  exports bounded-cardinality depth, oldest-age, eviction, corruption, and fsync metrics. P0–P2 apply
+  backpressure instead of silently shedding; A3 will consume the exposed peek/ACK and retry contract.
+
 - **Authority-surface precondition on AI-triage promotion.** Evaluation reports record
   `gate_reachable_pairs`, and the promotion boundary requires at least one counterfactual pair the
   deterministic policy could actually exempt (`--minimum-gate-reachable-counterfactual-pairs`,
--- a/cmd/synapse-agent/detect.go
+++ b/cmd/synapse-agent/detect.go
@@ -5,15 +5,14 @@ import (
 	"errors"
 	"fmt"
 	"log"
-	"path/filepath"
 	"sort"
 	"strconv"
 	"strings"
 	"time"
 
+	"github.com/KKloudTarus/synapse-ce/internal/adapter/agentspool"
 	"github.com/KKloudTarus/synapse-ce/internal/domain/detection"
 	"github.com/KKloudTarus/synapse-ce/internal/domain/shared"
-	"github.com/KKloudTarus/synapse-ce/internal/infrastructure/detectsink"
 	"github.com/KKloudTarus/synapse-ce/internal/infrastructure/ebpf"
 	"github.com/KKloudTarus/synapse-ce/internal/infrastructure/fleetclient"
 	detectuc "github.com/KKloudTarus/synapse-ce/internal/usecase/fleet/detect"
@@ -56,31 +55,49 @@ func (r *runner) startDetection(ctx context.Context, cred fleetclient.Credential
 		log.Print("detection: enrolled credential has no canonical agent id; detection engine disabled")
 		return
 	}
-	sinkPath := filepath.Join(r.cfg.stateDir, "detections.jsonl")
-	sink, err := detectsink.New(sinkPath)
+	durable, identity, err := r.openTelemetrySpool(ctx, cred)
 	if err != nil {
-		log.Printf("detection: %v; detection engine disabled", err)
+		log.Printf("detection: open durable telemetry spool: %v; detection engine disabled", err)
 		return
 	}
-	sensor := ebpf.NewSensor(host, agent, classes)
+	rawSensor := ebpf.NewSensor(host, agent, classes)
+	sensor, err := agentspool.NewDurableSensor(rawSensor, durable, identity)
+	if err != nil {
+		log.Printf("detection: wire durable telemetry sensor: %v; detection engine disabled", err)
+		_ = durable.Close()
+		return
+	}
+	sink, err := agentspool.NewDetectionSink(durable)
+	if err != nil {
+		log.Printf("detection: wire durable detection sink: %v; detection engine disabled", err)
+		_ = durable.Close()
+		return
+	}
 	eng, err := detectuc.NewEngine(sensor, sink, host, agent, detectuc.Options{
 		Classes:       classes,
 		CPUCeilingPct: r.cfg.detectCeiling,
 	})
 	if err != nil {
 		log.Printf("detection: %v; detection engine disabled", err)
-		_ = sink.Close()
+		_ = durable.Close()
 		return
 	}
+	if err := r.startSpoolMetrics(ctx, durable); err != nil {
+		log.Printf("detection: agent metrics listener unavailable: %v", err)
+	}
 
-	log.Printf("detection engine starting: classes=%s ceiling=%.0f%% sink=%s", r.cfg.detectClasses, r.cfg.detectCeiling, sinkPath)
+	log.Printf("detection engine starting: classes=%s ceiling=%.0f%% durable_spool=%s", r.cfg.detectClasses, r.cfg.detectCeiling, r.telemetrySpoolDir())
 	// One-shot coverage report shortly after start, so the operator can see which classes actually came
 	// up on this host and which are gaps — never silently assume a class is observing.
 	go func() {
 		select {
 		case <-ctx.Done():
 		case <-time.After(3 * time.Second):
-			log.Printf("detection coverage: %s", formatCoverage(eng.Coverage()))
+			coverage := eng.Coverage()
+			log.Printf("detection coverage: %s", formatCoverage(coverage))
+			if err := agentspool.RecordCoverage(ctx, durable, coverage, time.Now().UTC()); err != nil {
+				log.Printf("detection: persist coverage/sensor state: %v", err)
+			}
 		}
 	}()
 	go func() {
@@ -88,7 +105,9 @@ func (r *runner) startDetection(ctx context.Context, cred fleetclient.Credential
 		if err != nil && !errors.Is(err, context.Canceled) {
 			log.Printf("detection engine stopped: %v", err)
 		}
-		_ = sink.Close()
+		if closeErr := durable.Close(); closeErr != nil {
+			log.Printf("detection: close durable spool: %v", closeErr)
+		}
 	}()
 }
 
--- a/cmd/synapse-agent/detect.go
+++ b/cmd/synapse-agent/detect.go
@@ -82,29 +82,38 @@ func (r *runner) startDetection(ctx context.Context, cred fleetclient.Credential
 		_ = durable.Close()
 		return
 	}
-	if err := r.startSpoolMetrics(ctx, durable); err != nil {
+	runCtx, cancelRun := context.WithCancel(ctx)
+	if err := r.startSpoolMetrics(runCtx, durable); err != nil {
 		log.Printf("detection: agent metrics listener unavailable: %v", err)
 	}
 
 	log.Printf("detection engine starting: classes=%s ceiling=%.0f%% durable_spool=%s", r.cfg.detectClasses, r.cfg.detectCeiling, r.telemetrySpoolDir())
 	// One-shot coverage report shortly after start, so the operator can see which classes actually came
 	// up on this host and which are gaps — never silently assume a class is observing.
+	coverageDone := make(chan struct{})
 	go func() {
+		defer close(coverageDone)
+		timer := time.NewTimer(3 * time.Second)
+		defer timer.Stop()
 		select {
-		case <-ctx.Done():
-		case <-time.After(3 * time.Second):
+		case <-runCtx.Done():
+		case <-timer.C:
 			coverage := eng.Coverage()
 			log.Printf("detection coverage: %s", formatCoverage(coverage))
-			if err := agentspool.RecordCoverage(ctx, durable, coverage, time.Now().UTC()); err != nil {
+			if err := agentspool.RecordCoverage(runCtx, durable, coverage, time.Now().UTC()); err != nil && !errors.Is(err, context.Canceled) {
 				log.Printf("detection: persist coverage/sensor state: %v", err)
 			}
 		}
 	}()
 	go func() {
-		err := eng.Run(ctx)
+		err := eng.Run(runCtx)
 		if err != nil && !errors.Is(err, context.Canceled) {
 			log.Printf("detection engine stopped: %v", err)
 		}
+		cancelRun()
+		// A timer that fired concurrently may still be persisting coverage. Wait
+		// for it before closing the shared spool so no operation races Close.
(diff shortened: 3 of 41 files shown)


nghiadaulau (changes requested):
Strict review — go-arch + security subagents + local verification on macOS (CI intentionally off). This is genuinely high-quality work: frame CRC32C + multi-layer decode, crash recovery with torn-write/corrupt-frame gaps and atomic segment repair, dual-copy generation-counter state, per-priority ACK epochs, `0700`/`0600` at-rest perms with `O_EXCL` creation, clean dependency rule + compile-time port assertions. Build/vet/gofmt clean; `go test -race ./internal/infrastructure/spool/... ./internal/adapter/agentspool/... ./cmd/synapse-agent/...` all green.

But there are **two real defects that break A2's core promise — a *bounded* durable spool with honest backpressure** — so requesting changes.

## 🔴 Blocker 1 (Major) — the gap journal is unbounded and lives outside the `MaxBytes` quota
`quota.go` gates only `s.totalBytes` (WAL segment bytes); `gapBytes` is tracked (`gaps.go`, `spool.go`) but never checked against `cfg.MaxBytes`. Under a sustained control-plane outage — exactly what A2 exists to survive — P0–P2 (must-not-shed) fill the quota, all P3 segments are evicted, and then the P3 sensor path drop-and-advances (`agentspool/sensor.go:168-173`): every *distinct* dropped P3 event (new `EventID`) misses the `lastRejected[P3]` dedup in `saturatedLocked` (`quota.go:57-62`) and appends **one durable unknown-coordinate gap + one fsync** (`gaps.go`). On the highest-volume lane (process/network) that is unbounded disk growth + an fsync storm precisely when the host is already stressed — the opposite of the `Config` "bounded resource" guarantee.
**Fix:** bound/rotate the gap journal against a budget, and **coalesce** the ongoing P3-saturation case into a single incrementing-`Count` gap instead of one unknown-coordinate gap per dropped event.

## 🔴 Blocker 2 (Medium, but a live-lock) — an oversized record is reported as *retryable* saturation
`quota.go:15-16`: when `required > cfg.MaxBytes`, `ensureCapacityLocked` returns `saturatedLocked` → `SaturatedError`, which `Unwrap()`s to `ErrTelemetrySpoolSaturated` (`errors.go:38`) — the sentinel producers treat as transient backpressure. The non-sheddable producers loop on it forever: `agentspool/sensor.go` (non-P3 branch), `detection_sink.go:46-47`, `health.go:74-75` all `for { … if !errors.Is(err, ErrTelemetrySpoolSaturated) break; waitForSpoolCapacity(ctx) }`. For a record larger than the entire quota this **never** fits, so it live-locks until ctx cancel. It is reachable: `cmd/synapse-agent/telemetry_spool.go:55-60` normalizes `MaxRecordBytes` down toward `MaxBytes` for a small `spoolBytes`, and a frame adds `meta+48` overhead (`spool.go:183`), so a payload near `MaxRecordBytes` can exceed `MaxBytes`.
**Fix:** return a **permanent** (non-retryable, `shared.ErrValidation`-class) error when `required > MaxBytes`, distinct from the transient saturation sentinel; and have `normalizeConfig` reserve frame overhead so `MaxRecordBytes + overhead ≤ MaxBytes`.

## 🟡 Should-fix
- **Per-`Enqueue` state fsync defeats P3 `SyncBatch`.** `spool.go:233` calls `persistStateLocked` on every enqueue, and `persistState` does synced-tmp-write + 2 renames + dir fsync (~2 fsync/record) even on the batch lane whose purpose is to avoid per-record fsync. Validate against the `make load` throughput SLO; consider deferring the state fsync on the batch lane (while keeping exact-gap bookkeeping correct).
- **Generic write failures mislabeled as quota gaps** (`spool.go:200-204`): a transient `Write` error (only `ENOSPC` special-cased) is recorded as `SpoolGapQuotaBackpressure/Eviction`. Add a generic-I/O gap reason (or reuse `TornWrite`) so the loss taxonomy stays honest.
- **Spurious unreconcilable gap on a retried write failure** (`spool.go:199-211`): an unknown-coordinate gap is appended and `AssignedThrough` is not advanced, so a backpressure producer re-enqueues and later succeeds — but the earlier unknown gap can never be cleared, over-reporting loss for a signal that was in fact delivered.
- **Duplicate on-disk coordinate hard-fails `Open`** (`recovery.go:70-72`) → disables the whole detection engine. Consider keeping the CRC-valid copy and gapping the other, rather than stranding an otherwise-usable spool.

## 🟡 Security (LOW, defense-in-depth)
`securePath` (`permissions.go:13`, `os.Chmod`) and recovery `os.ReadFile` (`recovery.go:121`, size-bound checked only *after* the full read) both **follow symlinks** and lack `O_NOFOLLOW`. Only exploitable by an actor who can already write into the agent-owned `0700` spool dir (root/same-user), or via a misconfigured world-writable `SYNAPSE_AGENT_STATE_DIR` — and creation is symlink-safe via `O_EXCL` — so it is LOW. Harden: `Lstat`-guard / `O_NOFOLLOW` on segment/state/gap opens, `Stat().Size()` check before `ReadFile`, and document that the state dir must be non-world-writable.

**Dependency rule: PASS** (port→domain only; spool→domain+ports; agentspool→domain+usecase, no infra; cmd = wiring; assertions present). Fix the two blockers (and ideally the should-fixes) and I'll re-review — the foundation is sound.

XUanhoa04:
Addressed the requested changes in `e546bce` (`fix(agent): bound durable spool loss evidence`). Please re-review.

Blockers fixed:
- Bounded the gap journal within `MaxBytes` using a reserved `MaxGapBytes` budget. The reserve covers both the active journal and atomic-compaction scratch space, WAL segments cannot consume it, and `TotalBytes` now accounts for WAL + gap bytes.
- Coalesced sustained P3 saturation into one unknown-coordinate gap with an incrementing `Count`, with checkpoint persistence batched by count/interval and flushed on `Flush`/`Close`. The regression test covers 1,000 distinct rejected events, restart persistence, bounded journal size, and a handful of checkpoint sync cycles instead of one fsync per event.
- Permanently oversized frames now return `shared.ErrValidation`, not `ErrTelemetrySpoolSaturated`, so must-not-shed producers do not live-lock. Config normalization also reserves frame/header/metadata overhead and validates the resulting limits.

Additional review items addressed:
- P0-P2 transient backpressure and successfully rolled-back writes no longer create unreconcilable loss gaps.
- Added an explicit `io_failure` gap reason for P3 ENOSPC/write-loss cases instead of labeling them quota eviction/backpressure.
- Added `Lstat` symlink rejection for managed paths, pre-`ReadFile` segment size checks, and documentation that the state directory must remain agent-owned and non-world-writable.

Two suggested items remain intentionally fail-closed:
- P3 still commits state per accepted enqueue. Deferring that commit without a durable reservation protocol would make exact unsynced-tail loss evidence impossible after a crash.
- Duplicate on-disk coordinates still fail `Open`. Choosing between two CRC-valid records without authenticated precedence can silently accept ambiguous content; a safe repair policy needs a separate format/protocol design.

Verification completed:
- `go test ./...`
- `go vet ./...`
- focused spool/adapter/agent tests
- `gofmt` and `git diff --check`

I could not run the race build on this Windows host because no C compiler (`gcc`) is installed; the non-race suite is green. Please run/re-run the macOS race verification during re-review.

nghiadaulau (approved):
✅ **Approve + merge.** Re-reviewed the `e546bce3` update against my earlier request-changes — both blockers, the security LOW, and the two write-path should-fixes are resolved, each with a dedicated regression test. Re-verified locally (build/vet/gofmt clean; `go test -race` green across spool/agentspool/ports/cmd-agent/fleetagent, spool pkg 9.3s).

**🔴 Blocker 1 (unbounded gap journal + P3 fsync storm) — FIXED.** A `MaxGapBytes` reserve now lives *inside* `MaxBytes` (WAL quota = `MaxBytes − MaxGapBytes`, default `RecommendedGapBytes`), so WAL records can't consume the loss-evidence reserve and the gap journal is itself bounded (`gapJournalLimitLocked = MaxGapBytes/2`, leaving atomic-rewrite headroom). Sustained P3 saturation now **coalesces** into one incrementing-`Count` gap (`recordP3LossLocked`, batched fsync, overflow-guarded) instead of one durable record + fsync per dropped event; `mergeGap`/`compactGaps` merge same-key gaps. When the reserve still can't hold it, the spool **fails stop with `ErrGapJournalFull`** rather than grow unbounded or erase evidence; compaction is atomic (O_EXCL tmp → replace → dir-sync). `usedBytes = totalBytes + gapBytes` is now the tracked total. Covered by `TestDistinctP3SaturationCoalescesWithinGapBudget`.

**🔴 Blocker 2 (oversized-record live-lock) — FIXED.** `ensureCapacityLocked` returns a **permanent `shared.ErrValidation`** when `required > SegmentBytes || required > walMax` — distinct from the retryable `ErrTelemetrySpoolSaturated`, so a non-sheddable producer no longer retries forever on a record that can never fit. `normalizeConfig` + the agent composition root now reserve `FrameOverheadBudget`, guaranteeing `MaxRecordBytes + overhead ≤ SegmentBytes ≤ WAL`. Covered by `TestFrameLargerThanPermanentCapacityIsNotRetryable` + `TestEncodeFrameRejectsMetadataOutsideReservedOverhead`.

**🟡 Should-fixes #4/#5 — FIXED.** New `SpoolGapIOFailure` reason (generic write errors no longer mislabeled as quota gaps); a transient write failure now rolls back and returns without appending a spurious unreconcilable gap, and fail-stops if rollback itself fails. Covered by `TestNeverShedSaturationBackpressuresWithoutClaimingLoss` + `TestRepeatedBackpressureDoesNotCreateLossGaps`.

**🟡 Security LOW — FIXED.** `securePath` now `Lstat`s and refuses a symlink before chmod (and re-`Lstat`s after); recovery `scanSegmentLocked` `Lstat`s + refuses a symlink + checks `info.Size()` against the bound **before** `os.ReadFile`. `Config.Dir` doc now states it must be service-owned/non-world-writable. Covered by `TestSecurePathRejectsSymlink`.

**Non-blocking (optional future):** the per-`Enqueue` state fsync (should-fix #3) is now paired with a batched *gap* fsync but the state fsync itself still fires per record — worth validating against the `make load` P3 throughput SLO later; and the duplicate-on-disk-coordinate `Open` hard-fail (#6) remains fail-stop, which is a defensible integrity stance. Neither blocks.

Dependency rule PASS (unchanged). Excellent turnaround — merging.

Issue #593, opened 2026-08-17
Validation: benchmark harness, SLO definition, and blind-review AI-triage pilot


nghiadaulau:
> Part of #583 (production-readiness epic). Item **C3** · Priority **P1** (benchmark), **P2** (SLO / AI pilot) · Group C (Security Posture & Validation).

## Problem
Nothing has been measured on real data:
- No benchmark, load test, or latency/throughput measurement exists (`E-REQ-03`). The only numbers in the repo are configuration **limits** (2 GiB / 10 GiB workspace, 24h scan timeout, 512 MiB / 256 PID sandbox, 32-connection pool, DAST concurrency 4) — these are blocking thresholds, not performance.
- No SLO, error budget, dashboard, or DR drill (review §15.6).
- AI is intentionally constrained (agent + FP-triage default off, no LLM in the report path, distinct verifier required for gate exemption). That is **governance evidence, not accuracy evidence** — precision/recall on a user's data domain is unmeasured (review §15.7).

## Evidence
- `E-REQ-03`; review §15.5 (perf `U`), §15.6 (SLA `U`), §15.7 (AI accuracy `U`); `E-CFG-06`, `E-SEC-08`.

## Impact
Capacity planning, any SLA commitment, and any claim about AI-triage quality are unsupported until measured. Enabling AI gate-exemption without a measured precision/recall baseline would be a governance risk.

## Proposed work
1. **Benchmark harness (P1).** Measure: scan time by codebase size and language; concurrent-scan capacity before latency degrades; evidence-store growth per scan; and 103-migration time on data-heavy databases at several sizes.
2. **SLO definition (P2).** Once perf is known, define SLOs and an error budget; wire dashboards off the A1 (#584) metrics.
3. **AI-triage pilot (P2).** Run a pilot with **blind human review** to establish precision/recall **before** enabling gate exemption (`SYNAPSE_FP_TRIAGE_MODE=enforce` with a distinct `SYNAPSE_VERIFIER_MODEL`).

## Acceptance criteria
- [ ] Benchmark numbers exist for the four measures above, recorded and reproducible.
- [ ] SLOs are defined against measured data (not inferred from limits).
- [ ] AI-triage precision/recall are reported from a blind-review pilot before any gate-exemption is enabled in production.

## Dependencies
SLO work depends on **A1** (#584) metrics. Corresponds to review items P2-3 and P3-1.

pho-veteran:
[...]

Issue #589, opened 2026-08-17
Scale: remove API single-instance-per-role cap to enable horizontal scale / HA


nghiadaulau:
> Part of #583 (production-readiness epic). Item **B2** · Priority **P1** · Group B (Scale & Deployment Architecture).

## Problem
The API tier is hard-capped to **one instance per role**. A second API instance of the same role fails to take the per-role session advisory lock and exits immediately. There is no leader election, no active-active configuration, and no health-based failover. The worker tier already has a horizontal path via `FOR UPDATE SKIP LOCKED`; the API tier does not.

## Evidence
- `E-ARCH-03` — per-role advisory lock; review §15.3 (availability = `U`) and §15.4 (scalability = `U`).

## Impact
No high availability for the API; a single instance is a single point of failure. Uptime, RPO, and RTO are undefined. Scaling the API horizontally requires removing this constraint in code — not just splitting migration out (that is B1).

## Proposed work
- After **B1** (#588) lands (migration is no longer a per-instance startup side-effect), remove or relax the API singleton advisory lock so multiple API instances can run concurrently behind a load balancer.
- Audit every startup side-effect to confirm it is safe to run N-way, or move it into the gated migration/init step.
- Characterize behavior under concurrency: max workers a single database sustains, max concurrent scans per API instance, and the 32-connection pool under load (all currently `U`).

## Acceptance criteria
- [ ] ≥2 API instances run concurrently behind a load balancer with no startup data races.
- [ ] A documented HA topology exists (LB + N API instances + gated migration).
- [ ] Concurrency limits (workers/DB, concurrent scans/instance, pool behavior) are measured and documented.

## Dependencies
**Blocked by B1** (#588) — do not remove the singleton cap until migration is a gated pre-step. Relates to the platform leader-election / horizontal-scale roadmap.

pho-veteran:
## HA implementation and remaining acceptance gate from PR #648

PR #648 implements the horizontal API foundation:

- removes the inappropriate API-wide singleton constraint while retaining gated production migration behavior;
- makes bootstrap-admin creation concurrency-safe and idempotent across replicas;
- preserves leader leases and durable queue claim/run-lock semantics for work that must remain singleton or safely distributed;
- adds bounded, low-cardinality PostgreSQL pool metrics;
- ships a production Helm topology with two API replicas behind a TLS ingress, readiness/liveness probes, rolling updates, PDB, and topology spreading.

### Staging result

[...]

Issue #583, opened 2026-08-17
Epic: production-readiness gaps — operations, scale, and validation (from ops evidence review @ 24693b22)


nghiadaulau:
## Context

This epic tracks the **production-readiness gaps** surfaced by an internal deployment & operations evidence review (doc v0.2.0) conducted against `main` at commit `24693b22f72f8ab3886c18e6a408aba42118c709` (`v0.1.8-10-g24693b2`). The review graded every claim by evidence status — `I` implemented, `P` partial, `D` design/doc only, `U` unverified — and cites `file:line`.

**Framing.** The security core is solid and tested: server-side execution gate (scope + authorization window + RoE, with denied-path tests) [E-SEC-03], append-only hash-chained evidence with report blocking on tamper [E-SEC-07], deterministic no-LLM report path (arch test incl. transitive imports) [E-SEC-08], MCP propose-only [E-SEC-09], RBAC + tenant isolation [E-SEC-02], multi-encoding secret redaction [E-SEC-05]. The gaps below are **not** in security features — they are in the **operational, scale, and validation layers** that a production deployment needs. None of these blocks development/demo use; all of them block an un-caveated production go-live.

Each item lists: **Missing** · **Evidence** · **Impact** · **Proposed work** · **Acceptance criteria**. Priority maps to the review's P0/P1/P2. Items are written so they can be split into child issues (`convert to issue` on each checkbox group). Related: #59 (roadmap/state).

---

### A1 · No metrics endpoint / OpenTelemetry not wired  `P0`
- **Missing.** There is no `/metrics` endpoint, no `promhttp`, and OpenTelemetry is not initialized at any composition root — the `go.opentelemetry.io` modules in `go.mod` are all `// indirect`. No env var configures telemetry. The fleet blue-team guide notes the telemetry-store boundary exists but the operator wiring is not connected.
- **Evidence.** `E-OPS-03` — `go.mod` (indirect OTel), `docs/guide/fleet-blue-team.md`; verified: `grep promhttp|"/metrics"` over `internal/`+`cmd/` returns nothing.
- **Impact.** Latency, error rate, queue depth, and resource usage cannot be observed from the app; every signal must be reconstructed from a reverse proxy or Docker/Postgres. Incidents are detected late (`R-07`).
[...]

PR #647, opened 2026-08-20
Feat/630 rulepack lifecycle

VietSory:
# PR Summary

Implements the detection-content lifecycle defined in #630 by introducing a deterministic, signed, and versioned RulePack workflow from authoring through validation, promotion, retrospective evaluation, and release gating.

The change establishes RulePack as a first-class detection-content artifact with explicit lifecycle state, compatibility metadata, canonical hashing, signature verification, promotion evidence, replay validation, ATT&CK coverage requirements, and purple/emulation evidence.

The goal is to make detection-content releases reproducible and auditable: the same RulePack content produces the same canonical digest, promotion decisions are backed by verifiable evidence, and release gates fail closed when required validation or provenance is missing.

## RulePack lifecycle

* Introduces versioned RulePack definitions and lifecycle state handling.
* Adds deterministic canonicalization and digest generation so logically identical content produces stable identities.
* Validates RulePack structure, matcher configuration, compatibility metadata, and lifecycle transitions before promotion.
* Adds compatibility and rollback-related metadata needed to reason about safe RulePack releases.
* Keeps promotion decisions separate from mutable runtime state so release evidence can be reviewed and reproduced.

## Signing and provenance

* Adds cryptographic signing and verification for RulePack-related artifacts.
* Promotion evidence is attested separately rather than implicitly trusted from ordinary validation output.
* Gate evidence includes provenance information required to associate a promotion decision with the exact content being evaluated.
* Invalid, missing, or mismatched signatures/digests cause validation or promotion to fail closed.

## Promotion and release gates

* Adds deterministic CI/release gates for deciding whether a RulePack is eligible for promotion.
* Promotion evaluates the required validation evidence instead of relying only on a successful build.
* Adds ATT&CK coverage checks and threshold enforcement.
* Adds purple/emulation evidence requirements for release qualification.
* Prevents incomplete or unverifiable gate evidence from being accepted as a valid promotion result.

## Corpus replay

* Adds replay support against positive and negative fixture corpora.
* Positive fixtures are evaluated before negative fixtures, with deterministic ordering within each corpus.
* Replay results participate in promotion validation so regressions in expected detections or false-positive behavior can block a release.
* Adds validation around fixture inputs and replay results to keep qualification reproducible.

## Retro-hunt support

* Adds retro-hunt evaluation support for applying RulePack content against historical evidence.
* Retro-hunt results can be incorporated into lifecycle qualification without changing the RulePack identity itself.
[...]

PR #648, opened 2026-08-20
feat: complete EPIC 583 production readiness

### LINKED ISSUE #587: Ops: verified backup/restore runbook + drill (and upgrade/rollback procedures)
> Part of #583 (production-readiness epic). Item **A4** · Priority **P0** (backup/restore), **P1** (upgrade/rollback) · Group A (Operations & Observability).

## Problem
There is no verified backup, restore, upgrade, or rollback procedure at the system level. Specifically:
- No backup script, no disaster-recovery runbook, no restore-verification procedure — only a recommendation to back up the database and the evidence object store together.
- No documented upgrade path and no binary↔schema compatibility table.
- Binary rollback **after** a newer migration has been applied is unverified; goose `down` on real data is unverified.

## Evidence
- `E-OPS-06`, `E-OPS-07` — `docs/guide/deployment.md`.
- Risks `R-06` (unrecoverable data loss), `R-08` (rollback may corrupt data).

## Impact
Report integrity depends on `pgdata` **and** the evidence object store being a **consistent pair** (report metadata lives in PostgreSQL, artifacts in the object store; either alone cannot reconstruct a report). Ad-hoc backups risk unrecoverable loss, and an untested rollback could corrupt data. DB↔object-store snapshot consistency during an active scan is currently unverified (`U`).

## Proposed work
1. **Backup/restore (P0).** Ship a documented procedure that snapshots PostgreSQL and the evidence object store at a coordinated point, plus a **restore drill** proving the evidence chain (`VerifyChain`) still passes after restore. Characterize whether a DB + object-store snapshot taken during an active scan is consistent.
2. **Upgrade (P1).** Document and rehearse a supported upgrade path (`v0.1.8` → next) on production-like data with no data loss; record binary↔schema compatibility.
3. **Rollback (P1).** Test binary rollback across a migration on a data **copy**; document whether it is safe, and if not, state the constraint explicitly (currently goose is only exercised `up`).

## Acceptance criteria
- [ ] A restore from backup succeeds and `VerifyChain` passes on restored data.
- [ ] DB + object-store snapshot consistency during an active scan is characterized and documented (moves the `U`).
- [ ] Upgrade `v0.1.8` → next rehearsed on staging with prod-like data, no data loss, result documented.
- [ ] Binary rollback across a migration tested on a data copy; outcome (safe / unsafe + constraint) documented with evidence.
- [ ] Runbook committed under `docs/guide/`.

## Dependencies
Backup/restore should exist before B3 (#590 production manifest) go-live. Rollback verification informs the upgrade story in B1/B3.

### LINKED ISSUE #590: Deploy: production deployment manifest (TLS, sandbox on, hardened, gated migration)

### LINKED ISSUE #592: Security: OIDC/SSO integration and role-model review
> Part of #583 (production-readiness epic). Item **C2** · Priority **P2** · Group C (Security Posture & Validation).

## Problem
Authorization is a **static** model with no external identity integration. Five roles and five permissions are declared in the domain layer and read once after package init. There are no custom roles, no per-resource-instance permissions, and no external identity-provider (OIDC/SSO) integration. Machine roles (`mcp`, `agent`) intentionally hold no human-REST permissions.

## Evidence
- `E-SEC-02` — `internal/domain/user/user.go` (static role/permission map); review §12.3; risk `R-15`.

## Impact
Organizations that require SSO or finer-grained authorization must compensate with network- and process-level controls, and identity management is fragmented across systems. This is an integration gap, not a safety defect — the existing RBAC + tenant isolation are implemented and tested.

## Proposed work
- Add OIDC/SSO integration: authenticate a user via an external IdP and map the resulting identity to a role.
- Evaluate custom roles and/or resource-scoped permissions for orgs that need finer granularity.
- Until SSO ships, document the compensating controls (network segmentation, proxy auth, process controls).

## Acceptance criteria
- [ ] A user can authenticate via an external OIDC provider and be mapped to a role.
- [ ] The role/permission model for the SSO case is documented.
- [ ] Compensating-control guidance is shipped for the pre-SSO state.

## Dependencies
Independent; aligns with the platform SSO/identity roadmap. Lower priority (`P2`) — the current static RBAC is safe, this is an enterprise-integration enhancement.

pho-veteran:
## Summary

- make API startup replica-safe and add bounded PostgreSQL pool observability
- add OIDC Authorization Code + PKCE BFF sessions, strict identity/role mapping, CSRF protection, and browser session UX while preserving bearer API tokens
- add digest-pinned production images, a production Helm chart, and reusable AWS staging automation
- add paired backup/restore and forward-only upgrade tooling, deterministic restore verification, benchmark reduction, and blinded FP-triage packets
- publish production topology, identity, recovery, drill-evidence, capacity, and SLO documentation

Closes #587.
Closes #590.
Closes #592.

Implements part of #589, #591, and #593 under #583, which remain open for the acceptance gates below.

## Validation

- rebased onto upstream `main` at `4c0c2f4`
- `pnpm --dir web test`
- `pnpm --dir web build`
- `go test ./docs/...`
- recovery wrapper tests
- Terraform static/validation checks
- Helm render/lint tests
- staged diff secret/path audit and `git diff --cached --check`

`make lint` was also run. It currently reports three pre-existing findings introduced by upstream commit `2800563` in `internal/infrastructure/spool/spool.go`, `internal/infrastructure/spool/state.go`, and `internal/adapter/agentspool/retry.go`; this PR does not modify those files.

The strict MkDocs build had passed before the rebase. After rebasing, the environment no longer has the `mkdocs` executable/module available; `go test ./docs/...` still passes against the rebased tree.

### #589 — representative concurrency limits

Two API replicas and failover were exercised successfully, and replica-safe startup plus bounded PostgreSQL pool telemetry are implemented. However, maximum workers per database, sustained concurrent scans per API instance, and pool saturation/recovery limits have not yet been measured under representative load. #589 therefore remains open.

### #591 — positive production sandbox execution

Production remains fail-closed when Bubblewrap cannot establish the required namespace/mount/seccomp isolation. Standard EKS managed nodes did not permit the current nested Bubblewrap transition under the non-privileged, drop-all-capabilities, RuntimeDefault posture. This PR intentionally does **not** add `privileged: true` or broad `SYS_ADMIN` to manufacture a passing result. A dedicated compatible sandbox runtime remains required before #591 can close.

### #593 — human blind review and representative capacity

The deterministic benchmark reducer and blinded packet generator are implemented. AWS testing measured only a low-rate readiness/failover path: 100/100 readiness requests and 90/90 failover probes succeeded, with readiness p50 732 ms, p95 784 ms, p99 948 ms, and max 1,546 ms. Each probe opened a fresh TLS connection, so these values are not an application-handler latency baseline.

Scan throughput, queue saturation, PostgreSQL pool waiting, evidence growth, and migration duration remain unmeasured. The blinded packet has not been submitted by a genuine human reviewer; AI/subagent review is not represented as human evidence. #593 therefore remains open.

## Scope hygiene

- root `.dockerignore` and `.gitignore` are unchanged and excluded
- local `artifacts/`, `reports/`, `.claude/`, `.serena/`, and `CLAUDE.md` are excluded
- Terraform backend/state/tfvars/plan/provider-cache files remain ignored locally
- long-lived deploy and documentation paths use responsibility-based names rather than EPIC/issue-number directories
- ADR deciders reference their governing issues (#587, #590, #592)

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -16,6 +16,19 @@
 # routes require it; GET /healthz and GET /readyz are intentionally public. Generate: openssl rand -hex 32
 SYNAPSE_API_TOKEN=
 
+# Optional browser OIDC BFF. Discovery is pinned to this HTTPS issuer; identities must already
+# be linked in the fixed tenant. Provider tokens and claims are neither persisted nor logged.
+# SYNAPSE_OIDC_ENABLED=false
+# SYNAPSE_OIDC_ISSUER=https://issuer.example
+# SYNAPSE_OIDC_CLIENT_ID=
+# SYNAPSE_OIDC_CLIENT_SECRET=
+# SYNAPSE_OIDC_REDIRECT_URL=https://synapse.example/api/auth/oidc/callback
+# SYNAPSE_OIDC_FRONTEND_URL=https://synapse.example/
+# SYNAPSE_OIDC_TENANT_ID=default
+# SYNAPSE_OIDC_GROUP_ROLE_MAPPING=synapse-admins=admin,synapse-reviewers=reviewer
+# SYNAPSE_OIDC_TRANSACTION_TTL=10m
+# SYNAPSE_OIDC_SESSION_TTL=8h
+
 # =============================================================================
 # Core / server
 # =============================================================================
--- a/cmd/synapse-api/main.go
+++ b/cmd/synapse-api/main.go
@@ -22,6 +22,8 @@ import (
 	"syscall"
 	"time"
 
+	"github.com/jackc/pgx/v5/pgxpool"
+
 	"github.com/KKloudTarus/synapse-ce/internal/adapter/httpapi"
 	"github.com/KKloudTarus/synapse-ce/internal/adapter/observability"
 	"github.com/KKloudTarus/synapse-ce/internal/domain/agent"
@@ -43,6 +45,7 @@ import (
 	"github.com/KKloudTarus/synapse-ce/internal/infrastructure/fleetca"
 	"github.com/KKloudTarus/synapse-ce/internal/infrastructure/llm/openai"
 	"github.com/KKloudTarus/synapse-ce/internal/infrastructure/logstream"
+	oidcadapter "github.com/KKloudTarus/synapse-ce/internal/infrastructure/oidc"
 	"github.com/KKloudTarus/synapse-ce/internal/infrastructure/persistence/file"
 	"github.com/KKloudTarus/synapse-ce/internal/infrastructure/persistence/memory"
 	"github.com/KKloudTarus/synapse-ce/internal/infrastructure/persistence/postgres"
@@ -138,6 +141,8 @@ import (
 	"github.com/KKloudTarus/synapse-ce/internal/usecase/fleetrolloutuc"
 	"github.com/KKloudTarus/synapse-ce/internal/usecase/fleetwork"
 	"github.com/KKloudTarus/synapse-ce/internal/usecase/fptriage"
+	identitybff "github.com/KKloudTarus/synapse-ce/internal/usecase/identitybff"
+	identityuc "github.com/KKloudTarus/synapse-ce/internal/usecase/identityuc"
 	"github.com/KKloudTarus/synapse-ce/internal/usecase/jsreach"
 	"github.com/KKloudTarus/synapse-ce/internal/usecase/leaderuc"
 	"github.com/KKloudTarus/synapse-ce/internal/usecase/llmverifier"
@@ -240,6 +245,10 @@ func main() {
 		log.Error("database migration posture invalid", "err", err)
 		os.Exit(1)
 	}
+	if err := cfg.ValidateOIDCPosture(); err != nil {
+		log.Error("OIDC posture invalid", "err", err)
+		os.Exit(1)
+	}
 
 	// Fail closed: no anonymous access. The token is never logged.
 	if cfg.APIToken == "" {
@@ -252,6 +261,7 @@ func main() {
 	acquirer := acquire.New().WithMaxWorkspaceBytes(cfg.MaxWorkspaceBytes).WithImageRootFS(cfg.ImageRootFSEnabled).WithComparisonDepth(cfg.ProjectGitComparisonDepth)
 
 	// Persistence: PostgreSQL when configured, else file + in-memory (dev).
+	var databasePool *pgxpool.Pool
 	var repo ports.EngagementRepository
 	var projectRepo ports.ProjectRepository
 	var assetStore ports.AssetRepository
@@ -269,6 +279,7 @@ func main() {
 	var commentRepo ports.CommentRepository
 	var retestRepo ports.RetestRepository
 	var userRepo ports.UserRepository
+	var identityStore ports.IdentityStore
 	var auditReader ports.AuditReader
 	var scanRepo ports.ScanRepository
 	var scanResultStore ports.ScanResultStore
@@ -379,27 +390,13 @@ func main() {
 			os.Exit(1)
 		}
 		defer pool.Close()
+		databasePool = pool
 		readinessChecks["database"] = func(ctx context.Context) error {
 			return postgres.CheckDatabaseReady(ctx, pool)
 		}
 		readinessChecks["migrations"] = func(ctx context.Context) error {
 			return postgres.CheckMigrationsReady(ctx, pool)
 		}
-		// Single-instance guard: until horizontal scaling the repos
-		// ignore tenant_id and there is no leader election, so two writers would race. A
-		// session advisory lock makes the assumption explicit + enforced – a second
-		// instance fails fast instead of silently corrupting state.
-		lockConn, ok, lerr := postgres.AcquireSingletonLock(startup, pool, "api")
-		if lerr != nil {
-			log.Error("single-instance lock check failed", "err", lerr)
-			os.Exit(1)
-		}
-		if !ok {
-			log.Error("another synapse-api instance holds the single-instance lock – this build runs ONE instance (stop the other first); horizontal scaling is P5")
-			os.Exit(1)
-		}
-		defer lockConn.Release()
-		log.Info("acquired single-instance advisory lock")
 		repo = postgres.NewEngagementRepository(pool)
 		projectRepo = postgres.NewProjectRepository(pool)
 		findingRepo = postgres.NewFindingRepository(pool)
@@ -407,6 +404,11 @@ func main() {
 		commentRepo = postgres.NewCommentRepository(pool)
 		retestRepo = postgres.NewRetestRepository(pool)
 		userRepo = postgres.NewUserRepository(pool)
+		identityStore, err = postgres.NewIdentityStore(pool)
+		if err != nil {
+			log.Error("postgres OIDC identity store init failed", "err", err)
+			os.Exit(1)
+		}
 		scanRepo = postgres.NewScanRepository(pool)
 		vulnerabilityInventory = postgres.NewComponentInventoryStore(pool)
 		scanResultStore = postgres.NewScanResultStore(pool)
@@ -491,6 +493,12 @@ func main() {
 		commentRepo = memory.NewCommentRepository()
 		retestRepo = memory.NewRetestRepository()
 		userRepo = memory.NewUserRepository()
+		var identityStoreErr error
+		identityStore, identityStoreErr = memory.NewIdentityStore(userRepo)
+		if identityStoreErr != nil {
+			log.Error("memory OIDC identity store init failed", "err", identityStoreErr)
+			os.Exit(1)
+		}
 		memoryInventory := memory.NewComponentInventoryStore()
 		scanRepo = memory.NewScanRepository(memoryInventory)
 		vulnerabilityInventory = memoryInventory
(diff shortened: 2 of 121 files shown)


nghiadaulau (changes requested):
Strict review — go-arch + security subagents + local verification on a fresh Postgres 17 (CI intentionally off). **This is a genuinely strong production-readiness PR and the OIDC/BFF/identity subsystem is excellent** — the security audit found **no critical/high/medium** issues: `coreos/go-oidc` (not hand-rolled) with ID-token sig/iss/aud/exp verified, PKCE S256, single-use state atomically burned, triple-bound nonce (server hash + `__Host-` cookie + id-token claim) compared constant-time, `__Host-` Secure/HttpOnly cookies, CSRF synchronizer token on unsafe cookie methods, a true BFF (provider tokens never persisted or sent to the browser — only an opaque, server-stored, hashed session token), secrets hashed/encrypted and sourced from env, RLS FORCE + composite tenant FKs on migration 0108, and the session resolves through the **same** `authz` chokepoint + `withEngTenant` as the bearer path (fail-closed 401). Build/vet/gofmt clean; `go test -race` green across all new packages; the PG suite (identity_store, migration 0108, RLS under a non-bypass role) passes on a fresh DB. The `sandbox/runner.go` change is a hardening self-check (argv-only, strengthens containment).

I'm requesting changes on a small, concrete set — none in the auth crypto itself.

## First, downgrading the flagged "multi-replica" concern (NOT a blocker)
Removing the API `AcquireSingletonLock` + Helm `api.replicaCount: 2` is safe as shipped, which I verified:
- The vuln **scheduler** is **fail-closed**: `cmd/synapse-api/main.go:2180-2183` `os.Exit(1)` if it's enabled on Postgres without `SYNAPSE_LEADER_ENABLED` — so it can never silently double-schedule.
- The Helm chart ships the scheduler **off** (api template sets neither the scheduler nor leader-election env), so a default 2-replica deploy neither crashloops nor double-schedules.
- The always-on API loops are competing-consumer/lease-safe on N replicas: `vulnerabilityWorker` claims via `FOR UPDATE SKIP LOCKED` (`vulnerability_action_store.go:347`), and the SCA stale-scan sweeper is lease-as-liveness + idempotent reclaim.
- 🟡 **Minor (Helm ergonomics):** because of the fail-closed guard, an operator who later enables the vuln scheduler with `api.replicaCount > 1` must also set `SYNAPSE_LEADER_ENABLED`, or the API pods crashloop. Please template/document that coupling in the chart so enabling the scheduler doesn't surprise an operator.

## 🔴 Requesting changes (must)
**Revert the `go` directive bump.** `go.mod:3` bumps `go 1.26.4 → 1.26.6`; CLAUDE.md pins **1.26.4**, and none of the three added deps need it — `coreos/go-oidc/v3` and `go-jose/v4` target `go 1.24.0`, `golang.org/x/oauth2` targets `go 1.25.0`. Set it back to `1.26.4` (and `go mod tidy`; `go-oidc`/`oauth2` also landed in the indirect region of the require block — tidy normalizes them to direct). A gratuitous toolchain-pin bump shouldn't ride in under a feature PR.

## 🟡 Should-fix
- **Absolute session-lifetime cap (security LOW).** `identitybff.Discover` → `identityuc.RotateSession` resets `ExpiresAt = now + SessionTTL` on a fresh record every poll and discards the original `CreatedAt`, so an actively-polled session never reaches a maximum age — a stolen-then-kept-alive cookie outlives the 8h intent indefinitely. For a security product, carry an immutable session origin/absolute-expiry through rotation and refuse rotation past it (force OIDC re-auth). (Revocation + `__Host-`/HttpOnly + rotation keep this LOW, but it's worth closing.)
- **Missing compile-time port assertion.** `internal/infrastructure/oidc/protector.go` — add `var _ ports.IdentitySecretProtector = (*SecretProtector)(nil)` (every sibling adapter has one; CLAUDE.md requires it so signature drift breaks `go build`).
- **Validate `OIDCFrontendURL` at startup (security LOW-2).** It's correctly a fixed config target (not an open redirect), but validate it parses as an absolute http(s) URL at load and fail closed on empty/garbage, mirroring the issuer check.

## 🟢 Minor / optional
- `internal/usecase/ports/ports.go`: the new `Restore*` types were inserted between the `BlobStore` doc comment and `type BlobStore interface`, so that comment now documents `RestoreEvidenceChain`. Move the `Restore*` block out of the way.
- `identitybff` depends on the concrete `*identityuc.Service` (sibling use case) rather than a port — not an inward violation, but extracting the small consumed surface into an interface would keep the ports discipline strict. Non-blocking.

Fix the `go` directive (must) + ideally the session cap + port assertion, and I'll re-review — the auth subsystem is ready.

nghiadaulau (approved):
✅ **Approve + merge.** I applied the review fixes directly (commit `501a4e0`, merged current main in `9f9a5b0`) and re-verified locally (CI intentionally off):

- **Must — `go` directive reverted** `1.26.6 → 1.26.4` (CLAUDE.md pin; no added dep needs it — go-oidc/go-jose target 1.24, oauth2 targets 1.25); `go mod tidy` clean.
- **LOW-1 — absolute session-lifetime cap implemented.** `identity.Session` now carries an immutable `OriginAt` through every rotation; `RotateSession` refuses to rotate a lineage past `identity.MaxSessionAge` (12h) — forcing OIDC re-auth — so a rotated/stolen-then-kept-alive cookie can no longer outlive its intent. New `origin_at` column on `oidc_sessions` (migration 0108, unshipped), persisted by the postgres store, carried by the memory store, fail-closed on a zero origin. Regression test `TestRotateSessionEnforcesAbsoluteMaxAge` (within-cap rotates + carries origin; past-cap → `ErrForbidden`).
- **Port assertion** `var _ ports.IdentitySecretProtector = (*SecretProtector)(nil)` added (build-time).
- **Doc comment** for `BlobStore` moved back onto its interface in `ports.go`.
- **LOW-2** (frontend-URL validation) was already enforced by `ValidateOIDCPosture` at startup — no change needed.
- **go-arch "multi-replica" concern** confirmed a non-issue: the vuln scheduler is fail-closed (`os.Exit` without leader election on Postgres), the chart ships it off by default, and the always-on loops are competing-consumer safe (`FOR UPDATE SKIP LOCKED` / lease-as-liveness). Left a note to template the scheduler↔leader-election coupling in the chart.

Re-verified on the merged tree: `go build ./...` clean, `gofmt -l` clean, `go test -race` green across identity/oidc/httpapi/identitybff + rulepack, and the PG suite (migration 0108 with `origin_at`, identity store round-trip, RLS under a non-bypass role) passes on a fresh Postgres 17 DB. Merged current main (rulepack #647) — the only conflict was an mkdocs nav entry, resolved keeping both.

The OIDC/BFF/identity subsystem was already security-CLEAN (no critical/high/medium). Merging — this lands EPIC 583 production readiness.

PR #649, opened 2026-08-21
feat(release): add deterministic evidence manifest

### LINKED ISSUE #556: [Mini Epic] Establish verifiable release provenance and artifact authenticity
# [Mini Epic] Establish verifiable release provenance and artifact authenticity

## Context

The repository already has a fairly complete security verification pipeline:

* dependency auditing;
* repository vulnerability, secret, and misconfiguration scanning;
* backend/frontend container scanning;
* CycloneDX dependency and image SBOM generation;
* SARIF upload to GitHub code scanning;
* pinned third-party workflow actions;
* a documented release process that records commit SHAs, image/model digests, CI/security runs, and release evidence.

However, there is still one documented gap between **artifact integrity** and **artifact authenticity**.

For example, the current ML bundle verifies artifact digests against `metadata.json`. This detects accidental or unauthorized modification relative to that metadata, but it does not establish who produced or approved the bundle.

The README already identifies release signing/attestation as a future hardening step, and the release runbook expects artifacts to move through a trusted release channel.

This epic proposes closing that gap without changing the application's detection model or deployment architecture.

## Problem

Today we can answer:

> "Does this artifact match the checksum recorded in its metadata?"

but we cannot reliably answer:

> "Was this exact artifact produced by the expected repository/release workflow from the expected source revision?"

Similarly, SBOMs and other security evidence are generated during CI, but there is no single verifiable relationship tying the promoted release evidence back to the exact release revision.

That leaves the release chain weaker than the existing build and security controls.

## Goal

Create a small, auditable release-provenance layer that binds promoted artifacts and security evidence to the exact source revision that produced them.

The intended flow becomes:

```text
source revision
      |
      v
quality + security gates
      |
      v
release artifacts / SBOMs
      |
      v
release evidence manifest
      |
      v
provenance / attestation
      |
      v
verification before promotion
```

The objective is not to introduce another security scanner.

The objective is to make the evidence already produced by the repository **traceable and verifiable as part of the release process**.

---

### 1. Generate a deterministic release evidence manifest

Add a machine-readable manifest for a release candidate containing the relevant evidence, for example:

* repository commit SHA;
* release/tag identifier;
* CI/security workflow identifiers where available;
* SBOM digests;
* promoted artifact digests;
* container image digests when images are published;
* ML `metadata.json` digest when a model bundle is part of the release;
* relevant build/runtime metadata.

The manifest should reference generated artifacts rather than committing generated runtime state into Git.

**Candidate change**

```text
ci: generate release evidence manifest
```

### 2. Add provenance/attestation for promoted release artifacts

Attach verifiable provenance to artifacts that are actually promoted through the release process.

Prefer an identity-based CI mechanism where possible rather than introducing a long-lived signing secret.

The implementation should remain limited to release artifacts/evidence; ordinary PR artifacts do not need to become signed release assets.

**Candidate change**

```text
ci: attest promoted release artifacts and security evidence
```

### 3. Add release verification

Provide a deterministic verification step that validates:

* the evidence belongs to the expected repository/revision;
* artifact digests match the release manifest;
* required provenance/attestation is valid;
* an artifact from a different build/revision cannot silently be substituted.

This can initially live in the release workflow/runbook rather than adding provenance verification to normal application runtime.

**Candidate change**

```text
ci: verify release provenance before promotion
```

### 4. Document the trust boundary

Update the release and threat-model documentation to distinguish:

```text
checksum integrity
        !=
publisher authenticity
        !=
runtime trust
```

Document what is attested, what identity is trusted, how verification is performed, and what remains outside the guarantee.

**Candidate change**

```text
docs: define release provenance and artifact trust model
```

---

## Acceptance criteria

* [ ] A release candidate can produce a machine-readable evidence manifest tied to an exact commit SHA.
* [ ] Promoted release artifacts/evidence have cryptographically verifiable provenance or attestation.
* [ ] Verification fails when the artifact digest or expected source revision does not match.
* [ ] Existing SBOM and security scanning remain part of the release evidence rather than being duplicated by a second pipeline.
* [ ] Generated model files, SBOMs, scan output, and runtime state remain outside source control unless explicitly promoted as release artifacts.
* [ ] The release runbook documents both the verification procedure and its trust assumptions.
* [ ] Existing local development and normal PR workflows continue to work without requiring signing infrastructure.

---

## Non-goals

This epic intentionally does **not** include:

* replacing the current CI/security scanners;
* introducing a container registry solely for this work;
* signing every temporary CI artifact;
* changing the rule or ML detection algorithms;
* changing the current SQLite/single-node deployment boundary;
* treating provenance as proof that an artifact is vulnerability-free;
* committing generated ML models or release artifacts to the repository;
* claiming supervised detection accuracy without representative labeled data.

---

## Why this fits the project

This is an extension of controls that already exist rather than a new product direction.

The repository already:

1. pins dependencies and workflow actions;
2. scans dependencies, source, and images;
3. generates SBOMs and SARIF;
4. records artifact hashes;
5. verifies ML artifact integrity;
6. documents a controlled release process.

Provenance/attestation is the missing link between those controls and the final artifact that an operator is expected to trust.

It also closes an explicitly documented limitation: hashes prove consistency with metadata, but not publisher authenticity.

Keeping the work as a small release-security epic lets us improve that boundary incrementally without expanding the runtime architecture or changing the application's core purpose.

## Suggested implementation order

```text
release evidence manifest
          ↓
artifact/evidence attestation
          ↓
release verification gate
          ↓
trust-model + runbook documentation
```

Each step can be reviewed independently and should leave the repository in a usable state.

XUanhoa04:
## Summary

- add a deterministic `synapse-release-evidence-v1` manifest binding the complete promoted asset set to the repository, release tag, and exact source revision
- add a fail-closed generator/verifier that streams SHA-256 hashes and rejects identity drift, tampering, missing or extra assets, unsafe names, symlinks/non-regular files, unknown fields, duplicate/unsorted entries, and non-canonical JSON
- wire the manifest into release signing: rebuild derived evidence on retry, remove the existing `SHA256SUMS` self-reference, sign and locally verify the manifest, require its detached signature in the release gate, and include it in GitHub provenance attestation
- document the separate guarantees of checksum integrity, GPG publisher authenticity, and GitHub workflow provenance, including an exact operator verification procedure

## Why

The release pipeline already produced signatures, SBOMs, checksums, and GitHub attestations, but there was no single deterministic object tying all promoted bytes to one repository/tag/revision. Consumers could verify individual pieces without proving that the downloaded directory was the complete intended release set.

The manifest deliberately excludes itself and `release-evidence.json.sig` to avoid a recursive digest. Its detached signature authenticates the manifest; the manifest then authenticates every other release asset. GitHub attestation independently binds the signed manifest to the release workflow identity.

Manual signing runs now fail unless the workflow identity, checked-out commit, and tag commit agree. This prevents a dispatch from a moving branch from producing provenance for a different release source.

## Verification

- `go test ./...`
- `go vet ./...`
- `go build ./cmd/...`
- `pnpm run typecheck` in `web/`
- `go test ./internal/infrastructure/releaseevidence ./cmd/synapse-release-evidence -count=1`
- `go run github.com/rhysd/actionlint/cmd/actionlint@v1.7.12 .github/workflows/release-sign.yml .github/workflows/release-gate-negative.yml`
- `git diff --check`
- staged LF shell source passes `bash -n`

Focused race tests were not available on this Windows host because the Go race build requires cgo and no C compiler is installed. The new package has no mutable global state and the complete non-race suite is green.

Closes #556


nghiadaulau (approved):
✅ **Approve + merge.** Strict review — go-arch + security subagents + local verification (CI intentionally off). This is a carefully-built, fail-closed release-integrity feature and the substance is **clean on both reviews and my own read**.

**Core (`internal/infrastructure/releaseevidence` + `cmd/synapse-release-evidence`):**
- **Canonical, injective encoding** — fixed struct field order, artifacts sorted + strictly-ascending-unique, fixed-width hex digest, decimal size, `SetEscapeHTML(false)`, no maps, no timestamp; and `Verify` re-encodes and `bytes.Equal`s against the raw file, so reordering, field-boundary shifts, duplicate JSON keys, whitespace, `1e3` numbers, unknown fields, and trailing data are all rejected as non-canonical.
- **Fail-closed verification** — identity/revision/tag drift, hash/size/kind mismatch, missing OR extra assets, unsafe/traversal names, symlinks / non-regular files, duplicate/unsorted entries all return an error; the CLI prints "verified" only on `nil`.
- **Path & TOCTOU safety** — lstat-rejects a symlinked root, pins the manifest to a direct child, streams SHA-256 via `io.Copy` (bounded memory), and fences each read with `SameFile` + size/mtime re-checks. Bounded (4096 assets / 4 MiB / 255-char names); `O_EXCL` 0600 write.
- **Self-exclusion is not a smuggling vector** — only the manifest + its exact `.sig` are excluded; any other unlisted file fails the count/name check. The `.sig` is authenticated separately by the mandatory `gpg --verify`.
- **dependency rule PASS** (stdlib-only library; cmd is a pure composition root; no cycle, no server wiring). Build/vet/gofmt clean; `go test -race` green.

**Recommended follow-ups (not blocking — none affect the merged Go feature; Actions is disabled repo-wide so there is zero current runtime exposure):**
- 🟡 **security LOW (own PR):** `release-sign.yml` interpolates `${{ github.event.release.tag_name || inputs.tag }}` (and the derived `steps.tag.outputs.tag`) directly into `run:` shell — the standard Actions script-injection sink, in a job with `contents/id-token/attestations: write`. It's **largely pre-existing** (this PR extends it), so — consistent with keeping one concern per PR — please harden it in a focused workflow PR: route untrusted values through `env:` and reference `"$TAG"` in the script. (Realistically low: needs repo write + the Go tool re-validates `--release`/`--revision`.)
- 🟢 go-arch minors (optional): `Artifact.Kind` is fully derived from `Name` and validated on both sides — a one-line doc note ("derived from Name; validated") would prevent a reader assuming it's authoritative; and `Generate`/`Verify` don't thread `context.Context` (defensible for a one-shot CLI).
- pre-existing (noted, out of scope): `verify-artifacts-signed.sh` checks `.sig` existence, not validity — actual validity is enforced by the separate `gpg --verify`.

Merging — this closes #556 (deterministic release evidence).

PR #650, opened 2026-08-22
fix(ebpf): add native arm64 CO-RE artifacts

### LINKED ISSUE #631: Cross-cutting · Agent integrity & compatibility matrix (arm64/CO-RE-BTF/tamper/secure-update; fixes D5)

XUanhoa04:
## Summary
- build and embed separate amd64 and arm64 eBPF artifacts, replacing the unreproducible `vmlinux.h` dependency with a minimal repository-owned CO-RE header
- probe host architecture, kernel BTF, and required CO-RE types directly; report explicit network coverage gaps while keeping the other sensor classes independent
- add structural and native ABI tests, a reproducible eBPF build target, portability documentation, and native amd64/arm64 CI coverage

## Scope
This is the D5/arm64 portability slice of #631. It does not implement watchdog/tamper resistance, secure update, key protection, or the full OS/runtime matrix, so it does not close #631.

## Testing
- `go build ./...` on Linux and Windows
- `go vet ./...` on Linux and Windows
- `go test -count=1 ./internal/infrastructure/ebpf` on Linux and Windows
- privileged native x86_64 load test: all process, network, file, and privilege classes loaded
- privileged native x86_64 network test: a real UDP call was captured and decoded as port 53
- privileged native x86_64 end-to-end sensor/domain-rule test
- `GOOS=linux GOARCH=arm64 go test -c ./internal/infrastructure/ebpf`
- `shellcheck scripts/ebpf/build.sh`
- `golangci-lint v2.12.2 run ./internal/infrastructure/ebpf/...` (0 issues)

The added GitHub Actions matrix performs the privileged load/attach and UDP argument-decoding tests natively on both amd64 and arm64 runners.

One unrelated local full-suite exception: the Helm render test could not execute from the Windows-mounted WSL checkout because `render_test.sh` had CRLF line endings (`set: Illegal option -`). All other packages passed; the script is unchanged by this PR and GitHub's Linux checkout uses repository LF attributes.


nghiadaulau (approved):
✅ **Approve + merge.** Strict review — go-arch + security subagents + local verification (CI intentionally off). The verifiable-here parts are **clean on both reviews and my own read**, and this correctly fixes D5 (removes the `-D__TARGET_ARCH_x86` pin → per-arch objects).

**Verified on this host:**
- **Build-tag partition correct** — `linux&&amd64` / `linux&&arm64` / `linux&&!amd64&&!arm64` (stub) give every linux GOARCH exactly one `embeddedObjectArch` + `*Obj` definition, no overlap/gap; non-linux uses the existing `_other` files. `go build`/`go vet` clean native (darwin) and cross for **linux/amd64 + linux/arm64** (+ the `linux/arm` stub path); the linux test binaries compile for both shipped arches; darwin unit tests pass; gofmt clean.
- **Fail-closed capability probe** — `CORE=true` only after arch-match + BTF load + required types (`sock`/`msghdr`/`sockaddr_in`); any miss → actionable `Reason`, `CORE=false`. Table-tested.
- **Class independence preserved** — unsupported arch (`embeddedObjectArch==""`) → FAILED gaps + `ErrSensorUnavailable`; a network `requiresCORE && !CORE` miss records a **DEGRADED network gap and continues** so process/file/privilege still load (they're stable-layout tracepoints). `detect_linux.go` change is arch-selection + probe only — decoders/attach/lifecycle untouched; load remains the authoritative final test.
- **C safety (read)** — `netconn.bpf.c` uses `BPF_CORE_READ` only (no raw user-ptr deref), fixed-size ringbuf reserve, bounded `bpf_get_current_comm`; `core_types.bpf.h` declares only the read fields under `preserve_access_index` and the arch `pt_regs` offsets are asserted by `TestNetworkObjectsUseTargetCallingConvention`. Object bytes come only from compile-time `go:embed` (no runtime path-load from a writable location). No secrets in the build script or probe reasons.
- The committed `.bpf.o` are real eBPF ELF; netconn genuinely differs per-arch (the D5 fix), the tracepoint objects are arch-neutral.

**Cannot verify on this macOS reviewer host (factored into the decision):** the `.bpf.o` **byte content vs source** (no clang/bpftool) and the **privileged native load/attach + real UDP decode** (Linux+root+amd64/arm64 only). Those rest on the contributor's native CI matrix + the reproducible `scripts/ebpf/build.sh`. Committing prebuilt eBPF blobs is the **pre-existing** repo pattern; this PR *improves* it (repo-owned build, drops the unreproducible `vmlinux.h`, removes the x86 pin).

**Recommended follow-up (not blocking; CI-side, and Actions is off):**
- 🟡 **security LOW (F1) — tighten blob reproducibility.** Pin the clang/LLVM major version in `build.sh` + the compat workflow (BPF codegen isn't byte-stable across versions), and add a gate that runs `make ebpf-generate` then `git diff --exit-code -- internal/infrastructure/ebpf/c/*.bpf.o` so a committed root-loaded object that doesn't reproduce from source fails. Re-run the `ebpf-compat` load/attach job out-of-band before relying on the arm64 sensor in production.
- 🟢 minors (optional): `TestNativeBuildEmbedsOnlyMatchingObjects` could `t.Skip` (not `t.Fatal`) on an unsupported linux arch; per-field docs on `HostCapabilities`.

Scope is honest — the PR states it's the D5/arm64 slice and does **not** close #631 (watchdog/tamper/secure-update/full matrix remain). Merging the portability slice.

PR #653, opened 2026-08-22
feat(fleet): ship signed detections from the agent (#625)

### LINKED ISSUE #625: A4 · Detection transport & signing-key registration/resolver wiring (carries A0.1/A0.2/A0.5 tails)
> **Phase A4** · child of #594 · **P0** · depends on: A3 (#624), A0.1/A0.2/A0.5 (#606/#607/#610). Blocks: A5 (#626), A7 (#628), C (#638).
> **Carries deferred A0 tails:** A0.1 server-verify for detection batches (#606) · A0.2 signing-key **registration endpoint + proof-of-possession over the wire + rotate/revoke routes + wire the resolver** (#607) · A0.5 **construct `detectledger.NewService`** in the composition root (#610).

## Goal
Bring the **detection-batch ingest path** live and independent of the telemetry queue, wiring the A0 primitives that today exist only behind interfaces/tests. This is the phase that finally composes `detectledger.NewService`, the real signing-key resolver, and the agent-plane key-registration endpoint.

## Scope
- **Agent signing-key registration endpoint** (#607): agent posts its Ed25519 signing public key + a `fleetagent.ProveKeyPossession` proof bound to its canonical `AgentID`; handler **MUST** call `fleetagent.VerifyKeyPossession` before `ports.AgentSigningKeyStore.Register` (the port doc already mandates this). Operator **rotate/revoke** routes (RBAC: PermAdminister). Rotation-with-overlap + anti-rollback already enforced by the store (#616/migration 0105).
- **Construct `detectledger.NewService`** at the composition root (`cmd/synapse-api`) — today only `NewReader` is wired; the writer is referenced only by tests (A0.5 anchor note). Inject the postgres `AgentSigningKeyRepository` as its `AgentKeyResolver` (method already named `ResolveSigningKey`, so the store satisfies it directly).
- **Detection-batch ingest**: independent detection queue (not blocked by telemetry backlog); envelope carries `KeyID`, `TelemetryRef { EventID, AgentSessionID, StreamID, BatchID, BatchSequence, Offset, EventDigest, Role, OccurredAt }`, `RulePackID/RuleVersion/RulePackDigest/DetectorVersion`; server-verify identity + key + `VerifyBatchWithKey` fail-closed; idempotent seal via A0.5 `SealOnce` (the concrete atomic impl lands in A5, wired here).
- Canonical `AgentID` on the sealed detection (#606) — never display name.

## Target (files/interfaces)
- `internal/adapter/httpapi` — agent-plane `POST /fleet/keys` (register), operator `POST /agents/{id}/keys/{keyID}/revoke` + rotate; agent-plane detection-ingest route.
- `cmd/synapse-api/main.go` — replace the A0.5/A0.2 wiring anchor notes with real construction of `detectledger.NewService(records, chain, resolver=AgentSigningKeyRepository, audit, clock, ids, retention)` + `SetDetectionIngest`.
- Reuses `internal/domain/fleetagent/signingkey.go`, `internal/usecase/fleet/detectledger`, `internal/usecase/ports/agent_signing_key.go`.
- Reserve slots: `AgentSigningKey` (live), `TelemetryRef` (owned here — the detection→telemetry citation), `DetectionEvidenceEnvelope` (sealed in A5).

## Tests
- Registration: a valid PoP proof registers the key; a proof made with a different private key, or a re-bound agent/purpose, is refused (`VerifyKeyPossession`); registering without PoP is impossible via the endpoint.
- Rotate-with-overlap: old + new keys both verify during the window; a revoked/expired key fails verification (fail-closed) — end-to-end through the live resolver.
- Detection ingest: identity + KeyID + signature verified server-side; wrong-agent / wrong-purpose / unknown-key / tampered batch all refused + audited; **logically** idempotent re-delivery seals once against the `SealOnce` interface. (The crash-atomic **D3 closure** — a seal that survives a crash between chain-write and projection-write — is A5's (#626) concrete `evidence.Service.SealOnce`, NOT claimed done here.)
- `detectledger.NewService` is actually constructed (composition-root test / smoke); detection queue is independent of telemetry backlog.
- Postgres RLS on `agent_signing_keys` under the live path; `-race`.

## Acceptance / exit
- [ ] Agent registers a signing key with proof-of-possession via a live endpoint; control plane resolves by KeyID and verifies fail-closed (closes #607 acceptance).
- [ ] Rotation-with-overlap + revoke/expire enforced end-to-end.
- [ ] `detectledger.NewService` composed in production; detections sealed exactly once with canonical AgentID + KeyID (closes #606 detection-side + #610 wiring).
- [ ] Detection transport independent of the telemetry queue.

## Non-goals
The concrete atomic `evidence.Service.SealOnce` + `DetectionEvidenceEnvelope` sealing/retention (#626); rulepack lifecycle (#630).

Part of #594.

XUanhoa04:
## Summary

Completes the remaining agent-side tail of #625 after the server ingest, key resolver, and registration routes landed in #652. Confirmed detections now leave the agent through an independent, crash-recoverable P1 delivery loop instead of remaining only in the local WAL.

Closes #625.

## Changes

- Add an independent P1 detection shipper with durable batch sequence/membership, bounded batches, exact retry after lost responses, and WAL ACK only after complete 2xx admission.
- Generate purpose-bound Ed25519 keys on the agent, persist private state with 0600 permissions and atomic fsync/rename publication, register through proof-of-possession, rotate before expiry, and rotate once on a rejected/revoked key.
- Extend the fleet client with the live key-registration and detection-ingest requests plus typed HTTP/network errors for the existing capped-jitter retry policy.
- Let dedicated transports peek one priority lane so P0/P2/P3 backlog cannot consume the detection shipper's read budget.
- Share the detection item wire contract between the agent and detectledger; collapse identical duplicate WAL records and fail closed on conflicting content for one detection ID.
- Wire the shipper into `synapse-agent` behind `SYNAPSE_DETECTION_ENGAGEMENT_ID`, document its configuration and operations, and complete OpenAPI coverage for the live fleet ingest/key routes.
- Add a live-path integration test covering durable WAL -> HTTP key PoP registration -> signed batch -> resolver/signature/content verification -> exactly-once evidence sealing -> P1 ACK.

## Safety / recovery behavior

- Pending membership is persisted before the network call and rebuilt unchanged after restart.
- A lost response reuses the same key, batch sequence, and membership; server `SealOnce` handles the replay.
- A successful server response never advances local sequence until P1 ACK succeeds; restart detects an ACKed pending batch and completes state without resending.
- Corrupt key/state, mismatched WAL membership, wrong-agent detections, conflicting duplicate IDs, and a second 403 all stop fail-closed with the WAL intact.

## Checklist

- [x] `make build vet typecheck` passes
- [x] `go test -count=1 ./...` passes
- [x] Targeted Linux `go test -race -count=1` passes for fleetagent, detectionship, agentstate, fleetclient, spool, httpapi, and synapse-agent
- [x] Preserves safety invariants: credential-derived identity, private keys stay local and out of logs, signed content membership, and ACK-after-admission

### Documentation

- [x] New `SYNAPSE_*` variables are in `docs/guide/configuration.md`
- [x] New agent flags are documented with their corresponding environment variables
- [x] Live fleet HTTP routes are described in `api/openapi.yaml` without adding coverage debt
- [x] Operational setup and failure/recovery behavior are documented in the existing fleet blue-team guide
- [x] No dashboard changes (`pnpm run typecheck` passes)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,13 @@ and this project aims to adhere to [Semantic Versioning](https://semver.org/spec
 
 ### Added
 
+- **Independent signed agent detection delivery.** Confirmed detections now drain from an isolated P1
+  WAL lane into crash-recoverable batches, using an agent-owned Ed25519 key registered with
+  proof-of-possession. Pending sequence/membership survives restart and lost responses, local records
+  are ACKed only after complete server admission, rejected or expiring keys rotate without changing
+  the pending batch sequence, and a live-path test covers WAL → HTTP → key resolution → signature and
+  content verification → exactly-once evidence sealing.
+
 - **Native amd64 and arm64 eBPF sensor artifacts with capability-based CO-RE probing.** The agent now
   embeds only architecture-matched objects, detects kernel BTF and required network types without
   kernel-version gating, reports unsupported capability as an explicit coverage gap, and validates
--- a/api/openapi.yaml
+++ b/api/openapi.yaml
@@ -42,6 +42,8 @@ tags:
   description: Tenant-scoped vulnerability feeds, exposure correlation, risk transitions, and durable operations
 - name: remediation-sla
   description: Risk-based remediation deadlines, immutable assessments, policy versions, and human-owned lifecycle governance
+- name: fleet
+  description: Enrolled-agent transport and operator fleet administration
 
 paths:
   /healthz:
@@ -1749,6 +1751,141 @@ paths:
         '404': {$ref: '#/components/responses/NotFound'}
         '409': {description: Lifecycle version changed since it was loaded, content: {application/json: {schema: {$ref: '#/components/schemas/Error'}}}}
 
+  /api/v1/fleet/telemetry:
+    post:
+      tags: [fleet]
+      operationId: ingestFleetTelemetryBatch
+      summary: Verify and ingest a signed canonical telemetry batch
+      description: A3 transport endpoint. Verifies credential identity, key lifecycle, schema, manifest signature, event digests, incarnation-aware sequencing, and returns the highest-contiguous ACK.
+      requestBody:
+        required: true
+        content:
+          application/json:
+            schema:
+              type: object
+              additionalProperties: false
+              required: [Manifest, Events]
+              properties:
+                Manifest:
+                  type: object
+                  description: fleetagent.TelemetryBatchManifest with a signed StreamPosition and explicit loss counts.
+                  additionalProperties: true
+                Events:
+                  type: array
+                  items:
+                    type: object
+                    additionalProperties: false
+                    required: [EventID, Class, Payload, ObservedAt]
+                    properties:
+                      EventID: {type: string}
+                      Class: {type: string, enum: [process, network, file, privilege]}
+                      Payload: {type: string, contentEncoding: base64, description: Canonical telemetry event bytes}
+                      ObservedAt: {type: string, format: date-time}
+      responses:
+        '200':
+          description: Batch classified and highest-contiguous ACK returned
+          content:
+            application/json:
+              schema:
+                type: object
+                required: [accepted, duplicate, ack, provenance, gap_open]
+                properties:
+                  accepted: {type: boolean}
+                  duplicate: {type: boolean}
+                  ack: {type: integer, format: uint64, minimum: 0}
+                  provenance: {type: string}
+                  gap_open: {type: boolean}
+        '400': {$ref: '#/components/responses/BadRequest'}
+        '401': {$ref: '#/components/responses/Unauthorized'}
+        '403': {$ref: '#/components/responses/Forbidden'}
+        '404': {$ref: '#/components/responses/NotFound'}
+
+  /api/v1/fleet/keys:
+    post:
+      tags: [fleet]
+      operationId: registerFleetSigningKey
+      summary: Register an agent content-signing key with proof-of-possession
+      description: The canonical agent identity comes from the bearer credential, never the request body.
+      requestBody:
+        required: true
+        content:
+          application/json:
+            schema: {$ref: '#/components/schemas/AgentSigningKeyRegistration'}
+      responses:
+        '201':
+          description: Key registered or idempotently re-registered
+          content:
+            application/json:
+              schema: {$ref: '#/components/schemas/AgentSigningKeyRegistrationResult'}
+        '400': {$ref: '#/components/responses/BadRequest'}
+        '401': {$ref: '#/components/responses/Unauthorized'}
+        '403': {$ref: '#/components/responses/Forbidden'}
+        '404': {$ref: '#/components/responses/NotFound'}
+
+  /api/v1/fleet/detections:
+    post:
+      tags: [fleet]
+      operationId: ingestFleetDetectionBatch
+      summary: Verify and ingest an independently signed agent detection batch
+      description: Verifies authenticated agent identity, key lifecycle, signature, exact membership, and per-detection content digests before exactly-once evidence sealing.
+      requestBody:
+        required: true
+        content:
+          application/json:
+            schema: {$ref: '#/components/schemas/AgentDetectionBatchRequest'}
+      responses:
+        '200':
+          description: The complete batch was admitted or idempotently skipped
+          content:
+            application/json:
+              schema: {$ref: '#/components/schemas/AgentDetectionBatchResult'}
+        '400': {$ref: '#/components/responses/BadRequest'}
+        '401': {$ref: '#/components/responses/Unauthorized'}
+        '403': {$ref: '#/components/responses/Forbidden'}
+        '404': {$ref: '#/components/responses/NotFound'}
+
+  /api/v1/agents/{id}/keys:
+    parameters:
+    - $ref: '#/components/parameters/AgentId'
+    get:
+      tags: [fleet]
+      operationId: listAgentSigningKeys
+      summary: List an agent's content-signing keys
+      responses:
+        '200':
+          description: Signing-key lifecycle records
+          content:
+            application/json:
+              schema: {$ref: '#/components/schemas/AgentSigningKeyList'}
+        '401': {$ref: '#/components/responses/Unauthorized'}
+        '403': {$ref: '#/components/responses/Forbidden'}
+        '404': {$ref: '#/components/responses/NotFound'}
+
+  /api/v1/agents/{id}/keys/{keyID}/revoke:
+    parameters:
+    - $ref: '#/components/parameters/AgentId'
+    - $ref: '#/components/parameters/SigningKeyId'
+    post:
+      tags: [fleet]
+      operationId: revokeAgentSigningKey
+      summary: Revoke one agent content-signing key
+      responses:
+        '200':
+          description: Key revoked monotonically
+          content:
+            application/json:
+              schema:
+                type: object
+                required: [agent_id, key_id, state]
+                properties:
+                  agent_id: {type: string}
+                  key_id: {type: string}
+                  state: {type: string, const: revoked}
+        '400': {$ref: '#/components/responses/BadRequest'}
+        '401': {$ref: '#/components/responses/Unauthorized'}
+        '403': {$ref: '#/components/responses/Forbidden'}
+        '404': {$ref: '#/components/responses/NotFound'}
+
 components:
   securitySchemes:
     bearerAuth:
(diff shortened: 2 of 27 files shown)


nghiadaulau (changes requested):
Thanks for this — it's a strong, well-structured completion of the #625 agent-side tail. I reviewed it strictly (two independent passes: clean-architecture + security) and validated locally since CI is off: `go build ./...`, `go vet ./...`, `gofmt` all clean, and `go test -race` green across fleetagent, detectionship, detectledger, agentstate, fleetclient, spool, httpapi, synapse-agent, ports.

The core design holds up well: WAL ACK strictly after a complete 2xx, membership+sequence persisted before the network call, lost-response replay delegated to server `SealOnce`, private key handling (0600 + fsync/rename + parent-dir fsync, fail-closed on corruption, never logged), PoP bound to the credential-derived agent id, exactly-once rotation on a 403, and correct `batch.AgentID == authenticated agent` alignment with the A4 ingest. Nice work.

There are **no double-ship or lost-detection windows** — everything below fails safe. But two items touch attribution integrity and the crash-recovery headline of this PR, so I'd like them addressed before merge, plus a few polish nits.

### Please address before merge

**1. Pending batch is not bound to the configured engagement (attribution).**
`internal/usecase/fleet/detectionship/service.go` — `State`/`PendingBatch` persist sequence/epoch/through/eventIDs + the key, but not `EngagementID`; `NewService` only guards the key's `AgentID` against `cfg.AgentID`, never `cfg.EngagementID` against a pending batch. If an operator changes `SYNAPSE_DETECTION_ENGAGEMENT_ID` while a batch is pending (unshipped / lost-response), `DeliverOnce` re-signs the *same* detections with the *new* engagement and ships them there — a cross-engagement re-attribution of already-confirmed detections. On a chain-of-custody platform that's worth closing even though it's operator-gated. Fix: persist `EngagementID` in `State`/`PendingBatch` and fail closed in `NewService`/`DeliverOnce` when the configured engagement differs from a pending batch's, mirroring the existing `AgentID` guard.

**2. Hard reboot inside the ACK window wedges the shipper (crash-recovery).**
`internal/usecase/fleet/detectionship/service.go` `pendingAlreadyACKed` recognizes an ACK only when `lane.CurrentEpoch == s.state.Pending.Epoch`, and `SpoolStats.HighestACKed` is reported for the current epoch only. A machine reboot bumps `CurrentEpoch` (Boot changes). So the sequence "batch delivered + WAL ACKed → crash before `completePendingState` → power-loss/reboot" leaves `Pending.Epoch < CurrentEpoch`: `pendingAlreadyACKed` returns false, the ACKed records are already reclaimed, and `recordsForBatch` then returns `ErrDeliveryState` — a permanently stuck, fail-closed shipper needing manual state-file removal. No data loss (the batch was server-admitted before ACK), but it's an availability gap on exactly the power-loss path this PR advertises as recoverable. Fix: expose the ACK for `Pending.Epoch` specifically (a per-epoch `Stats` field or a spool "acked-through(priority, epoch)" query) so a past-epoch delivered batch is recognized instead of wedging.

### Nits (nice to have, not blocking)

3. `internal/infrastructure/agentstate/detection.go` / `fleetclient` — the `StateStore`/`Transport` ports live in the `detectionship` usecase package rather than `internal/usecase/ports` (the repo convention is that all I/O ports live in `ports`). Also `fleetclient.Client` has no `var _ detectionship.Transport = (*Client)(nil)` compile-assertion — the binding is only checked at the `cmd` wiring site; `agentstate` has the `StateStore` assertion, please add the matching one for `Transport`.
4. `internal/infrastructure/agentstate/detection.go` imports `fleetclient.SecretModeEnforced()` (infra→infra) for a one-liner (`runtime.GOOS != "windows"`); it belongs in `internal/platform`.
5. `internal/domain/fleetagent/batch.go` `DetectionBatchItem.Validate` folds the inner `Detection.Validate()` error in with `%v`, dropping its (already `ErrValidation`-wrapped) chain — prefer `%w`/`errors.Join` so the inner context stays inspectable.

Happy to re-review promptly once 1 and 2 are in. (CI stays off — I'll re-validate locally.)

XUanhoa04:
@nghiadaulau Thanks for the strict review — all blocking items and the three nits are addressed in `b561fd9` (the PR is also rebased onto current `main`, including #655).

1. **Engagement attribution binding:** `DetectionPendingBatch` now persists `EngagementID`; state validation requires it, `NewService` rejects a configured/pending mismatch, `DeliverOnce` repeats the guard, and the signed batch uses the persisted engagement rather than re-reading mutable configuration. Covered by `TestPendingEngagementMismatchFailsClosed` and the lost-response state assertion.
2. **ACK-window reboot recovery:** `SpoolStats.EpochACKs` now exposes the durable ACK map for every `(priority, epoch)`, deterministically ordered. `pendingAlreadyACKed` resolves the pending epoch directly, even after `CurrentEpoch` advances. Covered on both sides by `TestStatsRetainsACKForPastEpochAfterBootChange` (real disk spool close/reopen with a new boot id) and `TestRecoveredACKFromPastEpochCompletesPendingStateAfterReboot` (no resend, sequence commits).
3. **Ports / compile assertion:** detection transport, state-store contracts, and their persistence DTOs moved to `internal/usecase/ports`; both `agentstate.DetectionStore` and `fleetclient.Client` now have compile-time interface assertions.
4. **Platform boundary:** Unix-mode capability moved to `internal/platform/fssecurity`; the infra→infra import is gone and credential/state callers share the platform helper.
5. **Error chain:** `DetectionBatchItem.Validate` now wraps the inner `Detection.Validate` error with `%w`.

I also updated the operator guide for engagement fail-closed behavior and historical-epoch ACK recovery.

Validation on the rebased tree:
- `go test -count=1 ./...`
- `go build ./...`
- `go vet ./...`
- Linux/Clang `go test -race -count=1` across fleetagent, ports, detectionship, detectledger, agentstate, fleetclient, spool, httpapi, and synapse-agent
- changed-file `gofmt` and `git diff --check`

Ready for re-review when convenient.

XUanhoa04:
@nghiadaulau Follow-up: the branch is now rebased onto current `main` at `f2ab7fb` (including #658 A6 redaction and #659 fleet docs/coverage), and the new conflict is resolved.

Resolution details:
- Kept #659's more precise server-side fleet configuration descriptions, while retaining this PR's two agent shipper settings.
- Verified #658 compatibility: detections are redacted before P1 persistence, and the shipper signs/hashes the resulting canonical WAL payload unchanged.
- Removed the five now-stale fleet route entries from `api/testdata/openapi-coverage-debt.txt`, since this PR documents those operations in OpenAPI; `go test ./api` confirms the debt guard is clean.

Current head: `bc8e738`.

Re-validation after rebase:
- targeted non-race tests across API, agentspool/privacy, fleetagent, ports, detectionship/detectledger, agentstate/fleetclient/spool, httpapi, and synapse-agent
- the same expanded package set under Linux/Clang `-race`
- `go build ./...`, `go vet ./...`, and `git diff --check`

The PR is ready for re-review again.

nghiadaulau (approved):
Re-reviewed strictly (clean-architecture + security) after the revision, validated locally since CI is off: `go build ./...`, `go vet ./...`, `gofmt` clean; full `go test ./...` green; `go test -race` green across detectionship, agentstate, spool, fleetclient, synapse-agent.

All findings from the prior round are resolved:

- **Engagement-binding (attribution) — CLOSED.** `DetectionPendingBatch.EngagementID` is added + validated, persisted with the batch before the network call, refused on a config mismatch in both `NewService` and `DeliverOnce` (fail-closed `ErrForbidden`), and the signed `BatchMessage` binds the engagement — so reconfiguring `SYNAPSE_DETECTION_ENGAGEMENT_ID` with a pending batch can no longer re-attribute confirmed detections.
- **Reboot-in-ACK-window wedge — FIXED, correct.** `spool.Stats` now emits per-(priority,epoch) `EpochACKs` from the durable ACK highwater map (which survives an incarnation change — proven by `TestStatsRetainsACKForPastEpochAfterBootChange`), and `pendingAlreadyACKed` matches the pending batch's own epoch. No silent-loss false-positive: exact-epoch + `HighestACKed >= Through`, ACK only after a complete 2xx, epochs never reused, one pending batch at a time.
- **Ports relocation + missing compile-assert — FIXED.** Interfaces live in `usecase/ports/detection_delivery.go`; both `var _ ports.DetectionTransport` and `var _ ports.DetectionStateStore` assertions present; no cycle.
- **infra→infra import + `%v` fold — FIXED.** `UnixModeEnforced` moved to `internal/platform/fssecurity`; `batch.go` wraps with `%w`.

The core shipper invariants (WAL ACK strictly after 2xx, membership+sequence persisted pre-send, lost-response replay reuses key/sequence/membership via server SealOnce, exactly-once rotation on 403, credential-derived AgentID, 0600 key never logged) still hold after the ~-133-line rework, and content integrity + duplicate-id fail-closed are intact. You also properly documented the four fleet routes in openapi.yaml and retired their debt entries. Both reviewers: no remaining or new findings.

Approving and merging. Solid follow-through on the review.

Issue #624, opened 2026-08-18
A3 · Batching + transport + idempotent ingest/ACK/gap persistence (carries A0.1/A0.3/A0.4 tails)


nghiadaulau:
> **Phase A3** · child of #594 · **P0** · depends on: A1 (#622), A2 (#623), A0.1/A0.3/A0.4 (#606/#608/#609). Blocks: A4 (#625), A5 (#626), A6 (#627), A7 (#628), C (#638).
> **Carries deferred A0 tails:** A0.1 server-side ingest-verify (#606) · A0.3 persist schema version through storage/transport + two-version acceptance (#608) · A0.4 batching/transport + ingest+ACK + persisted queryable gaps (#609).

## Goal
Build the **agent→control-plane telemetry transport**: batch the spool (#623) into a signed manifest, ship it, and have the control plane ingest it **idempotently**, verify identity **server-side**, persist the **schema version** and **explicit gaps**, and return a **highest-contiguous ACK** so the agent can delete acked batches. This is the live ingest path all the A0 contracts were written against.

## Scope
- **Agent side**: assemble `TelemetryBatchManifest { ProtocolVersion, SchemaVersion, BatchID, AgentID, AgentSessionID, AssetID, StreamID, Sequence, PreviousSequence, EventTimeMin/Max, ObservedCount, KeptCount, SampledOutCount, TruncatedCount, DroppedCount, SamplingPolicyDigest, EventIDs[], EventDigests[], PayloadDigest }` (three **separate** loss buckets `SampledOutCount ≠ TruncatedCount ≠ DroppedCount` per D2/A0.6 #611; `SamplingPolicyDigest` = digest of `{SamplingAlgorithm, SamplingPolicyID, Seed, Version}`; per-event `TruncationReason`/`LossDisposition` live on the first-class `TelemetryLoss`/`TelemetryGap`, not here); the canonical **uncompressed** TRANSPORT-batch commitment (then compress for the wire) — for wire integrity + ACK only; the **permanent evidence** commitment + retention/proof/Merkle semantics are A5's (#626), NOT computed here — signed with the purpose-bound signing key (`Purpose=telemetry-batch`), **KeyID in the envelope**; long-poll/push transport (no control-plane→agent dial); backpressure from 429/5xx into the A2 retry.
- **Server side (ingest endpoint)**:
[...]

Issue #608, opened 2026-08-18
A0.3 · Versioned telemetry event & batch schema (decoupled from agent version)


nghiadaulau:
> Phase **A0** · child of #594 · **P0** (relates to defect **D4**).

## Goal
Version the telemetry **event & batch schema independently of the agent version**, so mixed-version fleets can ship telemetry and later migration stays cheap.

## Scope
- A `SchemaVersion` field on every event envelope **and** every batch — **decoupled from `AgentVersion`** (two separate lifecycles).
- Forward/backward compatibility rules (additive fields, deprecation policy, min/max supported schema per ingest).
- Ingest keys on `schema_version`, never on the agent version.

## Acceptance / exit
- [ ] Two different `schema_version`s ingest correctly from the **same** agent build (schema/agent decoupling proven by test).
- [ ] The ingest side declares the schema range it supports and rejects/upgrades out-of-range envelopes explicitly (no silent misparse).
- [ ] Compatibility rules documented.

## Dependencies
Foundational, alongside **A0.1**. Feeds the canonical envelope work (A1) and every signed payload (A0.2/A0.4/A0.5).

nghiadaulau:
The versioning **contract + ingest gate** landed via #614 (merged): `internal/domain/telemetryschema` (`Current` + `[MinSupported,MaxSupported]` range + `Validate`), `SchemaVersion` on `ports.TelemetryBatch`, and a fail-closed check in `telemetry.Service.Ingest` — decoupled from the agent version, go-arch + security reviews CLEAN.

Keeping this open for two follow-ups: (1) **persist** the schema version in the columnar store (small migration + INSERT column) — lands with **A3** when the agent transport writes versioned batches; (2) the **two-version** acceptance is fully exercised once a **v2** schema is defined (today only v1 exists; the range mechanism is proven — accept in-range, reject below-min/above-max).

pho-veteran:
## Verified progress — A0.3 is partially complete

Re-verified against current `upstream/main` (`67864c3`) and #614. Telemetry schema and ingest tests pass.

Completed:
- schema version is a first-class batch field, independent of agent version;
- ingest declares and validates a supported range fail-closed;
- compatibility rules are documented in the domain package.

Still required by acceptance:
- define/support a second real schema version and prove two versions ingest from the same agent build;
- persist the declared schema version through the live telemetry storage/transport path (explicitly deferred to A3).

[...]

PR #654, opened 2026-08-22
Feat/624 telemetry transport final

VietSory:
Summary

Completes #624 fleet telemetry transport with signed fail-closed ingestion, durable WAL/gap handling, idempotent highest-contiguous ACK semantics, schema v1/v2 compatibility, fleet key management, and documented agent-plane HTTP contracts.

Validation
Exact head: 680b03fbdcfaf920ceb3b664856798d745c9a70d
Agent Package Matrix: PASS — Ubuntu, Debian, Rocky/RHEL, Windows MSI, libc-floor
Frontend test/typecheck/build: PASS
Go build: PASS
Go vet: PASS
Fleet telemetry/domain/usecase/infrastructure tests: PASS
OpenAPI/configuration drift for #624: resolved
HTTP Content-Type compatibility regression: resolved with regression tests
Known repository-level CI debt

Full repository CI still reports failures outside the #624 change-set, including existing Postgres migration/jobqueue tests and two release-evidence lint findings. No #624 telemetry regression was identified in those failures.

Safety / transport invariants
Agent identity and signing-key ownership are server validated
Protocol version is pinned to v1
Schema v1/v2 only
429/5xx retry without WAL loss
Terminal 4xx fails closed
Gap evidence is durable but does not advance transport ACK
Tenant RLS remains enforced

Ready for maintainer review.


nghiadaulau (changes requested):
Reviewed strictly with two independent passes (clean-architecture + security), and validated locally since CI is off: `go build ./...` and `go vet ./...` pass; the non-Postgres test suite and this PR's own telemetry Postgres tests (`-race`, fresh DB, migration 0110 applies) pass. The two full-suite Postgres failures you flagged (`TestMigration0059`, the jobqueue cross-tenant stats test) reproduce identically on `main` — confirmed pre-existing, not this PR's fault. Thanks for calling them out honestly.

This is a strong PR. Both reviews came back clean on substance:

- **No regression of any A3 (#651) guarantee.** Cross-agent stream isolation (agent_id stays in every identity key/PK), the optimistic-concurrency version CAS, tenant RLS FORCE + explicit `tenant_id` predicates, and idempotency are all preserved — and in several places strengthened (server-derived StreamID/session, host binding, the batch-commit equivocation guard).
- **The reintroduced `telemetry_transport_gaps` table is NOT the phantom-gap problem A3 deliberately avoided** — it's a transactionally-reconciled materialization of the ACK-derived gaps (opened and `resolved_at`-closed inside the same `SaveStreamState` transaction, read path filters `resolved_at IS NULL`), and `telemetry_agent_gaps` is a correctly-separated distinct concept (durable endpoint-origin loss that must survive a later sequence fill). Good design.
- **Security: clean, no blocker.** Signed gap reports use a distinct `synapse-telemetry-gap:v1` domain-separation tag (no cross-protocol signature replay), are verified fail-closed and agent/tenant-scoped, are idempotent on GapID, and never advance/regress the transport ACK. gzip ingest bounds both the compressed (8 MiB) and decoded (32 MiB) sizes (gzip-bomb safe), rejects unknown encodings, and decodes strictly. Agent private keys are 0600 + never logged/sent.

### One blocker before merge

**gofmt.** The tree is not `gofmt`-clean — 20 files. `make build vet` doesn't catch this; please run `gofmt -w .` (or `make lint`) and push. The affected files:

```
cmd/synapse-agent/telemetry_batch_journal.go
cmd/synapse-agent/telemetry_gap_shipper_test.go
cmd/synapse-agent/telemetry_shipper.go
internal/adapter/httpapi/fleet_telemetry_gap_handler_test.go
internal/adapter/httpapi/fleet_telemetry_handler_test.go
internal/domain/telemetry/envelope.go
internal/domain/telemetry/envelope_test.go
internal/infrastructure/persistence/memory/telemetry_batch_commit.go
internal/infrastructure/persistence/memory/telemetry_transport_store.go
internal/infrastructure/persistence/postgres/telemetry_agent_gap.go
internal/infrastructure/persistence/postgres/telemetry_agent_gap_query.go
internal/infrastructure/persistence/postgres/telemetry_batch_commit.go
internal/infrastructure/persistence/postgres/telemetry_transport_repo.go
internal/usecase/fleet/telemetryingest/gap_test.go
internal/usecase/fleet/telemetryingest/service.go
internal/usecase/fleet/telemetryingest/service_test.go
internal/usecase/ports/telemetry.go
internal/usecase/ports/telemetry_gap_reader.go
internal/usecase/ports/telemetry_gap_reader_test.go
internal/usecase/ports/telemetry_spool.go
```

### Two minor idiom nits (optional — not merge-blocking)

1. `internal/domain/fleetagent/sampling_policy.go` — a validator returns a bare `fmt.Errorf(...)` with no `%w`/`shared.ErrValidation` sentinel, unlike its sibling domain validators; please wrap it so the edge maps it to a 400.
2. `internal/usecase/fleet/telemetryingest/service.go` (`NewService`) — `bindings` is obtained by a runtime type-assertion on `transport` rather than an explicit injected dependency. It's fail-closed-validated in the constructor so it's safe, but an explicit `ports.TelemetryAssetBindingStore` constructor parameter would remove the hidden requirement that both interfaces be the same concrete type.

Push the gofmt fix and I'll re-validate locally and merge (the two nits can come along or be a follow-up — your call). Nice work on the hardening.

nghiadaulau (changes requested):
Re-reviewed after the upstream-merge + fixup commits. Validated locally (CI is off). Two hard blockers — the branch is currently in a worse state than the previous round, so I can't merge it yet.

### BLOCKER 1 — the tree does not compile (A6 regression from the merge)

`go build ./...` fails:

```
internal/domain/privacy/scrub.go:101:6: out.RedactionPolicyDigest undefined
    (type telemetry.TelemetryEnvelope has no field or method RedactionPolicyDigest)
```

The "merge upstream main" commit clobbered a field that A6 (#658, already on main) added: `TelemetryEnvelope.RedactionPolicyDigest`. This PR rewrites `internal/domain/telemetry/envelope.go` (23+/33-), and the conflict resolution dropped that field from the struct — but A6's `internal/domain/privacy/scrub.go` (also on main) still sets `out.RedactionPolicyDigest`, so the build breaks. Note `QualityRedacted` in `quality.go` survived, so it's specifically the envelope field that was lost.

This is not just a compile error — it silently regresses A6: the redaction-policy digest is A6's guarantee that "the redaction that scrubbed this telemetry travels with the data." Please restore `RedactionPolicyDigest string` on `TelemetryEnvelope` (additively, alongside your A3/#624 fields) so both A6 and this PR compose, and confirm `go build ./... && go test ./internal/domain/privacy/... ./internal/adapter/agentspool/...` is green.

### BLOCKER 2 — still not gofmt-clean (same as the previous round)

`gofmt -l` still reports 20 files (unchanged from my last review), e.g. `internal/usecase/fleet/telemetryingest/service.go`, `internal/domain/telemetry/envelope.go`, `internal/infrastructure/persistence/postgres/telemetry_transport_repo.go`, … Please run `gofmt -w .` (or `make lint`) and push.

---

The substance from the prior round still looks strong (server-authoritative session/stream/asset binding, batch-commit equivocation guard, signed gap reports, RLS on the new tables, gzip caps) — this is purely: (1) fix the merge that dropped A6's envelope field so it builds and A6 stays intact, (2) gofmt. Once `go build ./...` and `go test ./...` are green locally and `gofmt -l` is empty, I'll re-review the delta and merge. Also please re-check that your `internal/domain/telemetry/envelope.go` rewrite kept every field A5/A6 added (RedactionPolicyDigest is the one that broke; scan for others while you're there).

nghiadaulau (changes requested):
Re-reviewed after the rebuild fixes (dual go-arch + security audit, plus local verification — CI is off, validated locally). **The build/gofmt blockers from the last round are resolved and the substance is strong**, but adversarial security review surfaced one **HIGH** that has to be addressed before merge, because it defeats this PR's own attribution guarantee.

## Local verification — all green
- `go build ./...` ✅, `go vet` ✅ (the dropped `TelemetryEnvelope.RedactionPolicyDigest` field is restored — A6 composes again).
- Full fleet + telemetry + privacy + `test/e2e` suites pass; migration **0110** applies cleanly to a fresh Postgres and all transport/fleet PG tests pass.
- Schema **v1/v2** compatibility is a genuine, well-shaped addition (satisfies the A0.3 #608 two-version acceptance): v1 stays readable, v2 makes the concrete delivery-incarnation identity mandatory.
- Signing / fail-closed-and-audited rejection / optimistic-CAS + forward-gap bound / gzip+body caps / agent-gap-vs-transport-gap separation / RLS on all four new tables — all verified solid.

## 🔴 HIGH — required before merge: cross-agent telemetry asset-binding takeover
The PR's stated guarantee is "the server binds agent id + asset from the authenticated credential, not from batch fields." The **agent-id** half is solid. The **asset** half is not.

The canonical host asset key is `hostKey(facts)` = `machine-id/<machine_id>` (else `hostname/<hostname>`), derived **entirely from agent-supplied host-inventory facts** — `internal/usecase/fleet/hostinventory/service.go:83,127-135` — and `Sync` applies no check that the reported host belongs to the calling agent. The 0110 trigger `synapse_sync_telemetry_asset_binding` then does an **unconditional** cross-agent displacement (`migrations/0110_telemetry_transport_tail.sql:39-49`): `DELETE ... WHERE asset_id = NEW.id AND agent_id <> reporting_agent`, then bind `reporting_agent → asset`.

Attack (single tenant, two enrolled agents — the normal EDR threat model where one host is compromised):
1. Agent B POSTs host inventory carrying victim A's `machine_id`/`hostname`.
2. `Sync` upserts A's asset `X` and stamps `reporting_agent_id = B` (server-authored from B's auth — authentic, but the **asset key itself is spoofable**).
3. Trigger silently rebinds `B → X` and deletes A's binding.
4. B's signed, authenticated telemetry now resolves `ResolveTelemetryAsset(B) = X` → lands attributed to victim host `X`; A's telemetry now fails closed (`asset_mismatch`). A's continuous inventory sweep re-reports the same key, so the binding **flaps** — recurring windows to inject telemetry under the victim's asset and to blind the victim.

For a chain-of-custody / evidence product this is cross-host attribution spoofing (evidence framing) plus victim telemetry denial, by any one compromised enrolled agent, with no monotonic/authorization guard and no audit. The underlying host-key spoofability predates this PR, but #654 is what promotes it into telemetry attribution, so it needs handling here.

**Fix (minimum to merge):** make the cross-agent rebind non-silent and guarded — refuse silent last-writer-wins displacement of an existing `asset_id → different agent` binding, and emit an audit + alert whenever an asset rebinds to a different agent. **Preferred:** anchor the agent↔asset binding to something the reporting agent cannot forge for a *different* host (establish/confirm it at enrollment or via a server-issued host credential) rather than trusting agent-declared `machine_id`/`hostname`. The deeper enrollment-anchored fix can be a tracked follow-up, but the silent-takeover path must be closed in this PR.

## 🟡 Should-fix
- **Doc-comment regression (`internal/domain/telemetry/envelope.go`):** the PR strips the safety-bearing docs — the 3-timestamp meaning (OccurredAt = kernel clock / ObservedAt = userspace decode / ReceivedAt = ingest, the D4 rationale) and, most importantly, `StampReceived`'s "the ONLY sanctioned way to set the control-plane timestamp" contract. The ordering check still runs, but dropping the invariant doc invites a future direct-assignment regression. Please restore at least the `StampReceived` sole-setter contract + the 3-timestamp semantics (and ideally the other exported-symbol docs removed for churn in `ports/telemetry.go` / `ports/telemetry_transport.go`).

## 🟢 Minor / nits
- **`gofmt`**: `test/e2e/harness.go` is still not gofmt-clean (one struct-field alignment at ~line 199). `gofmt -w test/e2e/harness.go` and push.
- **LOW — unbounded agent-gap rows:** `IngestGap` accepts unlimited distinct signed `GapID`s per agent with no per-(agent,stream,epoch) cap (`internal/usecase/fleet/telemetryingest/gap.go`). Comparable to normal event-storage growth so not the A0.4 hazard, but a per-(agent,stream,epoch) cap + reject/audit (mirroring `maxForwardGap`) would honor the "no unbounded hostile state" goal.
- **LOW — store-twin divergence:** the memory `BindTelemetryAsset` is stale-guarded per (tenant,agent) but does **no** cross-agent displacement, while the Postgres trigger enforces single-agent-per-asset. Beyond the parity gap, this is why unit tests (memory-backed) don't exercise the takeover above — please model the same displacement policy in memory and add a test asserting the intended cross-agent rebind behavior.
- **Nit:** `internal/infrastructure/spool/gaps.go` — remove the leftover `var _ = time.Time{}` import-guard.
- **Nit (go-arch):** `detectionEngagement`/`detectionShipInterval` moved to package-level mutable vars bound via `flag`/`init`; prefer keeping them on `config`.

Everything except the HIGH is quick. Fix the silent cross-agent takeover (minimum: guarded + audited rebind) + gofmt + restore the `StampReceived`/timestamp docs, push, and I'll re-review and merge. Thanks for the thorough transport work — the signing/fail-closed/idempotency core is genuinely well done.

nghiadaulau (changes requested):
Round 4 — thanks, the hardest part is done 🎯

## ✅ HIGH (cross-agent asset-binding takeover) — CLOSED, verified on every path
I ran a dual review (go-arch + security) plus a real test on a clean Postgres. Confirmed:
- **`reporting_agent_id` is server-authoritative** — taken from the agent id authenticated via the client-cert CN (`fleet_handler.go:112-141` → `hostInventory` :666-669 → `hostinventory/service.go:200-214`), **never** from the body/`inv.Facts`. A malicious agent cannot set `reporting_agent_id` to a victim. This is the crux, and it's correct.
- **Trigger is fail-closed** on both INSERT and UPDATE OF attributes (`0113:64-66`); the `EXISTS ... agent_id <> reporting_agent → RAISE 23505` branch (`:42-51`) blocks a cross-agent rebind; the `ON CONFLICT (tenant,agent)` branch cannot be used to move a binding to another agent; `UNIQUE(tenant_id, asset_id)` is the race backstop.
- **The memory twin now has the guard** (`memory/telemetry_transport_store.go:388-392` → `ErrConflict`), matching the trigger's semantics — last round the twin lacked it so unit tests couldn't catch the issue; now `telemetry_asset_binding_guard_test.go` covers it.
- **Use-case guard** writes an audit entry + `security.alert` (high) before returning `ErrConflict` (`hostinventory/service.go:131-179`).
- Real test: `TestTelemetryAssetBindingTriggerRejectsCrossAgentTakeover` PASSES on a fresh DB (migrations 0097→0113 apply cleanly), and the memory/hostinventory/telemetryingest unit tests are green. RLS on all four new tables; composite FKs `(tenant,agent)`/`(tenant,asset)` OK.

Three consistent defense layers — the takeover is closed. 👏

## ⛔ Blocker (small but required)
1. **gofmt-dirty:** `internal/infrastructure/persistence/memory/telemetry_asset_binding_guard_test.go` (struct-field alignment on the `TelemetryAssetBinding` literals). Just run `gofmt -w` on that file. This is a hard merge gate.

## 🔧 Should-fix (please do in this PR)
2. **(MAJOR) `23505` is not mapped to `shared.ErrConflict`** — `internal/infrastructure/persistence/postgres/asset_repo.go` `UpsertAsset` returns the raw pgx error. A takeover is usually caught earlier by `guardAssetBinding` (→409), but the **DB-level race backstop** (the trigger raising 23505 on `uq_telemetry_asset_bindings_asset`) currently surfaces as **HTTP 500 instead of 409**. This PR introduces the trigger that makes that path reachable, so it should map it: detect `pgconn.PgError.Code == "23505"` on that constraint → wrap as `shared.ErrConflict`.

## 📋 Fast-follow (does not block this PR — can be a separate PR)
3. **(LOW) `telemetry_agent_gaps` grows unbounded** — `GapID` is agent-supplied; ingest is idempotent/coalescing per `GapID`, but there is **no cap** on the number of distinct gap rows per `(tenant, agent, stream, epoch)`, and no rate limit at the handler. An enrolled-but-compromised agent could emit unbounded `GapID`s → storage-exhaustion DoS. Add a cardinality cap per `(tenant,agent,stream,epoch)` or a quota + retention, mirroring the ack-ledger forward-gap bound. (This LOW is carried over from the previous round.)
4. **(LOW, test-only) PG repo `BindTelemetryAsset` (`:493-503`)** maps a cross-agent conflict to a generic error instead of `ErrConflict` like the memory twin. There is **no production caller** today (prod uses the trigger; this method is test/e2e only), so impact = 0, but if it's ever wired to a live path, `errors.Is(err, ErrConflict)` would misclassify. Either add an `EXISTS` pre-check + `ErrConflict` mapping, or document it as test-only and assert it's never wired.
5. **(INFO) defense-in-depth for orphaned host assets:** when an agent re-points to a new asset (reimage / machine-id change), the old asset loses its binding row → the trigger guard no longer fires for it, leaving only the use-case layer reading `attributes.reporting_agent_id`. Not exploitable today (host assets always carry `reporting_agent_id`), but a comment/test asserting host assets never lose `reporting_agent_id` would be good.

## Verdict: **Request changes**
The substance is excellent — the HIGH is firmly closed, three defense layers, tests green. The only things left are **(1) gofmt** (required) and I'd really like you to also handle **(2) map 23505→ErrConflict** (so the race backstop returns a clean 409 instead of 500). Once (1)+(2) land I'll merge right away; (3) can be a fast-follow. Thanks for sticking with it through 4 rounds 🙏

_(CI is off by request — I verified locally: build/vet/gofmt (except the file above)/unit + PG integration on a clean DB.)_

nghiadaulau (approved):
Round-5 ✅ — both blocking items from Round-4 are landed and verified:
1. gofmt clean (`gofmt -l internal cmd` empty; commit 9439c7c3).
2. 23505 → ErrConflict wired at the tenant tx-wrapper (`normalizePersistenceError`, tenant.go:59/82/94), so the UNIQUE(tenant,asset) race backstop returns a clean 409 on the production host-inventory bind path (not just the test-only BindTelemetryAsset). Covered by tenant_error_test.go + telemetry_asset_binding_guard_test.go; postgres unit tests + affected-package build pass locally.

The HIGH (cross-agent asset-binding takeover) stays closed. Fast-follow (3) telemetry_agent_gaps cardinality cap can be a separate PR. Approving + merging A3.

PR #657, opened 2026-08-23
fix(ai-triage): admit release approvers from an operator-owned allowlist

### LINKED ISSUE #656: fix(ai-triage): release approvals are proven human by a denylist the code documents as fail-open
## Problem

The release ledger is the strongest human-governance artifact the AI-triage subsystem produces.
`synapse-fptriage-release` appends an immutable, hash-chained decision recording that one PM and one
Security reviewer approved a specific evaluation comparison, and the guide states the guarantee plainly
(`docs/guide/ai-triage-evaluation.md:339-340`):

> One PM and one Security reviewer must independently add approvals in canonical order. They must be
> distinct human identities; model names and reserved machine principals are rejected.

What enforces "human" is `validateReleaseApprovals` (`internal/usecase/sca/fptriage_release.go:436-460`).
Most of what it checks is solid: the approval binds the exact review digest, carries a rationale, uses a
UTC timestamp, fills exactly the PM and Security roles, and the two reviewers must be distinct. For
*human-ness* it has exactly one test — `aitriagereview.IsMachineOrModelActor(reviewer, models...)` at
:449 — applied to the `reviewer` string the manifest itself supplies. `AIEvaluationReleaseManifest` is
documented as "operator-owned approval input" (`:40-41`) and `Reviewer` is a free-form field inside it.

That check resolves to `machineActor` (`internal/domain/aitriagereview/review.go:238-246`), a
seven-prefix denylist whose own doc comment says what it is (`:225-237`):

> This is the SECOND line: RBAC is the first, and machine roles are granted nothing on the human REST
> API (user.rolePermissions), so a machine cannot reach Decide over HTTP at all. **This exists for the
> in-process callers that do not pass through that gate.**
>
> A prefix DENYLIST fails open on the next identity scheme somebody invents. The durable fix is to invert
> it -- require an actor to be a known human principal -- but that needs a definition of human identity
> that this package does not own, so it is left as a follow-up rather than guessed at here.

The release path is exactly the in-process caller that comment names. `synapse-fptriage-release` is an
offline CLI over JSON files: grepping the release types under `internal/adapter/` and `cmd/synapse-api/`
returns nothing, so no HTTP handler and no RBAC stands in front of it. On this path the denylist is not
the second line, it is the only one — and any identity family that does not begin with one of
`agent: llm: mcp: system: machine: bot: service:` reads as human and is recorded in the ledger as human
approval.

## What changed since that comment was written

The deferral was explicit and, at the time, correct: the package did not own a definition of human
identity. The project owns one now — #648 supplied it.

`synapse-fptriage-blind` authenticates its reviewer considerably more strictly than the release boundary
authenticates a PM or Security approver:

| | `synapse-fptriage-blind` | `synapse-fptriage-release` |
| --- | --- | --- |
| Where the identity comes from | `--reviewer`, "never read from submission JSON" (`cmd/synapse-fptriage-blind/main.go:25`) | the `reviewer` field of the manifest being validated |
| Positive membership test | `--human-reviewers`, "private operator-owned allowlist file", required (`main.go:26`, `:195`) | none |
| Machine-prefix denylist | yes, behind the allowlist (`internal/usecase/sca/fptriage_blinded.go:315-324`) | yes, as the only check |
| Authentication of the artifact | HMAC-SHA256 receipt, key at least 32 bytes (`fptriage_blinded.go:25`, `:30-38`) | none |

The asymmetry runs the wrong way. The blinded packet measures precision and recall; the release ledger
authorizes the promotion. The weaker artifact is the authenticated one.

## Scope — what I am not claiming

- **This is not privilege escalation.** The manifest is operator-owned and the command is offline;
  anyone able to author it already stands inside the trust boundary. The claim is about *evidence
  integrity*: a ledger entry that exists to prove two humans signed off currently preserves a
  self-asserted string that passed a prefix test. #648 applied a stricter standard to a lesser artifact,
  which is the argument that it belongs here too.
- **The feedback path is fine as it stands.** `validateFeedbackReview`
  (`internal/usecase/sca/fptriage_feedback.go:324`) and the approval check at `:366` revalidate a durable
  snapshot whose `DecidedBy` did reach the system through the HTTP RBAC gate, so there the denylist
  genuinely is the second line the comment describes. I am not proposing to change it.
- **I am not proposing to unify the duplicated denylists.** `internal/usecase/offensivepolicy/enforce.go:31-37`
  documents why it mirrors `machineActor` rather than importing it — two policies that should be free to
  diverge — and I think that reasoning holds.
- **The denylist stays.** This adds a positive test in front of it; it does not replace it, and it does
  not touch `machineActor` itself.

## Proposed work

1. Add an operator-owned human-approver allowlist at the release boundary, in the shape #648 already
   established: `--human-approvers <file>`, one identity per line, operator-private, required for any
   decision that records approvals.
2. Thread it into `validateReleaseApprovals` so an approval must match an allowlisted identity **and**
   clear the existing denylist and model-identity checks. A missing or empty allowlist fails closed
   rather than degrading to current behaviour.
3. Document it in the approval walkthrough of `docs/guide/ai-triage-evaluation.md`, beside the sentence
   that promises distinct human identities.
4. Tests: an approver absent from the allowlist is rejected; an allowlisted approver still fails the
   denylist and model-identity checks; a missing allowlist fails closed; every existing ledger behaviour
   is unchanged.

Roughly the size of #579 — one usecase function, one CLI, the guide, and tests. No ledger schema change,
no change to any promotion criterion, and no change to a gate decision.

## Acceptance criteria

- [ ] A release decision that records approvals requires an operator-supplied human-approver allowlist,
      and fails closed without one.
- [ ] An approver identity outside the allowlist is rejected even when it clears the machine-prefix
      denylist.
- [ ] Digest binding, distinct-reviewer, model-identity, and role validation are unchanged and still
      covered by tests.
- [ ] The guide documents where the allowlist comes from and that it is operator-private.

## Related

Follows the trusted-boundary pattern introduced by #648 under #583. It touches the same evidence chain as
#593's blind-review acceptance gate, but at the release end rather than the pilot end, and does not
overlap the benchmark or pilot work tracked there.

thx2an:
Closes #656.

## What this changes

Recording an AI-triage release decision now requires `--human-approvers`, a private operator-owned
allowlist that every PM and Security approver identity must appear in.

Before this, `validateReleaseApprovals` established a great deal about an approval — that it binds the
exact review digest, fills the PM and Security roles in canonical order, carries a rationale and a UTC
timestamp, and that the two reviewers are distinct — but its only test of *human-ness* was
`aitriagereview.IsMachineOrModelActor` applied to the `reviewer` string the manifest itself supplies.
`AIEvaluationReleaseManifest` is documented as operator-owned approval input, so the ledger's central
claim rested on the artifact's own assertion, checked against a denylist of the identity families this
codebase already mints. `machineActor`'s comment says what that costs:

> A prefix DENYLIST fails open on the next identity scheme somebody invents. The durable fix is to invert
> it -- require an actor to be a known human principal -- but that needs a definition of human identity
> that this package does not own, so it is left as a follow-up rather than guessed at here.

The same comment notes the denylist is the *second* line behind RBAC. `synapse-fptriage-release` is an
offline CLI over JSON files with no HTTP handler, so on this path there is no first line — it is exactly
the in-process caller the comment says it exists for and cannot cover alone.

I did not have to guess at a definition of human identity, because #648 already supplied one.
`synapse-fptriage-blind` takes the reviewer from outside the artifact being validated (`--reviewer`,
"never read from submission JSON"), requires a private operator-owned `--human-reviewers` allowlist, and
seals the result with an HMAC receipt. This applies that same pattern to the artifact that authorizes the
promotion rather than the one that measures a pilot.

## Design decisions worth reviewing

**Admission only, never re-validation.** `validateReleaseApproverAllowlist` is called from
`ApplyAIEvaluationRelease` and deliberately not from `Validate`. `LoadAIEvaluationReleaseLedger` validates
every historical decision on every load, so requiring allowlist membership there would mean an approver
leaving the company retroactively invalidates the decisions they signed and stops the ledger loading at
all. A decision keeps the approvers it was admitted with;
`TestAIEvaluationReleaseLedgerValidatesWithoutTheApproverAllowlist` locks that down, and also asserts the
*next* decision still has to clear the allowlist in force now.

**The denylist keeps its say.** The allowlist is a positive test added in front of the existing checks,
not a replacement. Each allowlist entry is re-checked against the denylist so allowlisting a model
identity cannot launder it into an approval — covered by a test.

**Exact match.** Approver *distinctness* is compared case-insensitively; allowlist membership is exact,
so one entry admits one spelling of one identity. Untrimmed entries are rejected. This mirrors
`validateBlindFPTriageReviewer` rather than inventing a second matching rule for the same subsystem.

**`--print-review-digest` does not require the allowlist.** It runs before anyone has approved anything,
so demanding the file there would gate printing a digest on a list with nothing yet to check. The digest
tests pass `""` for the flag, which is the assertion.

**The output guard covers the new input.** `--output` already refuses to overwrite an input artifact; the
allowlist path joins that list.

## What this does not change

- No ledger, manifest, comparison, or report schema version changes; existing ledgers load unchanged.
- No promotion criterion, threshold, policy, or gate decision is touched.
- `machineActor` itself is unchanged, and so are the other consumers of the denylist. The feedback path
  (`validateFeedbackReview`, `validateFeedbackApproval`) revalidates a snapshot whose `DecidedBy` did
  reach the system through the HTTP RBAC gate, so there the denylist genuinely is the second line and I
  left it alone. I am also not proposing to merge the duplicated prefix lists —
  `offensivepolicy/enforce.go` documents why it mirrors rather than imports, and that reasoning holds.
- This is not a fix for privilege escalation. The manifest is operator-owned and the command is offline;
  the change is about what the ledger can be said to prove.

## Tests

- `TestAIEvaluationReleaseRequiresAllowlistedHumanApprovers` — no allowlist, empty allowlist, an approver
  missing from it, a case-differing entry, an untrimmed entry, and a machine-principal entry are each
  rejected; allowlisting a model identity does not admit it; allowlisted approvers still succeed.
- `TestAIEvaluationReleaseLedgerValidatesWithoutTheApproverAllowlist` — signed history validates and
  re-loads with no allowlist in scope, while the next decision still needs one.
- `TestRunRequiresAPrivateHumanApproverAllowlistToRecordADecision` — an absent allowlist and a
  group-readable one are both refused at the CLI, and no ledger file is left behind.

I checked the new tests are falsifiable rather than merely green: stubbing
`validateReleaseApproverAllowlist` to return `nil` fails all six subtests plus the history test, and
stubbing the CLI loader to default to a built-in list fails both CLI subtests.

## Verification

- `gofmt -l .` clean; `go build ./...`; `go vet` over `internal/usecase/sca/...`,
  `cmd/synapse-fptriage-*`, `internal/domain/aitriagereview/...`.
- `go test -count=1` green for `internal/usecase/sca`, `internal/usecase/fptriage`,
  `internal/domain/aitriagereview`, and all five `cmd/synapse-fptriage-*` commands.
- `make ai-triage-verify` passes.
- `golangci-lint v2.12.2` (the CI pin) over the changed packages: **0 issues**.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -45,6 +45,14 @@ and this project aims to adhere to [Semantic Versioning](https://semver.org/spec
   precondition reports pair counts in dedicated `*_count` failure fields, so the basis-point fields
   of a promotion failure always mean a rate.
 
+- **Operator-owned human approver allowlist on AI-triage releases.** Recording a promotion or rollback
+  now requires `--human-approvers`, a private operator-owned allowlist an approver identity must appear
+  in. The release manifest names its own PM and Security approvers, so previously the only test of their
+  humanness was a reserved machine-prefix denylist, which by construction cannot recognise an identity
+  scheme it has never seen; the allowlist admits identities from outside the artifact being validated.
+  It is enforced when a decision is admitted, not when stored history is re-validated, so an approver
+  leaving the allowlist never invalidates a ledger they already signed.
+
 - **AI-triage escape rate measured against the exemptible population.** Evaluation reports now carry
   `exemptible_true_positives` and `exemptible_escape_rate` alongside the corpus-wide rate, and the
   promotion boundary reads the exemptible denominator. A true positive a human-review floor holds
--- a/cmd/synapse-fptriage-release/main.go
+++ b/cmd/synapse-fptriage-release/main.go
@@ -22,16 +22,17 @@ func main() {
 	baselinePath := flag.String("baseline", "", "baseline synapse-ai-triage-evaluation-v4 report (promotion only)")
 	candidatePath := flag.String("candidate", "", "candidate synapse-ai-triage-evaluation-v4 report (promotion only)")
 	outputPath := flag.String("output", "", "new path for the append-only release ledger")
+	humanApproversPath := flag.String("human-approvers", "", "private operator-owned allowlist file, one human approver identity per line")
 	printDigest := flag.Bool("print-review-digest", false, "print the digest PM/Security must approve without writing a ledger")
 	flag.Parse()
 
-	if err := run(*manifestPath, *ledgerPath, *comparisonPath, *baselinePath, *candidatePath, *outputPath, *printDigest, os.Stdout); err != nil {
+	if err := run(*manifestPath, *ledgerPath, *comparisonPath, *baselinePath, *candidatePath, *outputPath, *humanApproversPath, *printDigest, os.Stdout); err != nil {
 		fmt.Fprintf(os.Stderr, "synapse-fptriage-release: %v\n", err)
 		os.Exit(1)
 	}
 }
 
-func run(manifestPath, ledgerPath, comparisonPath, baselinePath, candidatePath, outputPath string, printDigest bool, stdout io.Writer) error {
+func run(manifestPath, ledgerPath, comparisonPath, baselinePath, candidatePath, outputPath, humanApproversPath string, printDigest bool, stdout io.Writer) error {
 	manifestPath, ledgerPath, comparisonPath = strings.TrimSpace(manifestPath), strings.TrimSpace(ledgerPath), strings.TrimSpace(comparisonPath)
 	baselinePath, candidatePath, outputPath = strings.TrimSpace(baselinePath), strings.TrimSpace(candidatePath), strings.TrimSpace(outputPath)
 	if manifestPath == "" {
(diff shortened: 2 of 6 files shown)


nghiadaulau (approved):
Reviewed strictly (clean-architecture + security passes) and validated locally since CI is off: `go build ./...`, `go vet`, `gofmt` all clean; `./internal/usecase/sca/...` and `./cmd/synapse-fptriage-release/...` tests green.

This is a real separation-of-duties hardening of the model-promotion release gate. Before, `ApplyAIEvaluationRelease` trusted the manifest's self-declared approvers with only a machine-prefix *denylist* — "these two are human" was a claim the artifact made about itself. Adding a positive, operator-owned allowlist closes that gap, and the design is careful:

- **Fail-closed** at both layers — an empty/missing/comment-only allowlist is a refusal, not a skip, and the check runs *inside* `ApplyAIEvaluationRelease` so even a direct programmatic caller passing nil is refused. The only ledger-writing path goes through it; `--print-review-digest` legitimately skips it (no decision recorded).
- **No machine laundering** — each allowlist entry is re-checked against `IsMachineOrModelActor` with the same models set, exact-match (no case-fold/whitespace bypass), and the existing distinct-humans / PM-then-Security / digest-binding checks are untouched.
- **Correct immutability call** — enforced at admission, not in `Validate`/load, so a later staffing change can't retroactively invalidate or block loading a signed ledger, while the *next* decision still must clear the current allowlist.
- **Private-file guard** byte-for-byte mirrors the established `synapse-fptriage-blind` pattern (0600, no symlink/special, Windows refused), and the allowlist can't be the `--output`.

Minor, non-blocking (no change needed): the `!IsMachineOrModelActor(allowed, …)` clause in `releaseApproverIsAllowed` is effectively unreachable belt-and-suspenders (the reviewer is already denylist-checked earlier) — fine to keep; and `requirePrivateFile` is now duplicated verbatim in two `main` packages — acceptable per composition-root, worth an `internal/platform` helper only if a third copy appears.

Approving and merging. Nice, well-tested hardening.

PR #662, opened 2026-08-23
feat: harden horizontal execution for issues 589 and 591

### LINKED ISSUE #589: Scale: remove API single-instance-per-role cap to enable horizontal scale / HA

### LINKED ISSUE #591: Security: enforce sandbox-on in production and prove startup fail-closed

pho-veteran:
## Summary

This PR completes the remaining production execution-isolation and measured horizontal-capacity work for #589 and #591.

It keeps API/web and gated migrations on EKS, makes production API replicas dispatch-only, and moves untrusted tool execution to native non-root workers in a private EC2 Auto Scaling Group. Those workers share PostgreSQL through a fenced durable queue and invoke the exact hardened Bubblewrap runner behind a strict startup conformance gate. A root-owned, typed egress broker grants only signed, scope-bound network rules without giving the worker broad host privileges.

The branch was rebased onto current `upstream/main` (`4718df1`); the new queue migration is `0110_jobs_claim_fence.sql` because upstream now owns migration 0109.

### Dispatch-only production API

- Adds explicit `dispatch-only`, `worker`, and development/test `in-process` execution roles.
- Production `synapse-api` dispatches durable work and cannot construct or invoke the local untrusted-tool runner.
- Production `synapse-worker` requires PostgreSQL, worker execution role, Linux sandboxing, and a passing exact-runner posture before it claims jobs.
- Moves concrete shared wiring into `internal/composition` so `internal/platform` does not depend on infrastructure.
- Adds an architecture regression test for that inward dependency rule.

### Safe multi-worker execution

- Adds a monotonic queue `claim_fence` and fence-aware heartbeat, completion, retry, failure, and dead-letter operations.
- PostgreSQL claiming retains `FOR UPDATE SKIP LOCKED`; stale workers receive a typed stale-lease result and cannot publish terminal queue state.
- Worker handlers use claim-scoped contexts and an authoritative lease deadline.
- Lease loss/expiry and parent shutdown cancel active handlers, including heartbeat I/O.
- Parent-delivery cancellation is retryable, while configured child scan timeouts remain terminal.
- Async SCA reserves the durable job ID as its evidence identity, making commit-ambiguous redelivery idempotent even when AI-triage telemetry changes; finding-linked evidence remains byte-exact.

### Exact hardened execution boundary

- Adds `synapse-sandbox-check`, which constructs and runs the same production `sandbox.Runner` used by tools rather than duplicating Bubblewrap in deployment scripts.
- Enforces curated read-only filesystem bindings, an explicit writable work directory, namespace isolation, default-deny network, inherited seccomp, drop-all capabilities, executable SHA-256 verification, cgroup-v2 memory/PID limits, timeout/process-tree termination, bounded output, and shared secret redaction.
- Fails closed when Bubblewrap or any required strict control is unavailable; there is no unsandboxed production fallback.
- Adds strict full conformance and recovery probes for the production worker image.

### Signed least-privilege egress

- Adds a root-owned Unix-socket broker with a typed protocol—never arbitrary command text or argv.
- Validates worker peer credentials and exact Bubblewrap descendant/executable identity.
- Uses Ed25519 grants binding tenant, execution kind/ID, run ID, slot, Bubblewrap PID, canonical rules, issue/expiry time, and grant ID.
- Persists replay consumption before privileged mutation; exact grant replay is rejected and a failed setup burns the grant.
- Fences concurrent setup/cleanup for the same run ID.
- Keeps the worker itself non-root with an empty effective capability set; the broker receives only narrowly bounded host capabilities and no unrestricted sudo or broad `SYS_ADMIN`.

### Production deployment

- Adds a private AL2023 native-worker launch template and Auto Scaling Group with encrypted storage, IMDSv2, no public IP, no inbound application port, and SSM administration instead of SSH.
- Adds separate runtime/migration identities and native-worker IAM wiring.
- Packages worker, broker, runtime-env, and strict sandbox-check systemd units.
- Uses systemd delegation for per-run cgroup-v2 enforcement.
- Adds Helm external-worker posture so production EKS runs the API/web control plane without an in-cluster execution worker.
- Documents the topology in ADR 0008 and the deployment, configuration, security, architecture, and capacity guides.

## Measured acceptance

The AWS staging environment used two API replicas behind the EKS load balancer, one shared RDS PostgreSQL database, and four private native EC2 workers. AI was disabled during capacity measurements.

### #589 operating envelope

API admission through the load balancer:

| Concurrency / target | Requests | Success | p50 | p95 | p99 |
| --- | ---: | ---: | ---: | ---: | ---: |
| 32 / 500 req/s | 5,000 | 5,000 | 2 ms | 2 ms | 9 ms |
| 64 / 1,000 req/s | 10,000 | 10,000 | 1 ms | 2 ms | 55 ms |
| 128 / 2,000 req/s | 20,000 | 20,000 | 1 ms | 3 ms | 98 ms |
| 256 / 5,000 req/s | 50,000 | 50,000 | 1 ms | 13 ms | 155 ms |
| 256 / 5,000 req/s repeat | 50,000 | 50,000 | 1 ms | 20 ms | 215 ms |

API failover during sustained load completed 60,000/60,000 requests at about 999.883 req/s with p50/p95/p99 of 1/4/11 ms; the replacement became ready in 24.184 seconds and the deployment returned to 2/2 ready.

Four-worker scan ladder:

| Rung | Accepted / successful | Throughput | Queue recovery p95 | Pool p99 | Saturation | Correctness failures |
| --- | ---: | ---: | ---: | ---: | ---: | ---: |
| Healthy: c16 / 16 scans/s / 15 s | 240 / 240 | 8.79 scans/s | 6.697 s | 0 ms | 0 | 0 |
| First degraded: c32 / 32 scans/s / 15 s | 480 / 480 | 9.92 scans/s | 27.076 s | 1 ms | 1 | 0 |

The documented safe bounded-run envelope is the last repeated healthy rung: four workers, concurrency 16, 16 arrivals/s for 15 seconds, and 240 jobs. The first degraded rung preserves correctness but shows queue-recovery growth and the first pool-saturation interval. The guide explicitly warns not to treat backlog-drain throughput as a safe continuous-arrival target.

Terminating an active worker during the 480-job workload still produced 480/480 successful terminal jobs, no lost or double-completed work, and replacement readiness in 34.066 seconds. Evidence-chain verification remained intact. A separate 240-scan run grew the evidence ledger by exactly 240 rows; RLS-aware PostgreSQL relation/content growth was measured and documented.

### #591 isolation and recovery

- The replacement worker image passed all 12 strict production controls, including filesystem isolation, empty capabilities, default-deny network, seccomp, cgroup memory/PID limits, timeout/tree termination, executable integrity, bounded output, secret redaction, and recovery.
- Removing Bubblewrap makes `synapse-sandbox-check` fail closed with exit status 3; it does not run tools unsandboxed.
- A live tool run used the exact production runner and signed broker path: first setup accepted, exact still-valid replay rejected, sandboxed tool exit 0, replay journal grew once, and no residual namespace or Bubblewrap process remained.
- Active worker termination and Auto Scaling replacement preserved strict conformance.

## Architectural deviation from the original #591 proposal

The issue originally phrases fail-closed acceptance as the API refusing startup without Bubblewrap. The safer final boundary is different: production API replicas are dispatch-only and cannot instantiate the runner at all. The native worker is the sole untrusted-execution boundary, and it refuses startup/claims when Bubblewrap or another required strict control is absent. This avoids installing or enabling local execution capabilities in the horizontally scaled control plane while preserving the requested no-fallback guarantee where execution actually occurs.

## Relationship to #589 foundation

PR #648 already established two active API replicas, the load balancer, gated migrations, readiness/failover posture, and deterministic benchmark reduction. This PR adds the safe multi-worker queue fencing, native production execution tier, saturation/drain/replacement measurements, evidence-growth measurement, and reproducible safe operating envelope needed to complete the remaining acceptance criteria.

## Verification

All checks passed on the rebased commit:

- `pnpm --dir web build`
- `bash deploy/helm/synapse/testdata/render_test.sh`
- `bash deploy/aws/staging/tests/static.sh`
- `bash packaging/tests/static.sh`
- `terraform -chdir=deploy/aws/staging fmt -check`
- `terraform -chdir=deploy/aws/staging validate`
- `git diff --check upstream/main..HEAD`
- strict production-image sandbox conformance: 12/12 controls
- live signed-grant setup/replay/cleanup acceptance
- two-API admission and failover ladder
- four-worker healthy/degraded, termination/replacement, drain, pool, and evidence-growth runs

The web build passes with its existing large-chunk advisory warning.

## Publication hygiene

- Raw timestamped AWS observations, load-generator payloads, command outputs, and generated evidence bundles remain outside source control.
- Acceptance-only load-driver source is not included.
- No local AI endpoint/model configuration, runtime credentials, API tokens, signed grants, signing seeds, private keys, AWS credentials, or sensitive remote logs are included.
- User-owned protected paths (`.gitignore`, `.dockerignore`, `.serena/`, `.claude/`, and `CLAUDE.md`) are excluded.

Closes #589
Closes #591

## Review request

@nghiadaulau, please review the implementation, architecture decisions, measured acceptance evidence, and the #589/#591 mappings in this PR.


nghiadaulau (changes requested):
Reviewed in depth (dual go-arch + security audit of the crown-jewel execution-isolation boundary, plus local verification — CI is off, validated locally). **This is excellent, carefully engineered work.** The isolation model is sound and fails closed throughout; there are **no CRITICAL or HIGH findings**. I'm requesting two small changes to the conformance gate before merge, because they land on this PR's own security spine.

## Local verification — green
- `go build ./...`, `go vet`, `gofmt -l` all clean; cross-compiles for linux/amd64, linux/arm64, darwin/arm64.
- New/critical packages' unit tests pass: `egressgrant`, `executionmode`, `sandbox-check`, `sandbox`, `worker`, `egress`, `toolrunner`, `acquire`. egress-broker is 16/17 — the one failure is purely the macOS Unix-socket `sun_path` length limit (119 > ~104 chars); it passes on Linux.
- `0110_jobs_claim_fence` applies cleanly to a fresh Postgres; fence/claim/lease/jobqueue tests pass. (The one PG failure, `TestPostgresJobQueueAggregateJobQueueStatsAcrossTenants`, is **pre-existing on `main`** — a test helper seeds a tenant without the NOT NULL `name`; not introduced here.)

## Verified sound (the important part)
- **Grant model:** Ed25519 with `synapse-egress-grant-v1\0` domain separation over a canonical payload binding tenant/exec-kind/exec-id/run-id/slot/PID/issue+expiry/grant-id/rules; ≤5-min lifetime; signature verified and every claim re-checked field-by-field incl. `CanonicalRulesEqual` before any mutation — **a grant cannot be widened**.
- **Replay:** consumption is fsync-persisted (append + dir fsync) **before** the privileged setup; a failed/uncertain setup burns the grant and fails the next start closed; exact-replay + reuse rejected via in-memory + on-disk journal with `SameFile`/symlink/owner/mode TOCTOU hardening; concurrent setup/cleanup for one run id fenced.
- **Peer auth:** `SO_PEERCRED` uid == worker uid; sandbox PID opened via `pidfd` (PID-reuse-proof), proven a worker-owned descendant, proven the exact configured bwrap via `/proc/pid/exe` `SameFile` + required-flag cmdline check (`--unshare-all`/`--seccomp`/`--die-with-parent`/`--block-fd`; `--share-net` rejected), proven to hold private net/user/pid/mnt namespaces.
- **Scope server-side:** `egressgrant.Authorize` derives rules from the engagement scope (`compiler.Compile(eng.Scope)`), rejects any worker-supplied rules that differ, re-loads run state (`StatusRunning`), re-checks `LiveReconEnabled`, and re-runs `guard.Authorize` (scope + auth window) at grant time; only `execution_kind=recon` is issuable; issuer listener on a dedicated address with constant-time token compare, issuer token/seed forced distinct from API token + evidence seed.
- **Queue fencing:** `claim_fence` monotonic per claim; every terminal/heartbeat mutation gates on `status='claimed' AND claim_fence=$fence` → `ErrStaleLease`; a worker that loses its lease has its handler ctx cancelled and never publishes terminal state; async SCA evidence keyed on the durable job id keeps redelivery idempotent + byte-exact on the hash-chain.
- **argv-only (golden rule 1):** no `sh -c`/string-interpolated exec anywhere in production paths (all `sh -c` hits are `_test.go`); the broker wire protocol carries no command/argv fields. **dispatch-only** is structural fail-closed (`executionmode.DispatchOnly` errors every execution port; prod config refuses in-process; runner only constructed off the dispatch-only branch). **No secrets** in broker/client logs, errors, wire, or source.

## 🟡 Required before merge — harden the conformance gate to match its own claim
The PR's guarantee is a fail-closed startup gate that verifies the enumerated production controls. Two gaps mean the gate as shipped verifies less than it advertises:

1. **MEDIUM-1 — the blocking boot gate runs `-mode=startup`, not `-mode=full`.** `sandboxcheck.Run` only exercises `runner-ready` + `capabilities` + `network` in startup mode (`internal/composition/sandboxcheck/sandboxcheck.go:107-113`); the behavioral controls — seccomp actually denying ptrace, read-only-FS binds actually blocking writes / hiding host paths, binary-integrity (SHA-256) rejection, timeout kill, output-cap truncation, and secret redaction — are `fullChecks` only and never run by the gate that blocks the worker (systemd `ExecStart`, k8s `startupProbe`). Construction/`NewRunnerReady` do fail closed on the host/kernel-dependent legs (seccomp support, bwrap self-check, cgroup-v2 enforcement), and the in-process legs are deterministic Go, so this is defense-in-depth rather than an isolation break — but a regression that silently dropped e.g. the seccomp fd or `--remount-ro` wouldn't be caught at boot. **Please either run `-mode=full -strict` as the blocking boot gate, or document explicitly (in the systemd unit + helm template + the security guide) that startup mode intentionally verifies only host/kernel-dependent controls and `full` is the CI gate.**

2. **MEDIUM-2 — the k8s worker `startupProbe` drops `-strict`.** `deploy/helm/synapse/templates/worker.yaml:58` runs `-mode=startup` with no `-strict`, whereas systemd (`packaging/systemd/synapse-worker-sandbox-check.service:13`) and AWS staging (`deploy/aws/staging/worker.tf:156`) both pass `-strict`. Without it the check uses single-attempt `NewRunner` rather than `NewRunnerReady`, so the pod can report ready before cgroup delegation is confirmed (safe only because `runner.Run` fails each execution closed). This is a trivial, zero-downside parity fix — **add `-strict`** so "ready" implies limits are actually enforceable.

## 🟢 Non-blocking follow-ups
- **LOW-1 — replay journal never compacts** (`egressbroker/replay.go:102-177`): consumed grants append forever (only skipped on load); at the 600/min ceiling it grows unbounded and slows broker start. Roll/rewrite keeping only unexpired entries (atomic rename + dir fsync). Not a bypass.
- **LOW-2 — `SealWithID` non-exact content match** (`evidence/service.go:120-122,151-157`): scan evidence keyed on job id is idempotent + byte-exact to the *first* commit, but a materially divergent redelivery (recon/version drift) is silently accepted rather than flagged. Confirm this "first-commit authoritative" behavior is deliberate; if stricter binding is wanted, pass `exactContent=true` for scan evidence so a divergent redelivery surfaces `ErrConflict`.
- **arch nits:** `worker.go:132-134` logs "dead-lettering" on `TenantID.IsZero()` but just `return`s (job reclaimed on lease expiry → loops); park it via `Complete` to match intent. `egressbroker/server.go:117-127` accept loop has no `WaitGroup` draining in-flight handlers on shutdown (cancellation does propagate + `Cleanup` runs, so bounded). Consider extending `internal/platform/architecture_test.go` to also assert `usecase ↛ infrastructure`, not just `platform ↛ infrastructure`.

## ⚠️ Merge-ordering
`migrations/0110_jobs_claim_fence.sql` collides with PR #654 (also `0110`). Goose numbers must be unique — whichever lands second renumbers to `0111`. #654 is currently blocked on a separate HIGH, so this PR can keep `0110`; just flagging it.

Address the two conformance-gate MEDIUMs (MEDIUM-2 is one line; MEDIUM-1 is either a mode change or a documented contract) and I'll merge. Really strong PR — the grant/replay/peer-auth/fencing machinery is textbook.

nghiadaulau:
Pushed a fixup commit (`58c822cd`, on top of your `929b33b7`) that **addresses every finding from my review** — verified locally (build/vet/gofmt clean, all affected suites green, incl. a new compaction test; the one broker test that fails here is only the macOS Unix-socket path-length quirk). A focused security re-audit of the delta returned **SHIP — no regressions**, and the replay-compaction (the crown-jewel surface) was confirmed to preserve the no-replay invariant across every crash interleaving.

What changed:
- **MEDIUM-1/-2:** the fail-closed boot conformance gate now runs `-mode=full -strict` on every worker boot (systemd unit + k8s `startupProbe`, which also regains `-strict` with a wider timeout for the behavioral probes, + AWS staging validator). The gate now actually exercises seccomp/read-only-FS/binary-integrity/timeout/output-cap/secret-redaction before the worker may claim work, not just the host/kernel subset.
- **LOW-1:** the egress-broker replay journal now compacts — once it grows past 2× the live set (and a floor) it is rewritten keeping only unexpired records via owner-private temp → fsync → atomic rename → dir fsync. Crash-safe (old journal is a superset on any pre-rename crash; an expired grant can never be presented, so dropping expired records never enables a replay). Orphaned temp files are swept at load. Covered by a new test.
- **LOW-2:** documented the deliberate first-commit-authoritative semantics of the job-id-reserved scan evidence (kept `exactContent=false` — flipping it would break the commit-ambiguous redelivery idempotency the PR is built on).
- **arch nits:** the tenant-less worker guard no longer logs a "dead-lettering" it cannot perform; the broker accept loop drains in-flight handlers via a `WaitGroup` on shutdown; the architecture regression test now also asserts `usecase ↛ infrastructure`.

Your original `929b33b7` is preserved — merging via a merge commit. Thanks for the excellent PR; the grant/replay/peer-auth/fencing core needed no changes.

Issue #664, opened 2026-08-23
B2 · Endpoint visibility: network connection/flow entity (process-attributed)


nghiadaulau:
> Part of #594 · Phase **B** (endpoint visibility) · sub-item **B2**. Depends on **A1** (`NetworkObservation`, `ProcessEntityID`) and **B1** (process attribution).

## Goal
Turn raw network telemetry (A1 envelopes, `EventClass=network`) into a queryable **connection/flow entity** attributed to its originating process entity, so correlation can answer "which process opened this connection".

## Scope (portable core — eBPF collection deferred)
- In `internal/domain/endpoint`: `NetworkConnection { ConnectionID, TenantID, AssetID, ProcessEntityID, Proto, Direction, LocalAddr/Port, RemoteAddr/Port, Comm, FirstSeenAt, LastSeenAt, State }` with a stable `ConnectionID` = domain-separated, length-prefixed hash over (asset, process-entity, proto, direction, 5-tuple).
- Deterministic, event-time, idempotent fold projecting `connect`/`sendmsg` observations into connections + emitting `TimelineEntry` (B7). Re-observing the same flow widens `LastSeenAt`, never duplicates.

## Non-goals (tails)
Byte/packet counters + connection close (A1 network obs carries neither yet); live feed; persistence store.

## Tests
Stable ConnectionID; process attribution via `ProcessEntityID`; idempotent re-observation widens last-seen; egress/ingress + tcp/udp variants.

## Acceptance
A stream of network envelopes yields the correct set of `NetworkConnection`, each attributed to its process entity, deterministic + idempotent, with a timeline entry per new flow.

XUanhoa04:
I’m taking the explicit process-attribution honesty tail left by #670: network flows that reference an unobserved ProcessEntityID will carry an explicit unknown marker, then reconcile deterministically once that process is directly observed. I’ll include reorder-invariance and parent-stub regression coverage and avoid overlap with the open telemetry transport PR #654.

nghiadaulau:
**B2 network flow DONE.** `internal/domain/endpoint/connection.go`: `NetworkConnection` + stable domain-separated `ConnectionID`, `applyNetwork` fold widens LastSeenAt (no dup), `ProcessAttribution{unknown,observed}` with out-of-order reconciliation. Tests: connection_test.go.

[...]

PR #672, opened 2026-08-23
feat(endpoint): make network attribution gaps explicit (#664)

XUanhoa04:
## Summary

Completes the network-attribution-honesty tail called out by #670 for #664. A network flow now says explicitly whether its referenced process has actually been observed, instead of letting a populated `ProcessEntityID` imply evidence the projection may not have.

## Changes

- add validated `ProcessAttributionUnknown` / `ProcessAttributionObserved` domain states to `NetworkConnection`
- mark flows with a missing or unobserved process as `unknown`, while preserving any supplied entity id
- reconcile unknown flows when the matching process is directly observed, using a per-process connection index rather than a full scan
- keep `ProcessUnknown` parent stubs fail-closed: lineage references do not count as direct process evidence
- treat unknown-to-observed attribution resolution as a material `EndpointDiff` change
- cover network-before-process, process-before-network/reordered folds, parent stubs, missing ids, validation, and diff behavior

The reconciliation is monotonic and content-derived, so final entities remain deterministic and reorder-invariant. This is the portable honesty tail only; #664 remains open for its eBPF collection, persistence, counters, and close-event tails.

## Checklist

- [ ] `make build vet test typecheck` passes (GNU Make/Linux toolchain unavailable in this Windows environment)
- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `go test ./internal/domain/endpoint`
- [x] attribution/reorder regression tests repeated 100 times
- [x] Preserves the safety invariants

### Validation note

`go test ./...` passed every relevant package, including `internal/domain/endpoint`, but the aggregate command exits non-zero on two pre-existing Windows-only filesystem-security cases: `cmd/synapse-fptriage-release/TestRunCreatesPromotionAndRollbackLedger` (private input ACL verification is intentionally unavailable on Windows) and `internal/infrastructure/egressbroker/TestFileGrantReplayStoreCompactsExpiredRecords` (owner-controlled-directory check). `go test -race` is unavailable locally because this Go installation has CGO disabled.

Refs #664

### DIFF
--- a/internal/domain/endpoint/connection.go
+++ b/internal/domain/endpoint/connection.go
@@ -21,6 +21,21 @@ const (
 	ConnectionObserved ConnectionState = "observed"
 )
 
+// ProcessAttribution records whether the process named by a network observation has itself been
+// directly observed. A ProcessEntityID alone is not proof that the endpoint projection received the
+// corresponding process event: sensors can miss it, or network telemetry can arrive first.
+type ProcessAttribution string
+
+const (
+	// ProcessAttributionUnknown means the flow has no ProcessEntityID, or its referenced process has not
+	// been directly observed. The id is preserved when present, but consumers must not treat it as a
+	// verified process entity yet.
+	ProcessAttributionUnknown ProcessAttribution = "unknown"
+	// ProcessAttributionObserved means a validated process observation for ProcessEntityID is present in
+	// the same endpoint projection.
+	ProcessAttributionObserved ProcessAttribution = "observed"
+)
+
 // NetworkConnection is a process-attributed network flow reconstructed from network telemetry (B2). It is
 // keyed by a stable ConnectionID so re-observing the same flow widens its last-seen time instead of
 // creating a duplicate, and it links to the process that opened it by the A1 ProcessEntityID rather than a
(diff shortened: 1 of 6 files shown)


XUanhoa04:
@pho-veteran ready for review. This is the focused network-attribution-honesty tail explicitly left by #670 for #664: unknown is fail-closed for missing/unobserved process evidence, parent stubs do not upgrade it, and direct process observation reconciles it deterministically. Local domain tests, full build, and full vet are green; the two unrelated Windows-only full-suite failures are documented in the PR body.

pho-veteran (approved):
Reviewed for correctness, necessity, issue fit, clean architecture, and security.

No actionable P0–P3 findings. The change is still needed: #670 explicitly deferred this attribution-honesty tail, and the PR head is not already present on `main`. The implementation remains fail-closed, deterministic, reorder-invariant, and scoped to the endpoint domain. It correctly handles network-before-process, process-before-network, direct observations versus parent stubs, missing process IDs, validation, and material diff behavior.

Focused verification passed:
- `git diff --check`
- `go test ./internal/domain/endpoint -count=100`
- `go test -race ./internal/domain/endpoint`
- architecture review also confirmed `go build ./...` and `go vet ./...`

Approving this attribution-honesty tail. #664 should remain open for the separately deferred collection, persistence, counters, and close-event work.

PR #743, opened 2026-08-26
feat: add Python Tier-2 reachability and interprocedural taint

XUanhoa04:
## Summary

This adds an opt-in, deterministic Python Tier-2 semantic reachability engine and an interprocedural, value-slot-based taint engine. Target Python is parsed as source only; it is never imported, executed, compiled, or installed.

The implementation is intentionally split into seven reviewable commits but submitted as one feature PR.

## Architecture

```text
read-only workspace
  -> sandboxable synapse-ast/tree-sitter sidecar
  -> versioned + bounded Python semantic facts
  -> pure-Go semantic resolver/call graph
  -> Tier-2 reachability proofs
  -> pure-Go interprocedural value-flow graph
  -> gated CapSAST proposals
  -> distinct verifier confirmation
  -> persisted finding + SARIF codeFlows
```

- `internal/domain/pythonprogram` owns parser-independent facts, validation, semantic resolution, coverage gaps, and the conservative call graph.
- `internal/domain/taint` owns the Python value graph and class-specific source/sink/sanitizer catalog.
- `internal/usecase/pyreach` owns Tier-2 advisory-symbol reachability and Tier-1 supersession.
- `internal/usecase/taintscan` owns propose-only Python taint coordination.
- tree-sitter remains confined to `internal/infrastructure/tools/astwalk`; the application invokes it through the existing AST sidecar/provider seam.

### Tier-2 reachability

- Resolves lexical scopes, aliases, relative imports, local functions, constructors, `self`/`cls`, statically known receivers, conservative inheritance, and ambiguity.
- Matches exact PyPI PURLs plus advisory-provided affected symbols; it does not invent a callable from a distribution name.
- Partial graphs may prove a positive path. A Tier-2 `not_reachable` claim requires complete extraction and resolution; otherwise Tier-1 remains authoritative.
- Tier supersession and deterministic proof identities reuse the existing reachproof lifecycle.

### Interprocedural taint

- Tracks values through expressions, assignments, positional/keyword arguments, method receivers, parameters, returns, and selected attribute/container shapes.
- Keeps sibling calls separate and traversal bounded/cycle-safe.
- Models Flask, Django, and FastAPI request sources.
- Models SQLAlchemy/Django/raw SQL, command execution, path traversal, SSRF, XSS/template, unsafe deserialization, and redirect sinks.
- Sanitizers are taint-class-specific; e.g. HTML escaping cannot sanitize a command flow, and `yaml.safe_load` only neutralizes deserialization semantics.
- Findings are proposed at score zero under `system:python-taint-scan`; this coordinator has no verifier or self-confirmation port.

## Coverage honesty and product evidence

- Scan JSON exposes `complete`, `partial`, `unavailable`, and `not_applicable` semantic coverage, closed failure reasons, bounded fact counters, aggregated gap kinds, graph truncation, and proposal counts.
- Partial/unavailable analysis produces an operator warning; parser stderr, target paths, source text, expressions, and value IDs are excluded.
- Confirmed Python taint findings retain at most 64 canonical repository-relative source-to-sink positions.
- Migration `0115_finding_data_flow.sql` persists the trace as validated JSONB.
- SARIF 2.1.0 emits the sink as the primary location and the ordered witness as `codeFlows`/`threadFlows`, with coverage/truncation properties.

## Safety boundaries

- No target Python execution, import, package-manager invocation, or build hook.
- The sidecar accepts argv-only invocation and can run through the read-only sandbox runner.
- Facts, candidates, edges, files, nodes, proposal count, evidence frames, serialized bytes, and data-flow steps are bounded.
- Paths must be normalized, canonical, repository-relative, and traversal-free.
- Dynamic imports/execution, wildcard imports, unresolved calls/values, unsupported decorators/notebooks, parse recovery, unreadable files, and budget exhaustion become explicit coverage gaps.
- Positive evidence can survive partial coverage; missing coverage never becomes a clean/negative taint conclusion.

## Configuration and rollout

All new behavior is off by default:

- `SYNAPSE_PYREACH_ENABLED=true` enables existing Python Tier-1.
- `SYNAPSE_PYREACH_TIER2_ENABLED=true` enables Tier-2 and requires Tier-1 plus judgments.
- `SYNAPSE_PYTAINT_ENABLED=true` enables semantic taint proposals and requires judgments.
- `SYNAPSE_AST_BIN` may point to a CGO-enabled `synapse-ast`; the bundled/PATH lookup remains supported.

Recommended rollout: enable the CGO sidecar first, observe `analysis_coverage`, then enable Tier-2 reachability, and finally enable taint proposals with the normal distinct-verifier workflow.

## Validation

Passed locally on Windows/amd64:

- Focused domain, resolver, reachproof, pyreach, taint, coordinator, finding projection, memory/Postgres repository compilation, SARIF, rule catalog, config, SCA, AST provider, and CGO-free astwalk tests.
- `go vet ./...`
- `go build ./cmd/...`
- `pnpm run typecheck`
- `pnpm run build`
- `pnpm exec vitest run --maxWorkers=2` — 41 files / 327 tests passed.
- Synthetic value-flow regression corpus: TP=3, FP=0, FN=0, TN=2; complete=4, partial=1. Benchmark smoke run completed.

`go test ./...` reaches the same four pre-existing Windows-only/baseline failures recorded before implementation:

- private-file ACL verification in `cmd/synapse-fptriage-release`;
- owner-controlled replay-directory verification in `internal/infrastructure/egressbroker`;
- two existing telemetry failure-matrix assertions in `test/e2e`.

Environment-only skips/limitations:

- CGO parser golden tests cannot execute on this workstation because `CGO_ENABLED=0` and no C compiler is installed; CI must run the CGO-on sidecar job.
- PostgreSQL integration/migration tests compile but skip unless `SYNAPSE_TEST_DB_DSN` is set.
- `golangci-lint` is not installed locally; `go vet` is clean and CI remains authoritative for the configured lint suite.

## Review guide

The commit order follows the data path:

1. architecture/coverage plan;
2. versioned semantic facts and sidecar extraction;
3. pure semantic resolver;
4. Tier-2 reachability integration;
5. interprocedural value flow;
6. framework catalog and gated proposals;
7. product coverage, persisted witnesses, SARIF, hardening, docs, and migration.


nghiadaulau (changes requested):
Reviewed strictly with CI/CD off (verified locally — not re-enabling the pipeline). This is a well-structured, genuinely propose-only feature and the architecture is clean, but there are two hard blockers that must be fixed before merge.

## Blockers

### 1. Migration number collision — `0115` is already taken on `main`
This PR adds `migrations/0115_finding_data_flow.sql`, but `main` already ships `migrations/0115_endpoint_processes.sql` (merged in #750). Goose keys migrations by the numeric prefix, so two `0115` files will collide at startup (duplicate version 115) — the server won't boot after merge. The branch predates `0115_endpoint_processes.sql` (it isn't in the merge base), so this only surfaces post-merge.
**Fix:** renumber the migration to `0116_finding_data_flow.sql` (and update the matching test `internal/infrastructure/persistence/postgres/migration_0115_test.go` + any embed/reference) after merging the latest `main`.

### 2. Shipped test fails on a clean cgo build — `with … as` bindings are not extracted
`go test ./internal/infrastructure/tools/astwalk/` fails:

```
--- FAIL: TestPythonFactsForExtractsLoopComprehensionAndWithBindings
    python_facts_cgo_test.go:134: missing value-flow binding for "resource"
```

Root cause: `python_facts_cgo.go:162` handles the context-manager binding as
```go
case "with_item":
    e.bindingFact(node, scope, "alias", "value")
```
but in the pinned tree-sitter-python grammar (`smacker/go-tree-sitter v0.0.0-20240827094217`, grammar vendored in the module, so this is deterministic — not an environment issue), the `as <target>` lives in an inner `as_pattern` node; the `alias` field is on the `as_pattern`, not on `with_item`. So `node.ChildByFieldName("alias")` on the `with_item` returns nil → `targets` is empty → `bindingFact` returns early → no `resource` binding. There is no other `as_pattern` handling in the extractor.

Impact is not just a red test: every `with X() as y:` (e.g. `with open(path) as f`, `with requests.get(u) as resp`) is invisible to the taint value graph, which is a real false-negative for a security taint engine. Because a clean cgo checkout fails here, it looks like the `//go:build cgo` test path was never exercised before submission — please run `CGO_ENABLED=1 go test ./internal/infrastructure/tools/astwalk/...` locally.
**Fix:** descend into the `as_pattern` (read its `alias`/target and its subject `value`) for `with_item`, and keep the regression test.

## What's clean (verified locally)
- `go build ./...`, `go vet ./...`, and `gofmt -l` on all PR packages: clean.
- All other new/changed packages pass: `pythonprogram`, `taint`, `pyreach`, `taintscan`, `export` (SARIF), `sca`, `finding`, `judgment`, `findings`, `rulecatalog`, `config` — green.
- Architecture (dependency rule): PASS. `domain/pythonprogram` and `domain/taint` are pure (domain+stdlib only); `usecase/pyreach` and `usecase/taintscan` import only domain + `usecase/ports`; tree-sitter stays confined to `astwalk` behind the cgo tag; `sca/service.go` consumes the coordinator via `ports.TaintScanner`/`ports.TaintCoverageScanner`. `TestUsecaseDoesNotImportInfrastructure`/`TestPlatformDoesNotImportInfrastructure` pass.
- Propose-only invariant: PASS. `taintscan.PythonCoordinator` depends on a narrow `proposer` (Propose only, no Verify), emits `CapSAST` under `system:python-taint-scan`, seals no verdict, and never emits a clean/negative judgment (coverage is surfaced separately).

## Nit (non-blocking)
- `internal/domain/taint/python_valueflow.go:641` shadows the builtin `copy` (`copy := append(...)`); rename to `dup`/`sorted`.

## Security — no CRITICAL/HIGH (independently audited)
Adversarial pass against all six golden rules: PASS. Untrusted target is parse-only (tree-sitter in the `synapse-ast` sidecar, argv-only, never imported/compiled/executed/installed; `eval`/`exec`/`__import__` become coverage gaps). Wire document re-validated at the trust boundary; `ReferenceLiteral` drops literal values so secrets in source aren't captured; witnesses are positions/hashes; coverage reasons are a closed enum that forbids stderr/paths. Paths canonicalised + traversal-rejected on both write and read; resource caps + cycle-safe traversal throughout; migration JSONB validated/bounded in Go on write and read; sanitizers are taint-class-specific (HTML escaper can't neutralise a command flow). Coverage honesty holds: partial/incomplete extraction never yields a clean/negative conclusion.

Three non-blocking LOW / defense-in-depth items worth folding in while you're fixing the blockers:
- **LOW** `internal/infrastructure/tools/ast/provider.go:268` — the sandboxed-path error embeds child stderr (`truncate(string(res.Stderr),300)`), unlike the direct-exec path at `:288` which omits it. Today the sidecar only prints its own Go error text, but drop stderr here too to keep the "parser stderr never enters logs" invariant robust.
- **LOW** `cmd/synapse-api/main.go:81-84` — the Python taint provider is wired without the parity `log.Warn` that the Tier-2 reachability wiring emits at `:57-58` when `scaSandbox == nil`. Add the warning so an operator running Python taint unsandboxed in prod is told.
- **LOW** `internal/domain/taint/python_valueflow.go:530-571` — BFS stores a copied path slice per queued state; bounded by `maxPythonTaintWork` (4M) but the aggregate queued-path memory can grow large near the ceiling. Consider capping total enqueues / using parent-pointer back-links.

## Decision
**Changes requested.** The architecture and security posture are solid — but the migration-number collision (won't boot after merge) and the failing `with … as` extraction test (a real taint false-negative) are hard blockers. Fix those two, fold in the three LOWs + the `copy`-shadow nit if convenient, merge the latest `main`, re-run `CGO_ENABLED=1 go test ./internal/infrastructure/tools/astwalk/...` and `go test ./...`, then re-request and I'll re-verify + merge.

nghiadaulau (approved):
Both blockers fixed on the branch (pushed as `49198ddf`, with the latest `main` merged in) and re-verified locally with CI off:

- **Migration collision** → renamed the finding data-flow migration `0115` → `0116` (main ships `0115_endpoint_processes.sql`) and its integration test.
- **`with … as` false-negative** → added a `with_item` handler that reads the `as_pattern` subject + `alias` (`as_pattern_target`), let `targets` descend into `as_pattern_target`, and factored `bindingFactNodes` out of `bindingFact`. `TestPythonFactsForExtractsLoopComprehensionAndWithBindings` now passes on a CGO build.
- **LOW-1** ast provider no longer embeds child stderr in the sandboxed-path error (mirrors the direct-exec path); removed the dead `truncate` helper + import.
- **LOW-2** synapse-api warns when the Python taint provider runs unsandboxed (parity with Tier-2 reachability).
- **LOW-3** value-flow BFS now uses parent back-links + per-node depth instead of copying the full path into every queued state — O(nodes) memory near the `maxWork` ceiling, behaviour-preserving. Renamed the `copy`-shadowing local.

Verified: `go build ./...`, `go vet`, `gofmt` clean; taint / astwalk / ast / pyreach / taintscan / pythonprogram / postgres / finding suites green (CGO on). Full `go test ./...` shows only the two pre-existing macOS/main failures (`egressbroker` unix-socket-path and the fleet `test/e2e` telemetry-manifest test), both of which fail identically on `main` and are unrelated to this PR.

Architecture (go-arch) and security (no CRITICAL/HIGH) were already clean in the prior review. Approving + merging with a merge commit to preserve the commit history.

PR #756, opened 2026-08-28
feat(engagements): scan uploaded source packages

nnatuan03:
## Summary

Adds first-class source-package upload to the Scan Engagement flow, so users can create an Engagement and start a scan without exposing a repository URL or server filesystem path to the browser.

Supported packages are `.zip`, `.tar`, `.tar.gz`, and `.tgz`, capped at 512 MiB compressed.

## Changes

- Adds multipart Engagement creation with JSON `metadata` plus a binary `source` package.
- Stores source packages as tenant/Engagement-scoped artifacts in the shared object store and binds the Engagement scope to an immutable `uploaded-source/sha256/<digest>` target.
- Adds server-managed `kind: upload` scans: the client sends an empty target, the API resolves the stored locator, and the worker materializes the package before delegating to the existing archive extraction pipeline.
- Adds uploaded-source metadata retrieval and locks scan settings to the immutable package in Engagement Detail.
- Adds Create & Scan UX, upload validation, scan-start error recovery, OpenAPI coverage, operator docs, and end-to-end unit/UI tests.

## Security

- Normalizes uploaded filenames to a basename and allowlists archive extensions.
- Rejects empty or oversized uploads and caps request/metadata reads.
- Keeps object locators server-only and validates tenant/Engagement ownership before reads or scans.
- Verifies declared size and SHA-256 when storing, then re-verifies SHA-256 when materializing for every scan.
- Cleans temporary multipart files, materialized archives, extracted workspaces, and partial object-store writes on failure.
- Reuses the existing bounded archive extraction path, including traversal, symlink, and decompression-limit protections.

## Deployment Note

API and worker processes must point to the same `SYNAPSE_BLOB_*` bucket. The in-memory store is suitable only for single-process local development and is not durable across restarts.

## Validation — August 28, 2026

- `go test ./internal/domain/sourcepackage ./internal/infrastructure/sourceupload ./internal/infrastructure/blob`
- `go test ./internal/usecase/engagement ./internal/usecase/sca`
- `go test ./internal/adapter/httpapi ./api`
- `go test ./cmd/synapse-api ./cmd/synapse-worker`
- Full Vitest suite: 48 files / 363 tests passed
- `cd web && pnpm run build`
- Desktop and mobile browser QA passed with no console errors
- `go test ./...` only fails on pre-existing environment/flaky blockers: missing `helm` for the chart render test and `TestClientWaitReadyRetriesUntilBrokerListens` timing out in `internal/infrastructure/egressbroker`

## Checklist

- [ ] `make build vet test typecheck` passes — blocked only by the two unrelated full-suite issues listed above

### Documentation

- [x] New or changed `SYNAPSE_*` variables are in `docs/guide/configuration.md`, and security-relevant ones state the risk of a careless value — no new variables; shared-bucket requirements documented
- [x] New or changed CLI flags, subcommands, and exit codes are in `docs/guide/cli.md` — not applicable
- [x] New HTTP routes are described in `api/openapi.yaml` (do not grow `api/testdata/openapi-coverage-debt.txt`)
- [x] New guides are in `mkdocs.yml` nav, so the site actually publishes them — no new guide added
- [x] Claims match shipped behavior: no unbuilt artifact, unimplemented setting, or ingest-only format described as an export
- [x] Screenshots retaken if the documented UI changed, with no tokens, keys, or real customer data captured


pho-veteran (changes requested):
## Findings

### High - Reauthorize queued uploads at execution time

`internal/usecase/sca/service.go:1562-1588` and `internal/usecase/sca/service.go:1837-1880` execute the queued request without calling the shared execution guard again.

Impact: an uploaded package can be materialized and scanned after the engagement is completed/archived, its authorization window expires, its target is removed from scope, or its RoE changes. This violates the execution-layer invariant that authorization is checked immediately before every tool run. The queued `p.Now` also attributes the result to the earlier admission time.

Reproduction: enqueue an upload while its window is valid, hold the worker until the window expires, then deliver the job. A regression probe advanced the service clock from the admitted window to one hour later and observed `Acquirer.Acquire` being called.

Smallest safe fix: in `RunScanJob`, reload the current engagement and uploaded-package identity and call the shared guard with the current clock immediately before acquisition. Treat a denial as a blocked terminal job, and add a queued-upload test that changes the window/scope between enqueue and delivery.

### High - Bound ZIP parsing before `zip.OpenReader` allocates every entry

`internal/infrastructure/sourceupload/acquirer.go:38` routes user uploads into `internal/infrastructure/acquire/archive.go:111-121`. `zip.OpenReader` builds the complete `Reader.File` slice before the 200,000-entry loop guard runs; Go 1.27 `archive/zip.Reader.init` appends every central-directory header first.

Impact: a sub-512 MiB ZIP with millions of empty entries can exhaust worker memory before Synapse reaches its advertised entry limit. An OOM bypasses deferred cleanup and repeated delivery can also accumulate materialized archives/workspaces.

Reproduction: upload a ZIP containing more than 200,000 zero-length entries and start the scan. Memory grows during `zip.OpenReader`; the line 120 guard is reached only after all central-directory entries have already been allocated.

Smallest safe fix: reject ZIP source uploads until ZIP parsing/extraction runs in a fail-closed memory/CPU/disk-limited helper, or add a bounded central-directory preflight that rejects the archive before constructing per-entry objects. Add a high-entry-count regression test.

### High - Do not merge partial results from a different source identity

`internal/usecase/sca/service.go:2965-2969` always passes the engagement latest result to `mergeCachedScanResult`, while `internal/usecase/sca/service.go:3151-3179` never compares target/source identities.

Impact: a vulnerabilities-only upload scan can retain licenses and findings from an imported SBOM or another scoped target, then seal and persist the mixed result as if all findings came from the uploaded digest. This corrupts report and evidence attribution.

Reproduction: make an imported-SBOM or linked-target result latest for an upload-backed engagement, then run the upload in `vulnerabilities` mode. A regression probe with `current.Target=uploaded-source/sha256/...` and `previous.Target=imported-target` observed the previous license and finding being merged.

Smallest safe fix: merge complementary data only when immutable source identities match. For uploads require the same digest target; for Git/local targets also compare the relevant ref/commit identity. Add cross-target partial-scan coverage.

### Medium - Dispatch-only API and worker accept incompatible process-local stores

`cmd/synapse-api/main.go:649-656` and `cmd/synapse-worker/main.go:194-199` each create an independent `blob.Memory` when `SYNAPSE_BLOB_ENDPOINT` is empty. Production resolves the API to dispatch-only, but `internal/platform/config/config.go:1019-1031` does not require a shared blob endpoint.

Impact: the API returns 201 and queues the upload, but the standalone worker always gets `not found` from its empty memory store. An API restart also leaves persisted engagements permanently pointing at lost packages. Documentation warning does not prevent accepting work that cannot succeed.

Reproduction: run PostgreSQL-backed dispatch-only API plus worker with no blob endpoint, create an upload engagement, and start its scan. The API and worker hold different maps, so worker materialization fails.

Smallest safe fix: fail startup for cross-process execution without a shared object store, or disable the upload endpoint with a clear unavailable response in that topology. Add configuration/composition tests for API and worker.

### Medium - Portable transfer creates an unusable uploaded-source engagement

`internal/usecase/engagement/service.go:62` binds the digest target into engagement scope, but `internal/usecase/transfer/service.go:107-116` exports no source package and `internal/usecase/transfer/service.go:169-180` imports that scope under a new engagement ID.

Impact: the imported bundle appears source-backed, but `StartUploadedSourceScanWithOptions` looks up a package under the new ID and always returns not found. The supposedly portable engagement cannot repeat its scan.

Reproduction: export an upload-backed engagement, import the bundle, activate it, and run the source scan. The digest target survives but its tenant/engagement-scoped object does not.

Smallest safe fix: reject export/import of uploaded-source engagements until the bundle can carry and re-verify the archive, or version the bundle and transfer the package into the new tenant/engagement identity. Add a transfer round-trip test.

### Low - Replace the worker temp path in persisted SBOM identity

`internal/usecase/sca/service.go:2318` generates the SBOM from the extracted workspace and `internal/usecase/sca/service.go:2759-2766` stores that document unchanged.

Impact: `ScanResult.Target` is the immutable digest, but `SBOM.TargetRef` is a transient `synapse-ws-*` path. The latest-scan API therefore exposes a server filesystem path and records a non-repeatable SBOM identity.

Reproduction: complete an upload scan and inspect `sbom.TargetRef`. A regression probe returned the worktree temp directory instead of `uploaded-source/sha256/<digest>`.

Smallest safe fix: after cache/generation, set `doc.TargetRef = req.Value` for `TargetUpload` before manifests, findings, evidence, and persistence are built. Add serialization coverage.

## What this PR does well

The multipart boundary is physically capped, metadata decoding is strict, filenames are normalized and extension-allowlisted, ownership checks are tenant/engagement scoped, object bytes are size/hash verified on both save and materialization, and cleanup is composed across the materialized archive and extracted workspace. The client keeps locators server-only, preserves explicit scan-start recovery, and includes strong focused API/UI coverage. Dependency direction remains inward and external storage/acquisition are behind ports.

## Validation

Reviewed the PR description, complete base-to-head diff, all current discussions/reviews, and available checks. GitGuardian passed; no other GitHub checks were present.

Passed: focused tests for all changed Go packages plus transfer/acquire, `go vet` on related packages, builds for `synapse-api` and `synapse-worker`, web typecheck, 12 focused UI/API tests, the full 48-file/363-test Vitest suite, and the production web build.

Three temporary regression probes confirmed the execution-time authorization, cross-target merge, and SBOM identity failures; the probe file was deleted and the worktree was clean before submission. `go test -race` could not run because this Windows environment has CGO disabled.

nghiadaulau (approved):
Reviewed strictly with CI off (verified locally). Dual go-arch + security review. This is a carefully hardened upload path — the untrusted-archive threats are handled well: zip-slip/tar-traversal (safeJoin), symlink/device entries skipped, decompression-bomb caps (per-file 512 MiB, total 2 GiB, 200k entries via LimitReader), filename sanitization + extension allowlist, blob keys namespaced by sha256(tenant)/sha256(engagement) with no user-controlled component, server-computed SHA-256 re-verified at save/materialize/acquire, IDOR-safe (server-set locator, engagement∈tenant re-checked), and scope/auth-window enforced through the shared guard chokepoint. Dependency rule PASS, go vet + arch tests clean.

Fixes applied on the branch (`c8bc4b92`, latest main merged in):
- **MEDIUM (chain-of-custody)** — the upload wrote untrusted bytes to blob storage with no append-only audit entry (only the manifest carried the actor). Now records `engagement.source_uploaded` (actor, engagement, filename, sha256, size) on success, with a regression test.
- **minor** — the multipart hash path echoed raw io.Copy/Seek error text to the client; crafted size/empty errors are wrapped with ErrValidation and shown, a raw read error now goes through writeError (generic + logged).
- **minor** — the compensating source Delete is now tied strictly to the error path.

Accepted as-is (documented, LOW): in-process archive extraction (consistent with ADR-0007 pure-Go parsing), ParseMultipartForm /tmp spill under the per-request 512 MiB cap (bounded + PermOperate-authed), and the in-memory blob dev-only split-brain (already documented).

Verified: go build/vet/gofmt clean; engagement/httpapi/sourceupload/sca/sourcepackage suites green; web tsc/build/vitest 363/363. Approving + merging.

PR #757, opened 2026-08-28
fix(sca): harden remediation and license reporting

nnatuan03:
## Summary
- enforce exact package identity plus ecosystem-aware affected/fixed version validation
- select the minimum compatible remediation, preserve alternate fixes, and mark unresolved versions non-actionable
- preserve raw/SPDX license evidence, policy rationale, dependency evidence, and all affected locations
- expand API, frontend types, and Excel remediation/license audit output with backward-compatible export fallback

## Validation
- targeted Go tests for all touched scanner/domain packages: passed
- full `internal/usecase/sca` test suite: passed
- `CI=true pnpm install --frozen-lockfile`: passed
- `pnpm run typecheck`: passed
- Excel export regression suite: 4/4 passed
- `pnpm test`: 47/47 files, 358/358 tests passed
- `pnpm build`: passed (existing large-chunk warning only)
- `go test ./...`: relevant code packages passed

## Environment-only full-suite notes
- Helm chart tests could not run locally because the `helm` binary is not installed
- `TestClientWaitReadyRetriesUntilBrokerListens` failed only under the long macOS temporary path; rerunning with `TMPDIR=/tmp` passed


pho-veteran (changes requested):
## What is working well

- The change preserves the clean-architecture dependency direction. Domain and use-case packages remain inward-only, and the changed infrastructure adapters retain compile-time port assertions.
- The API additions are additive, the frontend mapper supplies defaults for older scan payloads, and the Excel path includes a deliberate legacy fallback.
- Bumping `CorrelationVersion` and retaining per-source detections, alternate fixes, rejected fixes, raw/SPDX license data, and all component locations are the right auditability goals.
- The new remediation and license columns make the exported result substantially more explainable.
- Focused tests for every changed Go package, `go vet ./...`, normal and CGO-free builds, web typecheck/build, and the 4 Excel regression tests passed in review.

## Blocking findings

### High

1. **A candidate rejected by one source can still be recommended** (`internal/domain/vulnerability/correlate.go:389-415,440-462`). The `rejected` set never removes a candidate from `byVersion`. I reproduced OSV rejecting `1.2.0` while Grype accepts it; the result is still `FIX_AVAILABLE` with `1.2.0`. Exclude rejected candidates using ecosystem comparator equality and emit no actionable recommendation for accepted/rejected disagreement. Also return correlator-local rejections in `RejectedFixedVersions`.

2. **Unknown ecosystems can recommend a downgrade** (`internal/domain/vulnerability/correlate.go:476-488`). Installed `2.0.0` plus candidate `1.0.0` and an empty ecosystem becomes an actionable fix because the fallback only checks that the strings differ and look installable. Fail closed when no comparator exists, or derive a validated ecosystem from the PURL first.

3. **The correlation key can merge distinct package identities** (`internal/domain/vulnerability/correlate.go:252-264`). A versionless PURL merges different installed versions; the PURL-less fallback omits ecosystem; qualifiers such as distro and architecture are stripped. I reproduced `demo@1.0.0` and `demo@2.0.0` becoming one vulnerability. Key by canonical ecosystem, package, installed version, and identity-relevant qualifiers.

4. **Explicit affected versions are ignored while validating fixes** (`internal/infrastructure/tools/osv/scanner.go:241-250,344-375`, `internal/infrastructure/tools/ownadvisory/source.go:117-140`). A version can appear as both a fixed event and in `affected[].versions`, yet Synapse recommends it because both validators call `advisory.Affected` with a nil explicit-version list. Parse and aggregate explicit versions and reject internally inconsistent candidates.

5. **License metadata can hide a mandatory obligation** (`internal/infrastructure/tools/license/scanner.go:517-565`). `MIT;link=https://example.test AND GPL-3.0-only` is truncated to `MIT`, changing a deny obligation into an allow result. Accept metadata only as a fully parsed terminal suffix; trailing operators or license tokens must fail closed.

### Medium

6. **Exact identity is lost again after correlation and in Excel** (`internal/usecase/sca/service.go:3496-3507,3532-3564`, `web/src/lib/excelExport.ts:269-279,427-433`). Scope, first-party state, dependency paths, and locations are joined by name/version rather than `PackagePURL`. With npm and PyPI `demo@1.0.0`, the npm vulnerability inherited PyPI test/first-party metadata, and two vulnerabilities produced four Excel rows. Use one canonical identity key end to end.

7. **Dependency evidence is both misclassified and over-trusted** (`internal/usecase/sca/service.go:1246-1252,2105`). A direct `[app, dependency]` path is labeled `RESOLVED_TRANSITIVE`, while client-declared edges from an imported SBOM are labeled `CONFIRMED_DEPENDENCY`. Treat paths of length at most two as direct and carry graph provenance so imported edges remain client-declared/unverified.

8. **A recommendation can violate the configured policy** (`internal/infrastructure/tools/license/scanner.go:220-250`). `GPL-3.0-only OR LGPL-3.0-only` recommends GPL when it appears first because both options have high generic risk, although the default policy denies GPL and only warns on LGPL. Rank OR choices by policy verdict first, then risk, then a canonical tie-breaker.

9. **Component dedupe loses new audit data and is input-order dependent** (`internal/domain/sbom/dedupe.go:57,67-70,116-126`). A test-scoped duplicate seen first keeps the merged shipping package at test scope, and a later duplicate carrying the only `RawValue` for the same SPDX ID is discarded. Production should dominate background scopes, and duplicate license records should merge missing raw evidence.

10. **Version ranges are reported as resolved versions** (`internal/domain/sbom/sbom.go:269-280`, `internal/usecase/sca/service.go:1158-1161,3512-3527`). `1.x`, `1.*`, and `1.2.3 || 2.0.0` all become `RESOLVED`. Use ecosystem-aware concrete-version validation rather than checking only the first character.

11. **Raw license metadata can disclose credentials through the API and workbook** (`internal/infrastructure/tools/license/scanner.go:147-151`, `internal/usecase/sca/service.go:1199-1213`, `web/src/lib/excelExport.ts:650-664,748-763`). An imported license URL containing userinfo or a token is preserved verbatim in `Raw License`. Apply the shared redaction path before API/report output and keep restricted raw evidence by hash/reference if needed.

12. **Equivalent fix spellings create a false advisory conflict** (`internal/domain/vulnerability/correlate.go:395-425,491-513`). Two sources reporting `1.2.3` and `v1.2.3` produce `ADVISORY_CONFLICT` plus a duplicate alternative even though the ecosystem comparator considers them equal. Deduplicate and intersect candidates by comparator equality while retaining one deterministic display spelling.

The existing changed-package tests pass, but focused regression probes reproduced each behavior above. I recommend addressing the High findings before merge and adding the corresponding regressions.

nghiadaulau (approved):
Reviewed strictly with CI off (verified locally). Dual go-arch + security review: dependency rule PASS, no exec/secret/ReDoS/panic issues. The remediation branch-selection, validated-fixed-version, and license reclassification logic are sound.

Both reviewers flagged real coverage/false-negative risks in the OSV adapter (a security scanner must not silently miss vulns), which I fixed on the branch (`f5479d14`) and re-verified:
- **HIGH** — the identity-status gate dropped whole OSV-supported ecosystems our resolver doesn't model (Composer, Hex, Pub, Swift, Conan) from the query, silently reporting zero OSV vulns for them. Now OSV is queried by versioned PURL for every non-distro package; identity only enriches fix data. The emit gate drops a match only when a RESOLVED identity shows the affected blocks are a different artifact (your wrong-artifact FP fix is preserved); unresolved identity trusts OSV's PURL match. `matchingAffected` now compares names with EqualFold.
- **MEDIUM** — on disjoint cross-source fix sets the recommendation was the minimal (most-likely-still-vulnerable) version; now recommends the maximal candidate (most likely actually patched).
- **LOW** — fixed a slice-aliasing mutation in `mergeComponents`; dropped dead `ValidVersion`; documented `CompareVersions`/`FixedVersions`; renamed a shadowing local.

Added a regression test that an OSV match for an unmodeled ecosystem is emitted. Verified: `go build`/`vet`/`gofmt` clean; osv/vulnerability/sbom/advisory/sca suites green; web `tsc`/build/vitest 358/358 green + the excelExport harness 4/4 (compiled + run). Your intended false-positive reductions (OS-distro exclusion, wrong-artifact drop) are kept intact.

Approving + merging.

PR #754, opened 2026-08-27
feat(privacy,httpapi): A6 detection query field-RBAC, read audit, and CI test stabilization (#627)

VietSory:
## Overview

This PR completes the **A6: Detection Query Field Governance & Read Audit** implementation ([Refs #627](https://github.com/KKloudTarus/synapse-ce/issues/627)) and resolves all remaining test-harness failures on Linux CI. All CI workflows (Backend Linux, Backend Windows, Go Lint, Frontend Web, and AI Triage) are passing 100% green without requiring parallelism workarounds (`-p 1`) or diagnostic flags.

---

### A. Field-Level RBAC Projection on Detection Query Surface
- **Role-Based Field Filtering**: Implemented role-aware field projections (`ProjectDetectionForPrincipal`) in `internal/adapter/httpapi/detection_privacy.go`.
  - **`readonly`**: Sensitive direct-value fields (process path, argv arguments, remote IP addresses, file paths, and sensitive command names) are stripped/redacted to enforce data minimization.
  - **`analyst` / `reviewer` / `admin`**: Retain full evidence payload post source-side redaction for investigative and audit workflows.
- **Deep-Copy Projection**: Query results are deeply cloned prior to projection, preventing any caller from mutating ledger-owned records in memory.
- **Response Metadata**: Detection query responses declare the effective `field_scope` applied during projection.

### B. Fail-Closed Read Audit
- **Pre-Access Audit Recording**: The API handler emits a `detection.query` audit entry to the audit log *before* yielding access to sensitive detection data.
- **Fail-Closed Security**: If the audit logger encounters an error or database unavailability, the request fails closed immediately (HTTP 500), preventing unaudited reads.

### C. Source-Side Redaction Residuals Hardening
- **Glued Password Arguments**: Resolves residual `-p<value>` parsing edge cases for MySQL / MariaDB command-line invocations.
- **Internal-Space Credentials**: Handles credentials containing internal spaces without truncating or skipping trailing tokens.
- **Authoritative Process Scoping**: Argv inspection leverages authoritative process identity (`Path`/`Comm`) to guarantee correct redaction even when `argv[0]` is omitted or forged.

---

## 2. Test Harness Stabilization & CI Fixes

The production detection/privacy logic from #627 was validated as sound. Investigation revealed that the remaining Linux CI failures stemmed from runner permission constraints and shared integration database state contamination:

[...]

PR #758, opened 2026-08-28
feat(fleet): durable detection provenance, honest telemetry coverage, audit outbox

### LINKED ISSUE #610: A0.5 · Atomic evidence persistence (SealOnce/outbox; fixes D3)
> Phase **A0** · child of #594 · **P0** · fixes defect **D3**.

## Goal
Make evidence sealing **idempotent across the failure boundary** and the permanent evidence link **self-contained**.

## The bug this fixes (D3)
`detectledger.Ingest` does `HasDetection`(`service.go:167`) → `Seal`(`:173`) → `AppendDetection`(`:188`) as **separate** calls. `Seal` and `AppendDetection` are not in one transaction, so a retry after a post-`Seal` failure **re-seals** — duplicate permanent evidence.

## Scope
- Replace the read-then-seal-then-append flow with one of:
  - **A:** `evidence.SealOnce(idempotency_key)` (seal is itself idempotent), or
  - **B:** DB transaction inserts a pending detection + an **outbox** event → an evidence writer seals → updates the evidence id, or
  - **C:** deterministic evidence id + a reconciliation worker.
  Do **not** attempt a distributed transaction across columnar + relational + chain — use a saga/outbox with explicit intermediate state.
- Seal a **self-contained `DetectionEvidenceEnvelope`** (Detection + tenant/agent/session/asset + detection-batch identity + telemetry refs + rulepack digest + signing `KeyID` + content hash) so the permanent chain link does not depend on an expirable projection.
- Track **provenance state**: `Received → TelemetryDurable → CommitmentPending → CommitmentSealed → Acknowledged`; a detection citing not-yet-durable telemetry is `pending|complete|expired|broken`.

## Acceptance / exit
- [ ] A retry after a post-seal failure does **not** double-seal (proven by a fault-injection test at that boundary).
- [ ] The permanent evidence record is self-contained (reconstructable without the projection).
- [ ] Provenance state is tracked and queryable.

## Dependencies
Needs **A0.1** (canonical identity in the envelope) + **A0.2** (signing `KeyID`).

### LINKED ISSUE #611: A0.6 · Loss semantics (Sampled != Truncated != Dropped; fixes D2) + source-side privacy
> Phase **A0** · child of #594 · **P0** · fixes defect **D2**.

## Goal
Honest loss accounting (**Sampled ≠ Truncated ≠ Dropped**) + **source-side privacy** classification — so the coverage/confidence scores never lie and raw telemetry never leaks secrets by default.

## The bug this fixes (D2)
On overflow the service keeps the **prefix** and calls it a sample:
```
accepted.SampleRate = maxInt(batch.SampleRate, ceilDiv(len(events), s.budget))
events = events[:s.budget]        // telemetry/service.go:101-102 — tail truncation
```
Keeping the prefix drops the tail (where a late malicious sequence lives) and records it as an elevated `SampleRate`. Also `recordGap` ignores its audit-write error.

## Scope — loss semantics
- **Three distinct outcomes**, each recorded distinctly, never all mapped to `SampleRate`:
  - **Sampled** = intentional policy with a named algorithm · **Truncated** = batch cut by a limit · **Dropped** = observed then lost under pressure/failure.
- Replace tail-truncation with either a representative sampling policy **or** record it explicitly as `Truncated` with a reason.
- Batch manifest carries the three **separate** loss counts `ObservedCount, KeptCount, SampledOutCount, TruncatedCount, DroppedCount` (Sampled ≠ Truncated ≠ Dropped at batch level) + `SamplingPolicyDigest` (= digest over `{SamplingAlgorithm, SamplingPolicyID, Seed, Version}` — the reproducible policy identity). The runtime disambiguating detail — per-event `LossDisposition` and `TruncationReason` — lives on the **first-class `TelemetryLoss`/`TelemetryGap`** object (shipped #618), not folded into the manifest's policy digest.
- **Never-sample list (hard):** detections · privilege transitions · response-verification · sensor-state · coverage gaps · critical file mutation · key/config tampering · agent stop/update/decommission.
- `TelemetryGap`, `SensorState`, `CoverageWindow` become **first-class persisted objects**, not best-effort audit lines; `recordGap` must not swallow audit-write errors.

## Scope — source-side privacy
- Field classification `allow | redact | hash | drop`; **no environment variables collected by default**; bounded argv/path length; redact known secret patterns; safe defaults + tenant-configurable policy; ensure batch commitments don't retain PII after telemetry deletion.

## Acceptance / exit
- [ ] Overflow is recorded as `Truncated`/`Dropped` (not an elevated `SampleRate`); never-sample classes are never dropped/sampled.
- [ ] `TelemetryGap`/`SensorState`/`CoverageWindow` are queryable first-class objects; a gap is never presented as a complete window.
- [ ] Secrets/env are redacted at the source; a redaction test with a known secret passes.
- [ ] Coverage/confidence inputs reflect true loss.

## Dependencies
Needs **A0.1** (identity) + **A0.3** (schema). Feeds A2/A3 and the Phase-C **Coverage axis** (`RiskAssessment.CoverageVector` — the coverage-honesty signal, NOT a `RiskContext` member; `RiskContext` is `{Threat,Exposure,Behavior}`, #638).

pho-veteran:
Closes #610. Closes #611.

Both child issues share one mechanism — durable, retry-safe facts that cannot diverge from their audit record — so they land together.

## #610 — retry-safe detection provenance

- Durable `PendingDetectionV2` plus a self-contained detection evidence envelope, so a crash between admission and sealing is recoverable instead of lost.
- Provenance lifecycle (`Received` → `TelemetryDurable` → `CommitmentPending` → `CommitmentSealed` → `Acknowledged`, plus `Expired`/`Broken`) persisted as append-only, hash-chained custody history. Unique predecessor-link indexes make a forked provenance chain unrepresentable.
- `EvidenceChain.SealOnce` remains the only no-double-seal primitive. Concurrent first admission arbitrates by insert-first (`ON CONFLICT DO NOTHING`) — never check-then-seal.
- Detection signing keeps legacy v2 separator framing and adds unambiguous v3 framing. New attributed batches emit v3; unsupported versions have no signing bytes and fail validation.
- Provenance is reachable over HTTP and by query, so custody is auditable rather than internal-only.

## #611 — honest telemetry loss and deterministic coverage

- Source privacy is applied **before** WAL serialization, with a non-relaxable privacy floor, immutable policy history alongside a separately mutable active pointer, and a per-event `RedactionPolicyDigest`. Delayed telemetry is authorized against the historical digest that actually applied.
- Truthful accounting: sampled / truncated / dropped are distinct, `TruncatedCount <= KeptCount`, zero-kept manifests are valid, and protected classes get backpressure instead of silent loss.
- Signed telemetry gaps are durable — an append-only signed revision ledger plus a mutable coalesced projection. P3 WAL eviction persists gap evidence *before* deleting anything. An exact signed-digest retry is a no-op; incompatible `GapID` reuse is a conflict.
- Deterministic, revisioned coverage windows over half-open `[Since, Until)` intervals: state before `Since` participates as effective state, facts at `Until` are excluded, effective sensor state is selected per detection class, and unhealthy state prevents a "complete" claim. `CreatedAt` is excluded from revision identity so late facts produce new revisions rather than mutating old ones.
- Unknown-coordinate gaps degrade coverage honestly instead of fabricating counts or an epoch-anchored span.

## Durable fleet audit outbox

The review finding that drove this: **state mutation could precede its mandatory audit entry with no durable record of the obligation.**

The mutation and its *exact* immutable audit payload now commit in one tenant-local transaction (or one memory critical section). The payload is then delivered to the hash-chained audit log via `RecordOnce`, and only then acknowledged. A crash in between leaves the intention **pending**, and a per-tenant reconciler repairs it at startup and on a periodic sweep. No distributed transaction is introduced — this is a local outbox, not 2PC.

Covers privacy activation, telemetry batch commitment, signed gap revisions, and signed sensor state, in both PostgreSQL and memory.

One deliberate semantic note: the batch-commit audit means *batch identity became durable*, not "ingest complete". Event persistence and stream-state advancement remain separately audited.

## Defects found by running the persistence layer against a real PostgreSQL 17

This code had never met a live database before. Seven real bugs surfaced:

| Defect | Consequence |
| --- | --- |
| Composite mutators returned nothing | Callers audited a re-derived payload; an immediate delivery and a restart-time delivery could hash two different entries for one obligation |
| Outbox queries relied on RLS alone | Under a `BYPASSRLS`/superuser role, one tenant's sweep could read **and retire** another tenant's audit obligation |
| Memory outbox used `tenant + ":" + id` string keys | Intention ids contain colons, so tenant `a` with intention `b:x` collided with tenant `a:b` and intention `x` — a genuine cross-tenant isolation hole |
| `RecordOnce` collapsed on idempotency key without comparing content | An obligation could be acknowledged (and later reclaimed) though its exact payload never entered the chain |
| Gap extension compared full-precision timestamps | `timestamptz` truncates to microseconds, so honest widening carrying nanoseconds read as evidence *shrinking* and was rejected |
| pgx scans `timestamptz` in the session zone | An exact activation retry returned a value not identical to first admission |
| Memory privacy store resolved an empty tenant id | Collapsed to the *default* tenant, rejecting every other tenant's own sweep |

`FleetAuditIntent.Normalize` is now the single canonicalizer every backend shares, so one obligation cannot hash differently per backend.

## Review feedback addressed

**Merged latest `main`** — resolved both conflicts as directed. In `internal/domain/privacy/detection.go` I kept this branch's `MaxArgCount` bound but switched to main's command-scoped `redactArgvForCommand(p.Args, commandIdentity)` from #754, with command identity from `Process.Path` falling back to `Comm`. Ordering is deliberate and commented: the **count** bound first (dropping trailing elements can only remove a credential, never split one), then redaction, then the per-element **length** bound — so a length bound can never cut a recognized credential and leave a fragment behind. In `jobqueue_test.go` I took main's `randHex(t)` tenant form.

Main had also independently fixed the three test failures this branch was carrying, and did it better, so I dropped my versions in favour of main's: `logicalQueueTenantIDs` subsumes the inline dedup I had written for the same legacy empty-string-tenant bug (main's has its own unit test), and main's `setupJobQueue` provisions a real `NOSUPERUSER NOBYPASSRLS` role so the cross-tenant aggregation assertion now runs under *enforced* RLS instead of skipping. Main's changes also fixed the two Windows OS-ACL failures this body previously documented — they pass now.

**LOW / Rule 3 — salt exposure: fixed.** `newPrivacyPolicyDTO` is split by plane. The human plane omits `HashSalt` entirely — including on the admission echo, and for the admin who authored the policy, since history and the active pointer expose policies other actors authored. The agent plane keeps it via `newAgentPrivacyPolicyDTO`, because agents apply source privacy before telemetry leaves the host and that route is agent-credentialed and tenant-checked. The salt is still accepted on admission **input** (a hash disposition is invalid without one), so only the response direction changed; the digest still identifies a policy exactly, so governance callers lose no audit ability.

The regression test asserts on **raw response bytes**, not the decoded struct — `hash_salt` is `omitempty`, so a decoded struct reads `""` whether the field was omitted or present-and-empty, and only the wire bytes prove the salt never reached the client. It also asserts the agent plane *still* receives the salt, so a future over-correction that strips it everywhere fails loudly rather than silently breaking source-side hashing. I did not go with `PermAdminister` gating: the salt has no read-side use for a human at any privilege level, so omitting it is the narrower fix.

I also corrected the OpenAPI description, which called the salt "Non-secret" — it now documents the actual plane boundary.

**MINOR / arch — port placement: fixed.** `sensorstatejournal` no longer imports from `internal/usecase/fleet/sensorstateship`. The types moved to `internal/usecase/ports` as `SensorStateJournal` / `SensorStateDeliveryState` / `PendingSensorStateReport`, with `ValidateDeliveryState` becoming a `Validate()` method on the state to match the neighbouring port structs. Verified no infra file imports a non-`ports` usecase package.

**MINOR / arch — agent binary logic: not done, deliberately.** Moving the agent's retry/backoff and batch-framing out of `package main` into a `usecase/fleet/telemetryship` would restructure the agent binary well beyond these two issues, and you flagged it as optional and pre-existing. Happy to take it as a follow-up if you'd rather it land now.

The salt fix is **not** vault-backed yet — that is the "ideally" half of your suggestion and a larger change (key management, rotation, migration of existing plain-JSONB policies). I'd rather do it as its own PR than bolt it on here; say the word if you want it in scope.

## Migrations

0117–0125, all additive, RLS-forced, with append-only triggers and `ON DELETE RESTRICT`. No shipped migration was edited.

Migration 0125 is worth calling out: retention may reclaim an **already-delivered** intention, because the audit chain is the permanent record — but deleting an **undelivered** one is refused, since that row is the only durable proof that committed state still owes an audit entry.

## Reviewing this

Five commits, meant to be read in order:

1. **`feat(fleet): durable detection provenance, honest telemetry coverage, audit outbox`** — the implementation.
2. **`refactor(fleet): tighten audit outbox API surface and remove dead code`** — hygiene only: unexports `insertFleetAudit`, drops a dead duplicated conflict check, un-splices a doc comment, uses `DefaultInterval`.
3. **`fix(jobqueue): stop double-counting the legacy empty-string tenant`** — the queue-depth bug found while clearing the last two red tests (largely superseded by main's `logicalQueueTenantIDs` in the merge below).
4. **`Merge upstream/main`** — the two conflict resolutions above.
5. **`fix(privacy): keep the redaction hash salt off the human read plane`** — the review fixes.

Contents are source only. No fixtures, build artifacts, or logs. The `migration_NNNN_test.go` files follow the existing convention in this package (30 such files already at base); each covers its own migration's RLS, immutability triggers, and rollback refusal, so they are not duplicates.

## Verification

- `go build ./...` and `go vet ./...` clean; `gofmt` clean on changed files; `git diff --check` clean.
- **`internal/infrastructure/persistence/postgres` is fully green** against a disposable PostgreSQL 17 cluster via `SYNAPSE_TEST_DB_DSN`.
- **Every other package passes too** — `./internal/... ./cmd/... ./test/...`, zero failures. The two Windows OS-ACL failures noted in earlier revisions of this description are resolved by the merge.
- Architecture and security review passes were run against this branch; the blocker and all majors were remediated before this review, and your findings are addressed above.

`-race` was **not** run: CGO is unavailable in this environment, so treat race coverage as unverified rather than passing.

## Open question (unchanged)

**Postgres sweeps one audit-outbox store, memory sweeps three.** All three PostgreSQL repositories embed one `FleetAuditRepository` over the single `fleet_audit_intents` table, so the composition root registers one store in Postgres mode and all three in memory mode (where each owns a private map). Recovery is correct either way. If you'd rather the outbox be genuinely per-aggregate, that wants an aggregate discriminator column — I'd do it as a follow-up rather than widen this PR.

(The earlier `TenantLister` question is resolved: it stays as the small local interface, consistent with how `detectledger`'s runner already type-asserts its own tenant lister.)


nghiadaulau (changes requested):
Reviewed strictly with CI off (dual go-arch + security review, verified locally after merging current main into the branch to resolve conflicts and running build/vet/test). This is an unusually well-engineered PR — the evidence-integrity, append-only, tenant-isolation, and coverage-honesty properties that are its heart are enforced defensively at every layer (DB trigger → repo → use case → adapter): hash-linked provenance chain with `UNIQUE(...,previous_hash)` fork-guard + `VerifyChain` on append and read, idempotent `SealOnce`, `synapse_forbid_mutation` UPDATE/DELETE/TRUNCATE triggers on every evidence/audit table (migrations 0117-0125), server-authoritative agent identity/tenant on every ingest (no IDOR), coverage windows derived server-side taking the worst per-class state (an agent can't hide a server-observable gap), parameterized SQL, and bounded ingest. go vet + arch tests clean; the 26 changed packages' tests pass. **go-arch: PASS.** **security: no CRITICAL/HIGH.**

A few items to address before merge — I'm requesting changes rather than patching them myself so they land with your intent:

**LOW (security / Rule 3 — please fix): redaction hash-salt is over-exposed.**
`internal/adapter/httpapi/privacy_policy_handler.go` returns `HashSalt` in `privacyPolicyDTO` via `newPrivacyPolicyDTO`, which feeds the human read routes `getActivePrivacyPolicy` / `listPrivacyPolicyHistory` (both `authz(PermView,...)`); it's also persisted as plain JSONB (`migrations/0121`). The salt is what makes `DispositionHash` pseudonymization resistant to dictionary/rainbow attacks over low-entropy telemetry (usernames, paths, comm). Returning it to every `PermView` principal (who may also read hashed telemetry) lets an insider de-anonymize it. The agent plane (`fleet_handler.go activePrivacyPolicy`) legitimately needs the salt to hash source-side, but the human read DTOs do not. Fix: omit `HashSalt` from `newPrivacyPolicyDTO` on the human read paths (or gate it behind `PermAdminister`), and ideally treat it as vault-backed material rather than plain policy JSONB.

**MINOR (arch, optional):**
- `internal/infrastructure/sensorstatejournal/store.go` imports its port from `internal/usecase/fleet/sensorstateship` — the only new infra file that takes a port from a usecase package instead of `internal/usecase/ports`. The other ~12 new ports are correctly consumer-defined in `usecase/ports`; relocate `StateStore`/`DeliveryState` there for consistency.
- The agent binary (`cmd/synapse-agent/telemetry_shipper.go`) carries unit-tested retry/backoff, batch-framing, and ingest/digest-assembly logic in `package main`; per "cmd/* is composition root only" this ideally lives in a `usecase/fleet/telemetryship`. Pre-existing fat-edge-agent pattern, so optional.

**Also:** the branch currently conflicts with `main` (2 files — `internal/domain/privacy/detection.go` vs the #754 command-scoped `redactArgvForCommand`, and a `jobqueue_test.go` tenant-isolation line). Please merge the latest `main`: in detection.go, keep your `MaxArgCount` count-bound but call `redactArgvForCommand(p.Args, commandIdentity)` (main's newer command-scoped redaction) rather than `RedactArgv`; in jobqueue_test.go take main's `randHex(t)` form.

Decision: **changes requested** — address the salt exposure + merge main; the two arch items are optional. This is strong work and I'll re-review promptly.

pho-veteran:
Thanks for the careful review — all four items are addressed and this is ready for re-review. @nghiadaulau 

**LOW / Rule 3 — salt exposure (fixed).** `newPrivacyPolicyDTO` is now split by plane. The human plane omits `HashSalt` entirely — including on the admission echo and for the admin who authored the policy, since history and the active pointer expose policies other actors authored. The agent plane keeps it via `newAgentPrivacyPolicyDTO`, because agents hash at the source and that route is agent-credentialed and tenant-checked.

I went with omission rather than `PermAdminister` gating: the salt has no read-side use for a human at any privilege level, so omitting it is the narrower fix. The salt is still accepted on admission **input** — a hash disposition is invalid without one — so only the response direction changed, and the digest still identifies a policy exactly.

Two details worth flagging, since they're easy to get wrong:
- The regression test asserts on **raw response bytes**, not the decoded struct. `hash_salt` is `omitempty`, so a decoded struct reads `""` whether the field was omitted or present-and-empty — only the wire bytes prove the salt never reached the client.
- It also asserts the agent plane **still receives** the salt, so a future over-correction that strips it everywhere fails loudly instead of silently breaking source-side hashing.

I also fixed the OpenAPI description, which described the salt as "Non-secret" — that was contradicted by this change.

On the vault-backed half of your suggestion: not in this PR. It means key management, rotation, and migrating existing plain-JSONB policies, which I'd rather do as its own change than bolt on here. Happy to open it as a follow-up, or pull it in if you want it before merge.

**MINOR / arch — port placement (fixed).** `sensorstatejournal` no longer imports from `usecase/fleet/sensorstateship`. The types are now `ports.SensorStateJournal` / `SensorStateDeliveryState` / `PendingSensorStateReport`, with `ValidateDeliveryState` becoming a `Validate()` method on the state to match the neighbouring port structs. I verified no infra file imports a non-`ports` usecase package.

**MINOR / arch — agent binary logic (not done, deliberately).** Restructuring the agent's retry/backoff and batch-framing out of `package main` reaches well beyond these two issues, and you marked it optional and pre-existing. Say the word if you'd rather it land now.

**Merge (done).** Both conflicts resolved as you directed: `detection.go` keeps the `MaxArgCount` bound but calls `redactArgvForCommand(p.Args, commandIdentity)`, and `jobqueue_test.go` takes main's `randHex(t)` form. The ordering there is now commented, because it matters: count bound → redaction → length bound, so a length bound can never cut a recognized credential and leave a fragment.

One thing worth calling out from the merge: **main had already fixed the three test failures this branch was carrying, and did it better than I had.** `logicalQueueTenantIDs` subsumes the inline dedup I'd written for the same legacy empty-string-tenant bug (and has its own unit test), and main's `setupJobQueue` provisions a real `NOSUPERUSER NOBYPASSRLS` role so the cross-tenant aggregation assertion runs under *enforced* RLS instead of skipping. I dropped my versions in favour of main's. Main's changes also fixed the two Windows OS-ACL failures I'd documented as unverified — those pass now.

**Verification after all of the above:** `go build` / `go vet` / `gofmt` / `git diff --check` clean; the whole `postgres` package green against a disposable PostgreSQL 17; and `./internal/... ./cmd/... ./test/...` fully green with zero failures. `-race` remains unverified — CGO is unavailable in my environment, so I'm not claiming it passes.

The one open question in the description is unchanged: in Postgres mode all three repositories share one `fleet_audit_intents` table, so the composition root registers one outbox store there and three in memory mode. Recovery is correct either way; a genuinely per-aggregate outbox wants a discriminator column, which I'd do as a follow-up unless you want it here.

nghiadaulau (approved):
Re-reviewed strictly with CI off (verified locally). Both findings from my previous review are properly addressed:

- **LOW — redaction hash-salt off the human plane (fixed correctly).** `newPrivacyPolicyDTO` (the human read routes) now strips `HashSalt`, and a separate `newAgentPrivacyPolicyDTO` keeps it only for the agent-credentialed plane (agents must hash at source). There's a dedicated regression test `TestPrivacyPolicyHashSaltNeverLeavesTheAgentPlane` (admission echo, active-as-admin, active-as-readonly) asserting the salt never reaches a human principal, plus the digest still identifies the policy so governance loses nothing. Exactly the right fix.
- **Arch advisory folded in**: the `sensorstateship.StateStore` port I flagged was relocated into `internal/usecase/ports/sensor_state_journal.go`, so no infra file imports a usecase package for its port anymore.
- **Conflicts resolved**: latest `main` merged in; `git merge-tree` against current `main` (after #753 merged) shows 0 conflicts, MERGEABLE/CLEAN.

Verified locally: `go build ./...` clean, `go vet` clean, and the changed fleet/privacy/sensorstate packages' tests pass (incl. the salt regression test). The one thing the branch left imperfect was gofmt — 7 files (1 new `activation.go` + 6 modified) had struct-field-alignment violations; I applied `gofmt -w` (whitespace only, no logic change) and pushed it as `192cce37` so the branch is gofmt-clean. The go-arch + security substance was already strong in my prior review (hash-chained provenance, append-only triggers, tenant isolation, coverage honesty all enforced at every layer).

Approving + merging.

Issue #694, opened 2026-08-24
Epic: Assessment lifecycle, re-test chains, and trustworthy scan comparison


nnatuan03:
## Objective

Introduce a first-class **Assessment Cycle** so one initial Assessment and every Assessment Re-test can be managed, compared, audited, and explicitly closed without mutating historical scan evidence or existing Finding/SLA workflows.

## Gate decision

Architecture direction: **GO**.

Production implementation starts only through the delivery slices listed below. Semantic claims such as `Fixed`, `Re-opened`, `Final`, and `Cycle completed` must come from immutable, versioned inputs and explicit policy, never mutable current rows or incomplete scans.

## Canonical ownership and terminology

| Concept | Owns | Does not own |
|---|---|---|
| Engagement / Assessment | Scope, RoE, authorization window, execution lifecycle, scan/evidence container | Cross-Assessment ancestry or closure |
| Assessment Cycle | Membership, rooted ancestry, selected head, final path, cycle lifecycle | Scanner execution facts or Finding workflow |
| Scan Run / lane manifest | Immutable execution facts and producer provenance | Snapshot selection or lifecycle classification |
| Assessment Snapshot | Frozen comparison inputs, selected run lanes, coverage dimensions, content hash | Finding workflow/disposition |
| Finding Identity | Semantic correlation within one Cycle | Severity, version, evidence, assignment, SLA |
| Finding Observation | Immutable state of one identity in one Snapshot | Cross-Snapshot lifecycle claim |
| Comparison | Directional presence classification, change flags, immutable summary | Finding/SLA mutation or workflow carry-forward |
| Existing Finding/SLA/retest workflows | Triage, accepted risk, false positive, remediation, assignee, deadline, explicit verification | Scanner presence truth |
| Closure Manifest | Frozen final path and all effective immutable decision/projection references | Reconstructing history from mutable rows |

**Assessment Re-test** means a new Engagement/Assessment in a Cycle. Existing per-Finding retest means a governed verification decision for one Finding; it does not create an Assessment.

## Business Asset / Project boundary

Cycle persists one frozen boundary key: `standalone | asset | project | asset_project` plus exact non-null IDs required by that kind.

| Assessment association | Initial cycle | Re-test/member rule |
|---|---|---|
| Asset + Project | Allowed only when tenant-scoped `business_asset_projects` link exists | Exact same Asset and Project required |
| Asset only | Allowed | Exact same Asset and no Project required |
[...]

PR #779, opened 2026-08-30
feat(scanrun): promote scan_runs to tenant-owned and add sealed provenance v1 (#708)

### LINKED ISSUE #708: A2a · Tenant-scoped scan_runs promotion and sealed provenance v1
> Part of #694 · Prerequisite for #696. First production slice after #695.

## Goal

Promote current five-column `scan_runs` into a tenant-owned execution header and add normalized, sealed producer-lane provenance tables so Snapshot coverage uses queryable, constrained facts instead of opaque JSON assertions.

## Existing table decision

**Migrate `scan_runs` in place; do not replace it.** Retain `manifest` and `finding_keys` only as legacy source evidence and backward-compatible API data. Add normalized native provenance alongside it.

Existing rows are backfilled with tenant from their Engagement and permanently classified:

```text
provenance = legacy
terminal_status = unknown
```

Legacy `manifest`/`finding_keys` alone can never yield `complete` coverage. Only sealed native lane rows plus versioned server policy may do so.

## Required DDL shape

Exact migration may adjust constraint names, not semantics:

```sql
ALTER TABLE scan_runs ADD COLUMN tenant_id TEXT;
UPDATE scan_runs r
SET tenant_id = e.tenant_id
FROM engagements e
WHERE e.id = r.engagement_id;
ALTER TABLE scan_runs ALTER COLUMN tenant_id SET NOT NULL;

ALTER TABLE scan_runs ADD CONSTRAINT scan_runs_tenant_engagement_id_unique
  UNIQUE (tenant_id, engagement_id, id);
ALTER TABLE scan_runs ADD CONSTRAINT scan_runs_tenant_id_unique
  UNIQUE (tenant_id, id);
ALTER TABLE scan_runs ADD CONSTRAINT scan_runs_engagement_fk_tenant
  FOREIGN KEY (tenant_id, engagement_id)
  REFERENCES engagements(tenant_id, id) ON DELETE CASCADE;

ALTER TABLE scan_runs
  ADD COLUMN provenance TEXT NOT NULL DEFAULT 'legacy'
    CHECK (provenance IN ('native','legacy')),
  ADD COLUMN terminal_status TEXT NOT NULL DEFAULT 'unknown'
    CHECK (terminal_status IN ('building','succeeded','partial','failed','cancelled','unknown')),
  ADD COLUMN manifest_schema_version INT NOT NULL DEFAULT 0 CHECK (manifest_schema_version >= 0),
  ADD COLUMN manifest_hash TEXT CHECK (manifest_hash IS NULL OR length(manifest_hash) = 64),
  ADD COLUMN sealed_at TIMESTAMPTZ;
```

```sql
CREATE TABLE scan_run_lanes (
  tenant_id                   TEXT NOT NULL REFERENCES tenants(id),
  engagement_id               TEXT NOT NULL,
  scan_run_id                 TEXT NOT NULL,
  lane_key                    TEXT NOT NULL CHECK (btrim(lane_key) <> ''),
  producer                    TEXT NOT NULL CHECK (btrim(producer) <> ''),
  terminal_status             TEXT NOT NULL
    CHECK (terminal_status IN ('building','succeeded','partial','failed','cancelled')),
  target_kind                 TEXT NOT NULL CHECK (btrim(target_kind) <> ''),
  target_identity_schema_version INT NOT NULL CHECK (target_identity_schema_version > 0),
  target_identity_canonical   TEXT NOT NULL CHECK (btrim(target_identity_canonical) <> ''),
  evaluated_revision          TEXT NOT NULL DEFAULT '',
  authoritative_finding_kinds TEXT[] NOT NULL,
  included_scope              JSONB NOT NULL CHECK (jsonb_typeof(included_scope) = 'array'),
  excluded_scope              JSONB NOT NULL CHECK (jsonb_typeof(excluded_scope) = 'array'),
  started_at                  TIMESTAMPTZ NOT NULL,
  finished_at                 TIMESTAMPTZ,
  result_ref                  TEXT NOT NULL DEFAULT '',
  evidence_ref                TEXT NOT NULL DEFAULT '',
  result_sha256               TEXT CHECK (result_sha256 IS NULL OR length(result_sha256) = 64),
  manifest_schema_version     INT NOT NULL CHECK (manifest_schema_version > 0),
  manifest_hash               TEXT NOT NULL CHECK (length(manifest_hash) = 64),
  sealed_at                   TIMESTAMPTZ,
  PRIMARY KEY (tenant_id, scan_run_id, lane_key),
  FOREIGN KEY (tenant_id, engagement_id, scan_run_id)
    REFERENCES scan_runs(tenant_id, engagement_id, id) ON DELETE CASCADE
);
```

```sql
CREATE TABLE scan_run_lane_versions (
  tenant_id    TEXT NOT NULL,
  scan_run_id  TEXT NOT NULL,
  lane_key     TEXT NOT NULL,
  version_kind TEXT NOT NULL
    CHECK (version_kind IN ('tool','scanner','profile','rule_pack','advisory_database','correlation','schema')),
  name         TEXT NOT NULL CHECK (btrim(name) <> ''),
  version      TEXT NOT NULL CHECK (btrim(version) <> ''),
  digest       TEXT NOT NULL DEFAULT '',
  PRIMARY KEY (tenant_id, scan_run_id, lane_key, version_kind, name),
  FOREIGN KEY (tenant_id, scan_run_id, lane_key)
    REFERENCES scan_run_lanes(tenant_id, scan_run_id, lane_key) ON DELETE CASCADE
);

CREATE TABLE scan_run_lane_stages (
  tenant_id   TEXT NOT NULL,
  scan_run_id TEXT NOT NULL,
  lane_key    TEXT NOT NULL,
  stage_key   TEXT NOT NULL CHECK (btrim(stage_key) <> ''),
  status      TEXT NOT NULL CHECK (status IN ('succeeded','failed','skipped')),
  reason_code TEXT NOT NULL DEFAULT '',
  started_at  TIMESTAMPTZ,
  finished_at TIMESTAMPTZ,
  PRIMARY KEY (tenant_id, scan_run_id, lane_key, stage_key),
  FOREIGN KEY (tenant_id, scan_run_id, lane_key)
    REFERENCES scan_run_lanes(tenant_id, scan_run_id, lane_key) ON DELETE CASCADE
);
```

Add tenant/cursor/coverage indexes for `(tenant_id, engagement_id, created_at DESC, id)`, lane producer/target/status, version lookup, and failed/skipped stages. Enable/force tenant RLS on all tables following migration `0103_sla_governance.sql` ownership pattern.

## Canonical target identity v1

Coverage equality uses the tuple:

```text
(target_kind, target_identity_schema_version, target_identity_canonical, evaluated_revision)
```

`target_identity_canonical` is server-derived. Raw client target text is retained only as redacted provenance, never used for equality.

- **Repository:** lowercase scheme/host, strip userinfo/default port/trailing slash/`.git`, normalize percent encoding and dot segments, preserve path case unless a versioned provider adapter declares case-insensitive ownership. Resolve refs/tags to immutable full commit digest; `main`, tag, or short SHA alone is not an evaluated revision.
- **OCI image:** lowercase registry host and repository per OCI rules; canonical identity includes repository plus immutable `sha256:<64 lowercase hex>` digest. Mutable tag alone cannot prove equivalent coverage.
- **Host:** DNS lowercased, trailing dot removed, Unicode converted to IDNA ASCII; IP parsed and rendered with standard canonical form. Port/service scope is explicit and producer-versioned, never inferred from display text.
- **URL:** lowercase scheme/host, IDNA host, remove fragment/default port, normalize percent encoding and dot segments. Query treatment is producer-policy versioned; ignored/sorted keys are explicit. Credentials and tokens are stripped/redacted.
- **Cloud resource:** lowercase provider plus provider-versioned canonical account/project/subscription and stable resource ID. Display name, region alias, and console URL are not identity.

Every kind supplies equality/non-equality golden vectors, including URL/repo casing, `.git`, default ports, Unicode hostnames, IPv6 forms, mutable refs/tags versus immutable revisions/digests, and cloud display-name changes. Canonicalization version changes create a new identity domain; old versions remain readable.

## Expand/backfill/validate/cutover/contract

1. **Expand:** add nullable `tenant_id`, native provenance columns, `(tenant_id,id)` and `(tenant_id,engagement_id,id)` unique constraints/indexes, normalized child tables, and temporary compatibility reads. Keep original `PRIMARY KEY (id)` and every existing ID-only FK/query working.
2. **Discover dependents:** migration test queries PostgreSQL catalogs for every FK/index referencing `scan_runs(id)`. Any unlisted dependent fails the migration fixture.
3. **Backfill:** populate `tenant_id` from Engagement in bounded batches; legacy rows receive `legacy/unknown`. Backfill tenant columns on dependent tables before adding composite FKs. Orphan/ambiguous rows hard-fail with counts/sample IDs.
4. **Validate:** add composite FKs as `NOT VALID`, validate after reconciliation, prove `(tenant_id,id)` uniqueness, and run dual-read parity between ID-only and tenant-scoped repository paths.
5. **Application cutover:** repository APIs require `(tenant_id, run_id)`; native writes dual-write all normalized facts. Instrument/deprecate ID-only reads and require zero production calls for one release window.
6. **FK cutover:** rebind every dependent FK to tenant-first keys. Future FKs to Scan Run must include tenant and, when Engagement-owned, engagement ID.
7. **Contract:** only after validated zero ID-only callers/FKs, replace `PRIMARY KEY (id)` with `PRIMARY KEY (tenant_id,id)` and remove temporary global `UNIQUE(id)` compatibility shape.

Rollback before application/FK cutover disables new writes and removes new structures only when no native rows exist. Rollback after cutover keeps expanded columns/constraints and uses a dual-read-compatible application version; it must not restore ID-only ownership after cross-tenant duplicate run IDs become possible.

## Sealing and trust rules

- New native run starts `building`; terminal transaction writes lanes, versions, stages, artifact digests, canonical hashes, then seals header/lanes.
- Server derives terminal status, target, versions, scopes, stages, and authoritative kinds from owned execution state. Client cannot submit them as facts.
- Sealed native header/lane/version/stage rows are update-immutable; correction appends replacement provenance.
- `succeeded` requires non-empty authoritative kinds, terminal stages, versioned canonical target identity, immutable evaluated revision where required, producer-required versions, artifact digest/reference, and matching canonical hash.
- `partial|failed|cancelled|unknown|legacy` never become complete coverage.

## Compatibility

- Existing scan history/list/drift reads continue from retained columns through expand/backfill/validate.
- Repository port gains tenant ID and normalized lane reads; ID-only compatibility adapter is temporary, instrumented, and removed only in contract phase.
- Legacy row without matching Engagement is a migration hard failure with count/sample IDs; do not assign default tenant silently.

## Tests and acceptance

- [ ] Upgrade fixture proves every legacy row gets correct tenant, composite FK, and unknown coverage.
- [ ] Catalog-driven fixtures cover zero, one, and multiple dependent ID-only FKs plus pre/post-cutover rollback.
- [ ] Target canonicalization golden vectors prove equivalence/non-equivalence for repo, image, host, URL, and cloud kinds.
- [ ] Fresh/native runs persist queryable target, producer/lane, terminal state, versions, stages, scope, kinds, and artifact hashes.
- [ ] Client cannot forge coverage facts.
- [ ] Sealed immutability, canonical hash, transaction failure, concurrent terminal write, RLS, and cross-tenant FK tests pass.
- [ ] Existing ScanRun API/drift compatibility, metrics/logs/redaction, and repository checks pass.

## Non-goals

- Snapshot selection/comparability logic.
- Finding identity matching.
- Upgrading unverifiable legacy rows to native provenance.

VietSory:
## Summary
Closes #708 (Phase A2a of Epic #694).

Promotes `scan_runs` from a lightweight manifest record into a tenant-owned execution header and introduces normalized, queryable, sealed producer-lane provenance child tables (`scan_run_lanes`, `scan_run_lane_versions`, `scan_run_lane_stages`).

## Key Deliverables

1. **Target Identity v1 Canonicalization (`internal/domain/scanrun`)**:
   - Server-derived canonicalization for `repository`, `oci`, `host`, `url`, and `cloud_resource`.
   - Immutable revision enforcement (40-hex/64-hex commit digests, `@sha256:` image digests).

2. **Deterministic Provenance & Trust Contract**:
   - Deterministic SHA-256 manifest hash calculation (`ComputeManifestHash`) invariant to map and slice ordering.
   - Complete coverage predicate: `IsCompleteCoverage()` is `true` iff `provenance == 'native'`, `terminal_status == 'succeeded'`, and every child lane has non-empty authoritative finding kinds and valid target identity. Legacy or partial runs fail-closed to non-complete coverage.

3. **PostgreSQL Migration 0134 (`migrations/0127_scan_run_provenance.sql`)**:
   - Promotes `scan_runs` with `tenant_id`, `provenance`, `terminal_status`, `manifest_schema_version`, `manifest_hash`, `sealed_at`, `updated_at`.
   - Backfills `tenant_id` from `engagements` with zero-orphan verification.
   - Enforces composite unique constraints and composite foreign keys to `engagements(tenant_id, id) ON DELETE CASCADE`.
   - Creates child tables: `scan_run_lanes`, `scan_run_lane_versions`, `scan_run_lane_stages`.
   - Implements immutability trigger `fn_prevent_sealed_scan_run_mutation` on sealed rows.
   - Enables and forces Row Level Security via `CALL synapse_enable_tenant_rls(...)`.
   - Safe Down migration with RLS teardown.

4. **Persistence & Service Implementation**:
   - Extended `ports.ScanRunStore` and implemented `memory.ScanRunStore` with structured key `scanRunKey{TenantID, ID}`.
   - Implemented `postgres.ScanRunStore` with transaction-bound execution, row locking (`SELECT ... FOR UPDATE`) during sealing, and SQL error translation.
   - Application service (`internal/usecase/scanrun/service.go`) coordinating validation, hashing, sealing, and audit logging.

5. **Test Suite**:
   - Target canonicalization and domain model tests.
   - Goose migration inventory duplicate version guard test.
   - Migration schema and complete Rollback (`DownTo 126`) / Reapply (`UpTo 127`) tests.
   - Multi-tenant isolation and real unprivileged role RLS (`NOSUPERUSER NOBYPASSRLS`) tests.
   - Concurrent competing sealers serialization race test.
   - Microsecond timestamp precision roundtrip tests.
   - In-memory / PostgreSQL service parity tests.


nghiadaulau (changes requested):
## Strict review — CHANGES REQUESTED (not merged)

Reviewed locally (CI off): `go build` + `go test ./internal/domain/scanrun/... ./internal/usecase/scanrun/...` are **green**; go-arch **PASS** on the dependency rule; security audit confirms **tenant isolation + sealed immutability are solid**. The design is strong — but there are two must-fix items (one blocking, one high-value) plus parity/attestation gaps. Details below.

### 🔴 BLOCKER — migration number collision with `main`
`migrations/0127_scan_run_provenance.sql` collides with `0127_legal_holds.sql` (already merged, #790), and `main` is also at `0128_response_verification.sql` (#793). Two `0127_*` files break goose ordering. **Renumber to `0129_scan_run_provenance.sql`**, rename `internal/infrastructure/persistence/postgres/migration_0127_test.go` → `..._0129_test.go` (and any `0127` references inside), and **rebase onto current `main`**. This branch was cut before the privacy/response migrations landed.

### 🟠 HIGH — the run/header `manifest_hash` is order-dependent (breaks the "order-invariant" + reproducibility contract)
`internal/usecase/scanrun/service.go` (~L154-184) builds the aggregate hash by concatenating the per-lane hashes in caller-supplied `in.Lanes` order, with no sort. But lanes are persisted and re-read sorted by `lane_key` (`scan_run_repo.go` ~L457). Consequences:
- **Idempotent re-seal breaks:** the guard compares `existingHash == manifestHash` (memory ~L198 / postgres ~L340); a reordered but logically-identical re-seal fails equality → `ErrConflict "already sealed with different state"` instead of a no-op.
- **Reproducibility weakened (golden-rule 5):** a verifier recomputing the header hash from the stored, canonically-ordered lane set gets a different value → the header commitment isn't reproducible from persisted data.

Fix: sort the per-lane hashes by `lane_key` before aggregating (match the read order), and move the aggregation into a **domain** function (e.g. `scanrun.ComputeRunManifestHash([]Lane)`) — `crypto/sha256`/`encoding/hex` shouldn't live in the usecase layer as business logic. Add a **two-lane seal test asserting identical `ManifestHash` across submission orders** (it would currently fail, proving the bug). The per-lane `ComputeManifestHash` is correctly order-invariant already — only the aggregate is affected.

### 🟡 MEDIUM
1. **Memory repo cross-tenant read when ctx has no tenant.** `memory/scan_run_repo.go` `List` (~L74-84) / `Get` (~L100-108): when `shared.TenantFrom(ctx)` returns `ok==false`, the tenant filter is skipped and **all tenants'** runs are returned — while the postgres adapter falls back to `DefaultTenant` and stays isolated. Mirror the postgres `DefaultTenant` fallback (or hard-fail on missing tenant) for parity. (RLS protects the real DB path; this is the memory/test store, so latent, but it's a tenant-isolation parity gap worth closing.)
2. **Per-lane `ComputeManifestHash` omits `LaneKey` + `TerminalStatus`** (`model.go` envelope ~L296-308). The lane's identity and success/failure aren't attested by its own hash (only `Stages[].Status` is). Two lanes with identical facts but different `lane_key` hash the same. Not a live tamper hole (triggers protect stored `terminal_status`), but an attestation-completeness gap — include both in the canonical envelope.

### 🟢 LOW / nits
- **Postgres `SealScanRun` omits the `IsTerminal()` + `sealed_at.IsZero()` guards the memory store enforces** (`postgres/scan_run_repo.go` ~L313-316 vs memory ~L180-185) — port contract enforced inconsistently; add both for parity.
- **`TestComputeManifestHash_RandomizedPermutations` isn't randomized** (`model_test.go` ~L79-172): the permute helpers depend only on element index, so all 50 iterations apply the same fixed reordering with no seed — rename or drive with a seeded RNG.
- **Hardcoded `manifestSchemaVersion = 1`** on seal (`service.go` ~L188/203) — derive from the lanes/a named constant so a future v2 lane isn't mislabeled.
- **`IsCompleteCoverage` trusts the stored `manifest_hash` without recomputing** (`model.go` ~L282) — optional hardening: recompute per-lane to confirm stored facts match.
- **Document that `ON DELETE CASCADE` from engagements is vetoed for sealed runs** (the BEFORE-DELETE trigger raises) — correct chain-of-custody behavior, but the CASCADE is partly aspirational; operators should expect the delete to fail, not silently purge sealed provenance.

### ✅ Verified solid (credit where due)
Dependency rule clean (domain imports only domain+stdlib+idna; compile-time `var _ ports.ScanRunStore` on both adapters). Tenant isolation is defense-in-depth: `FORCE ROW LEVEL SECURITY` on all 4 tables + explicit `tenant_id=$1` predicate on every query + composite PKs/FKs making cross-tenant child rows structurally impossible + the **zero-orphan backfill genuinely fails closed** (`RAISE EXCEPTION` on any remaining NULL/'' before `SET NOT NULL`). Sealed immutability is complete: whole-row `IS DISTINCT FROM` triggers on the header + all 3 child tables (INSERT/UPDATE/DELETE), correct seal ordering (children while unsealed → header last, one locked tx), no re-parenting of sealed children. Target-identity canonicalization rejects URL/scp userinfo and forces full-hex/`sha256:` revisions (no secret smuggling), all identity flows into **parameterized** SQL. `IsCompleteCoverage` is correctly fail-closed — no legacy/partial run can over-claim.

Please fix the BLOCKER + HIGH (and ideally the two MEDIUMs) and re-request review — happy to re-review promptly.

VietSory:
@nghiadaulau All review findings have been resolved and verified against current main (migration 0129, domain-owned order-invariant aggregate hashing, memory DefaultTenant fallback, LaneKey/TerminalStatus in lane hash, postgres seal validation parity, and seeded RNG tests).

nghiadaulau:
**Reviewed locally (CI disabled) — strict review. Strong foundation, but holding merge on 1 medium + 2 low findings on the audit/attributability path.**

## Verified clean ✅
- **Build/test:** `go build ./...` + `go vet` clean; `domain/scanrun`, `usecase/scanrun`, `persistence/memory` tests green. Postgres tests (`scan_run_repo_test.go`, `migration_0129_test.go`) **skip** locally (no DB on this host) — SQL reviewed by hand instead.
- **Architecture:** dependency rule holds — `domain/scanrun` is pure (stdlib+domain), `usecase/scanrun` depends only on domain + ports (`TenantTransactionRunner`, `ScanRunStore`, engagement reader, audit). Legacy `ScanRunStore` methods preserved for back-compat. Idioms clean (`%w` wrapping, `New*` validation, context-first, tenant-scoped params).
- **Tenant isolation (rule 2):** migration 0129 enables RLS + FORCE on all 4 tables (`synapse_enable_tenant_rls`), composite FKs `(tenant_id,engagement_id)→engagements` and `(tenant_id,scan_run_id)→scan_runs`, plus explicit `WHERE tenant_id=$1` in the repo — defense in depth. `SealScanRun` derives `engagement_id` from the stored row, not caller input.
- **Sealed provenance (rule 5):** `ComputeManifestHash`/`ComputeRunManifestHash` are deterministic + order-invariant (sorted, nil/empty normalized, timestamps/identity excluded, SHA-256). Seal-once enforced at 3 layers: service (atomic via `TenantTransactionRunner`), Postgres (`SELECT … FOR UPDATE` + `WHERE sealed_at IS NULL`, idempotent), and DB trigger `fn_prevent_sealed_scan_run_mutation`.
- **Migration 0129 fail-closed:** tenant_id backfilled from engagements, `RAISE`s on any orphan before `SET NOT NULL`; enum CHECK constraints present; no data-loss on Up; Down drops policy + disables RLS correctly.

## Required before merge

**1. `internal/usecase/scanrun/service.go` (~L53,110-119,130,217-227,242) · medium · golden rule 6 (attributable + append-only audit)**
- `Actor` is `json:"actor,omitempty"` and never validated non-empty — a `CreateNativeScanRun`/`SealScanRun` can be recorded with an **empty actor**, i.e. a state change on immutable evidence with no attributable human/agent id. Reject blank `Actor` with `ErrValidation` in both.
- `recordAudit` is best-effort (`_ = s.audit.Record`) and runs **outside** the sealing transaction. A failure between commit and audit yields a **sealed, immutable run with no audit entry** — seal and its audit aren't atomic. Write the audit inside the same `TenantTransactionRunner` tx, or fail the op on audit-write error.

**2. `internal/usecase/scanrun/service.go:91-95` · low · rule 2**
The engagement∈tenant check (`GetByIDInTenant`) is guarded by `if s.engagements != nil`, so a nil-engagements wiring (memory/tests) silently skips the invariant. Prod is still safe (Postgres composite FK), but make `engagements` a required non-nil dep in `NewService` (or reject at write time when absent) so the invariant isn't wiring-dependent.

**3. `internal/usecase/scanrun/service.go:186-208` · low · rule 3/6**
The conflict audit path records raw `err.Error()` into audit metadata — a latent leak vector if the error chain ever carries a `result_ref`/DSN/internal path. Record a stable reason code (e.g. `"seal_conflict"`) instead.

## Note — overlap with #811
#811 (issue 755) also ships a `scan_run_provenance` migration and touches the same `scan_run_repo.go`/`ports.go`. These two implementations collide. This PR (#779, issue 708 A2a) is the focused, reviewable one — I'd land this after the fixes above, and have #811 drop/reconcile its scan-run piece. Flagging so the duplication is resolved deliberately, not by merge order.

Once 1–3 are addressed I'm happy to re-review and merge. Nice work on the RLS + seal-once + content-addressed hashing.

nghiadaulau (changes requested):
Re-reviewed head `3bf8ca2` (merge base `a681f0a0`) with build, vet, CGO-disabled build, unit tests, and Postgres integration tests against a fresh Postgres 17. Changes requested. Two new blockers, plus the three items from my 2026-09-03 review are still open (no commit since 2026-08-31).

## Blockers

**1. `make test` is red on this head.** `TestCodeQualityFindingsPersistWithScan` (`internal/usecase/sca/service_test.go:324`, present at the merge base) fails on `3bf8ca2` and passes on `main`:

```
--- FAIL: TestCodeQualityFindingsPersistWithScan (0.00s)
    service_test.go:367: scan history = [{ID:scan-job-1 ...}]   # expected 2 rows, got 1
```

Cause: `memory/scan_run_repo.go:64-65` changed the legacy `Save` from append to a map overwrite keyed by `(tenant, id)`, while `postgres/scan_run_repo.go:54` uses `ON CONFLICT (tenant_id, id) DO NOTHING`. The two twins now disagree on duplicate-ID semantics (memory last-write-wins, Postgres first-write-wins), and memory `Save` does not consult `IsSealed()` while `SaveScanRun` does. Pick one contract (reject with `ErrConflict` when the row is sealed, otherwise first-write-wins to match Postgres), enforce it in both stores, and give the sca test fixture distinct IDs.

**2. Migration 0129 fails on the production migration path.** Reproduced: fresh database migrated to 0128 with a `NOSUPERUSER NOBYPASSRLS` owner role, one engagement and one legacy `scan_runs` row inserted, then `goose UpTo 129` as that role:

```
ERROR: migration 0129: found 1 orphaned scan_runs rows with no matching engagement tenant_id (SQLSTATE P0001)
```

`engagements` has `FORCE ROW LEVEL SECURITY` since `0066:133`. The backfill `UPDATE scan_runs ... FROM engagements` (`0129:15-19`) runs with `app.current_tenant` unset, so it sees zero engagement rows, matches nothing, and the orphan check raises. `migration_0082_finding_promotions.sql:15-45` documents this exact hazard and lifts `FORCE` inside the DO block, restoring it at the end. The bundled `migration_0129_test.go` passes only because it runs as a superuser, which bypasses RLS. `docs/guide/deployment.md:206` says production runs `synapse-migrate` with a separate owner credential, so every deployment with scan history is blocked at the migrate step. Fix: copy the 0082 shape (`NO FORCE` on `engagements` inside the same DO block, backfill, orphan check, `FORCE` restored), and add a migration test that runs 0129 as a non-superuser owner with pre-existing rows.

**3. Open items from 2026-09-03**, unchanged in `internal/usecase/scanrun/service.go`:
- `Actor` (`:53`, `:130`) is never validated; empty or arbitrary actors are recorded on immutable evidence.
- `recordAudit` (`:242-247`) is `_ = s.audit.Record(...)` and runs after `tx.Run` committed. A seal can commit with no audit row, and an idempotent re-seal emits a second `provenance_sealed` entry. `ports.IdempotentAuditLogger.RecordOnce` (used in `usecase/promotion/service.go:253`) is the existing pattern.
- `engagements` is nil-guarded (`:91`); `NewService` accepts nil `engagements`, `tx`, `ids`, `audit`, against the "constructors validate" convention.
- Raw `err.Error()` goes into hash-chained audit metadata (`:192`, `:205`). `mapScanRunSQLError` (`postgres/scan_run_repo.go:641-646`) embeds `pgErr.Message` and passes unmapped errors through, including pgx connect errors that carry `host=... user=... database=...`. Record a classified reason code.

## High

**4. 0129 is not expand-only for the migrate-first window.** It sets `tenant_id NOT NULL` (`:34`) and forces RLS (`:195`) in one step. A pre-PR binary still inserts `scan_runs` without `tenant_id` and reads outside `WithTenant`, so during the documented migrate-then-deploy overlap every scan's run history insert fails on NOT NULL and the error is discarded (`_ = s.runs.Save(...)`, `sca/service.go:2239`, `:3157`), and `GET .../scan-runs` returns empty. Split expand (nullable column, backfill, child tables, trigger) from contract (NOT NULL, composite FK, RLS) one release later, or add a transitional `BEFORE INSERT` trigger deriving `tenant_id` from `engagements`, and stop discarding the `Save` error.

**5. `SaveScanRun` upsert can forge a sealed header.** `postgres/scan_run_repo.go:141-157`: `DO UPDATE SET` includes `sealed_at`, `manifest_hash`, `terminal_status`, so any port caller can flip an unsealed run to sealed with an arbitrary hash and no lanes, bypassing `SealScanRun`'s `FOR UPDATE`, lane insertion, and audit. The DB trigger does not protect this transition because the row is unsealed before the write. On conflict the statement also drops `engagement_id` and `provenance` from the SET list and reports success. Make `SaveScanRun` create-only: reject `SealedAt != nil || ManifestHash != "" || len(Lanes) > 0`, use `ON CONFLICT DO NOTHING`, return `ErrConflict` on zero rows.

**6. `ComputeManifestHash` is not deterministic over time values, and this already breaks the stacked consumer.** `domain/scanrun/model.go:127-133` hashes `LaneStage.StartedAt/FinishedAt` as `time.Time` through `json.Marshal` (`:366`). Probe: the same instant in UTC vs `+07:00` yields different hashes, and nanosecond vs microsecond precision yields different hashes. Postgres stores `timestamptz` at microseconds, so the stored `manifest_hash` cannot be re-derived from stored rows. The repo test passes only because inputs are pre-truncated (`scan_run_repo_test.go:425`) and `ComputeRunManifestHash` reuses `lane.ManifestHash` when non-empty (`model.go:390-393`). On #814/#815, `domain/assessmentsnapshot/snapshot.go:291` re-verifies the lane hash over lanes reloaded from Postgres, and every native-run finalization fails with `validation error: lane manifest hash mismatch` (10 Postgres tests in #815 and the snapshot repository test in #814 fail this way on a fresh database). Canonicalize inside the envelope (UTC, truncate to microseconds, fixed RFC 3339 layout) and add a zone and precision invariance test.

## Medium

**7. `DefaultTenant` fallback inverts the fail-closed tenant rule.** `postgres/scan_run_repo.go:42-45, 65-68, 100-103` and the memory twin substitute `DefaultTenant` when the context has no tenant. `tenant.go:44-53` states a missing tenant is deny; no other persistence store does this (`grep 'tenantID = shared.DefaultTenant'` has no other hits). A worker-originated scan with no bound tenant writes under `default`. Return `ErrValidation` and make callers bind the tenant.

**8. Sealed runs make engagement deletion a 500.** `fk_scan_runs_engagement ... ON DELETE CASCADE` (`0129:46-47`) fires the BEFORE DELETE trigger on sealed rows, which raises P0001 that no mapper translates (`postgres/engagement_repo.go:149`). Blocking the delete is right; map it to `ErrConflict` with a user-facing message, or declare `ON DELETE RESTRICT` in the schema.

**9. Port widening instead of a new port.** `ports.ScanRunStore` (`ports.go:623-632`) now has seven methods; `sca.Service` depends on four it never calls and `scanrun.Service` on three it never calls. Add `ports.ScanRunProvenanceStore` with the four new methods, satisfied by the same structs. `SealScanRun` takes 8 positional parameters that duplicate aggregate state; pass a command struct.

**10. Target identity canonicalization gaps** (`domain/scanrun/target_identity.go`): no length bounds on identities, lane keys, producers, refs; `https://github.com/org/repo%2F..%2Fother` percent-decodes then `path.Clean`s into `org/other` (use `EscapedPath`); `CanonicalizeCloudResourceTarget("aws","123/456","x")` and `("aws","123","456/x")` collide; OCI accepts an empty repository; `manifest_hash` and `result_sha256` are not validated as hex. Identity collisions let two targets share one coverage claim.

**11. Store-level seal accepts zero lanes and an unverified hash.** `postgres/scan_run_repo.go:313-326` and `memory/scan_run_repo.go:179-188`: `SealScanRun(ctx, t, r, StatusSucceeded, nil, 1, "", now)` returns nil and marks the run sealed with no lanes and an empty hash. The store never checks `manifestHash == ComputeRunManifestHash(lanes)`, and the idempotency check at `:348-353` then compares that unverified value. Reject zero lanes and empty hash in both stores, and recompute the hash inside `SealScanRun`.

**12. `ssh://git@host/org/repo` is rejected while `git@host:org/repo` is accepted** (`target_identity.go:112-125` vs `:132-134`; `target_identity_test.go:76-80` locks the rejection in). Same repository, two outcomes by spelling, and SSH remotes emitted as URLs cannot be sealed. Reject only when `u.User.Password()` is set, or reject both forms.

## Low

- `usecase/scanrun.Service` is not wired by any `cmd/*` or route. Name the consumer in the PR body or wire `CreateNativeScanRun`/`SealScanRun` into the scan path here.
- `target_identity.go:144-149` and `:282-317` re-implement `engagement.NormalizeDomain` / `NormalizeURL`; a future IDNA rule change will make scope identity and provenance identity diverge.
- `ListScanRuns` issues `1 + N + 2·N·L` queries (`scan_run_repo.go:297-303, 535-547`); load lanes, versions, and stages with `WHERE scan_run_id = ANY($1)` and group in Go.
- Down migration (`0129:201-228`) drops sealed provenance silently: with a sealed native run present, `DownTo 128` then `UpTo 129` leaves the run as `provenance=legacy, sealed=false, lanes=0`. Refuse Down when `EXISTS (SELECT 1 FROM scan_runs WHERE sealed_at IS NOT NULL)`.
- `migration_0129_test.go:128-182` discards the `Scan` errors (`_ = pool.QueryRow(...).Scan(&exists)`), so a failing query yields `exists=false` and the down assertion passes trivially. It checks table existence only, not columns, triggers, or policies.
- `Lane.EngagementID` is required by `Lane.Validate` (`model.go:176-178`) but overwritten by both stores with the run's engagement; a mismatching value is accepted silently.
- `IsCompleteCoverage` doc (`model.go:264`) says terminal stages are checked; a `succeeded` lane with a single `failed` stage returns true. Fix the code or the comment.
- PR body still references migration 0127; no `CHANGELOG.md` Unreleased entry for a schema and RLS change.

## Verified clean

`go build ./...`, `go vet ./...`, `CGO_ENABLED=0 go build ./cmd/...` pass. With `SYNAPSE_TEST_DB_DSN` set against a fresh Postgres 17, the PR-added tests pass (`TestPostgresScanRunStore_*`, `TestMigration0129*`, `TestMigrationInventoryNoDuplicateGooseVersions`, `TestService_PostgresParity`) and migrations 0001 to 0129 apply from empty as superuser. The hostile harness passes. Architecture test passes; `domain/scanrun` imports domain, stdlib, and `x/net/idna` only. RLS enabled and forced on all four tables, composite PKs and FKs on `tenant_id`, explicit `tenant_id = $n` in every query under `WithTenant`; `SealScanRun` derives `engagement_id` from the locked row; cross-tenant seal/get/list resolve to `ErrNotFound`. Trigger coverage is UPDATE and DELETE on the header and INSERT, UPDATE, DELETE on all three child tables with both OLD and NEW parents checked. Seal-once holds at service, `FOR UPDATE` plus `WHERE sealed_at IS NULL`, and trigger. All SQL is parameterized. `CompareRuns` moved from global-by-id to tenant-scoped, which closes a pre-existing cross-tenant read.

Once 1 through 6 are fixed and the 2026-09-03 items are closed, re-request review.

VietSory:
i have fixed all blocks.

nghiadaulau:
Để mở, chưa merge (theo quyết định của maintainer). Ghi lại ngữ cảnh cho lần rebase sau:

1. Reachability: đây là Phase A2a expand-only — có internal/domain/scanrun + ScanRunProvenanceStore + SealScanRunCommand service, nhưng chưa có HTTP route/dashboard surface nào chạm tới. Theo quy tắc của repo (một domain+store chưa reachable không land lên main), nên chờ phase wire route/UI trước khi merge.
2. Migration collision: migrations/0134_scan_run_provenance.sql đụng số với 0134_dast_runs.sql đã ship trên main (duplicate goose version = panic khởi động). Main hiện đã ở 0137 (0137_integration_framework từ #813), nên khi rebase cần renumber migration này lên 0138, và sửa migration_0134_test.go (đang hardcode UpTo/DownTo 133/134) sang 137/138.
3. Ngoài ra sẽ có text-conflict nhỏ ở CHANGELOG.md, internal/usecase/ports/ports.go (khối ScanRun), internal/usecase/sca/service.go, internal/infrastructure/persistence/postgres/engagement_repo.go — đều resolvable.

Bản thân feature là sound và additive (giữ nguyên legacy ScanRunStore); chỉ vướng reachability + renumber.

VietSory:
Rebased onto current `main` (`1c1c36a`); this head is mergeable.

All items from the strict reviews are addressed:

- Migration collision: provenance is now `0138_scan_run_provenance.sql` with an inventory guard; the branch has been rebased onto current main.
- Production migration path: a `NOSUPERUSER NOBYPASSRLS` owner test verifies legacy backfill. The migration temporarily lifts FORCE RLS only inside its transaction, backfills/fails closed, validates the legacy FK in that owner context, then restores FORCE RLS.
- Rolling deployment safety: this is expand-only. `scan_runs.tenant_id` remains nullable; a bridge plus SECURITY DEFINER trigger derives ownership for legacy writers; scan_runs RLS/NOT NULL/composite contract are deferred.
- Deterministic provenance: lane identity/status and canonical UTC/microsecond timestamps are committed; run hashes are computed in the domain from lane-key-sorted lane hashes. Tests cover order, zone and precision invariance.
- Store/service hardening: missing tenant fails closed; `SaveScanRun` is create-only and rejects forged sealed state; both stores recompute and verify seals, reject zero lanes, enforce terminal timestamps, and use typed seal commands.
- Audit and constructor contracts: actor/dependencies are required; audit `RecordOnce` is in the tenant transaction; only classified reason codes reach immutable audit metadata.
- API reachability: the composed scanrun service now backs existing `GET /engagements/{id}/scan-runs`; the dashboard shows native/legacy provenance, status, seal, lane count and complete coverage while retaining legacy fields.
- Delete behavior is explicit `RESTRICT` and mapped to `ErrConflict`.
- Test isolation: scan-run Postgres tests now use per-test databases, so sealed provenance cannot contaminate parallel migration rollback tests.

Validation on the final rebased SHA:
- `go build ./...`
- `go vet ./...`
- `go test ./...`
- targeted PostgreSQL 17 owner/migration/store/service tests
- frontend test, typecheck and build

All passed in the final staging run:
https://github.com/VietSory/synapse-ce/actions/runs/34264575450

The PR body still contains stale pre-rebase migration numbers; the current source of truth is migration 0138 and this comment. Please re-review this head. @nghiadaulau

nghiadaulau (approved):
Re-reviewed (CI disabled) — **all three findings from my earlier review are addressed. Approving.**

Verified locally on the updated branch: `go build ./...` + `go vet` clean; `domain/scanrun`, `usecase/scanrun`, `persistence/memory`, `persistence/postgres` tests all green.

Fixes confirmed:
1. **(was medium, rule 6)** `Actor` is now validated non-empty in both `CreateNativeScanRun` and `SealScanRun` (`service.go` L64/L141 → `ErrValidation`). And the audit is now the **final statement inside the sealing transaction** (`recordAudit(txCtx, …)`, L202) and **returns its error**, so an audit-write failure rolls the seal back — seal + audit are atomic. Uses `RecordOnce` (idempotency key) for exactly-once semantics. ✅
2. **(low, rule 2)** `engagements` is now a required non-nil dependency in `NewService` (L34), so the engagement∈tenant invariant is no longer wiring-dependent. ✅
3. **(low, rule 3/6)** the conflict audit records a stable `"reason": "seal_conflict"` instead of the raw error string (L224). ✅

Also confirmed: migration renumbered to 0138 (main is at 0137 — no collision), and the only merge conflict was a CHANGELOG union which I resolved on the branch. RLS+FORCE, composite FKs, seal-once (service + `FOR UPDATE`/`sealed_at IS NULL` + DB trigger), and deterministic content-addressed manifest hashing all still hold. Nice, thorough turnaround. LGTM.

PR #811, opened 2026-09-02
feat: deliver integration framework and assessment lifecycle rollout

nnatuan03:
> [!NOTE]
> **Ready for human review.**
>
> Revert PR #810 is merged, all issues closed prematurely by #809 have been reopened, and the implementation has been reapplied and revalidated on current `main`. Please do not auto-merge; normal review is required.

## Recovery context

- Replaces the prematurely merged PR #809.
- Reapplies the original implementation without code changes; the resulting tree matches #809 exactly.
- Places the feature commit on top of current `main` after revert PR #810.
- All issues listed under “Issues closed on merge” are open again.
- Ready for human review; do not merge automatically.

---

## Summary

This PR delivers two completed workstreams:

1. the provider-neutral Integration Framework with Jenkins as the first CI/CD adapter (`#755`), and
2. the executable Assessment lifecycle epic (`#694`) through tenant canary/read/UI rollout and rollback operations.

The implementation is additive and fail-closed. Legacy reads remain unchanged until a tenant is explicitly present in the shadow, read, and UI allowlists.

## Integration Framework / Jenkins

- Adds tenant-scoped Integration, Binding, Operation, ExternalRun, and write-only encrypted credential models.
- Adds a code-owned provider registry and narrow capability interfaces; core orchestration contains no Jenkins/provider switch.
- Adds Jenkins connection test, nested pipeline discovery, recent build polling, outcome normalization, exact-commit analysis correlation, and durable worker scheduling.
- Adds admin/read RBAC, tenant RLS, optimistic concurrency, idempotency, audit, redaction, SSRF-safe HTTP behavior, bounded payloads, and non-secret API projections.
- Adds OpenAPI coverage, Settings UI, deployment smoke fixtures, Helm/runtime configuration, and operator documentation.

## Assessment lifecycle domain and persistence

- Adds immutable Assessment Snapshots, versioned Finding Identity/Observation lineage, ancestry-aware lifecycle/neutral Comparisons, review candidates, and deterministic content hashes.
- Adds tenant-scoped Cycle APIs, atomic initial/re-test dual-write, frozen boundary inheritance, independent authorization requirements, relationship preview/commit, selected-head semantics, grouped list/detail projections, and RBAC-aware web flows.
- Adds immutable closure manifests, reopen supersession, deterministic final reports, closure history/download UI, and reference-integrity enforcement.
[...]

PR #813, opened 2026-09-04
feat: add secure CI/CD integrations

nnatuan03:
Split from #811 in response to the structural and security review.

Scope
- Provider-neutral integration domain, registry, API, persistence, scheduler, worker profile, and UI.
- Jenkins adapter with HTTPS-only transport, redirect rejection, dial-time address validation, bounded discovery, and credential-reflection rejection.
- Write-only, AAD-bound encrypted credentials and tenant-scoped RLS persistence.
- Migration 0130; merge after #779 so scan-run provenance remains migration 0129.

Review fixes included
- Material connection changes atomically disable the integration, invalidate pending work, and advance connection or credential revisions.
- Enablement requires a successful test for the exact immutable connection and credential revisions.
- Operation, durable job, and audit entry are committed atomically.
- Cancellation and archive invalidate claims and fence provider result publication.
- In-memory composition installs a run lock.
- Jenkins output URLs are reconstructed safely and total discovery work is bounded.

Validation
- Focused Integration, Jenkins, safe-HTTP, persistence, HTTP, config, and frontend tests pass.
- Go vet/race checks for the focused packages pass.
- Frontend typecheck and tests pass.

This PR contains no ScanRun provenance implementation and does not overlap #779.


nnatuan03:
@pho-veteran @nghiadaulau — review requested.\n\nSuggested review path:\n1. Review this PR independently against main.\n2. Focus on the former #811 Integration findings: material-update disable/invalidation, exact connection and credential revision test binding, atomic operation/job/audit writes, cancellation/archive publication fencing, in-memory run locking, Jenkins output sanitization, discovery budget enforcement, and HTTPS/dial-time SSRF checks.\n3. #779 should merge before this PR so migration 0129 precedes migration 0130. No ScanRun provenance code is included in this slice.\n\nThe PR body contains the focused validation results. Please keep findings specific to this slice; the downstream Assessment slices are #814–#816. Reviews can run concurrently, while merges must follow the documented order.

pho-veteran (changes requested):
Security, architecture, and correctness review against exact base `a5880d2d1d1d04e778d6725dc943c082981a0012` and head `66eee8315ee4ad52380b78fca94f5520a80cc5f0`.

1. **[P1] Credential replacement/deletion is not bound to the integration revision.** `SetCredential` and `DeleteCredential` first read the integration, but the credential store methods accept no expected version/revision (`internal/usecase/integrations/service.go:178-212`, `internal/usecase/ports/integration.go:19-20`, `internal/infrastructure/persistence/postgres/integration_store.go:183-245`). A stale PUT can validate for the old connection, race an endpoint update, and then attach that credential to the new endpoint; a subsequent test can disclose the old token to the unintended origin. A stale DELETE can likewise remove a newer replacement. Carry a caller-visible version/connection revision and compare it under the same row lock as the credential mutation.

2. **[P1] Poll admission has a TOCTOU that adopts an untested revision.** `StartOperation` checks `Enabled` on the pre-transaction object (`internal/usecase/integrations/service.go:296-332`), while both stores replace the operation revisions with whatever is current at persistence time without rechecking poll eligibility (`internal/infrastructure/persistence/postgres/integration_store.go:329-360`, `internal/infrastructure/persistence/memory/integration_store.go:373-410`). If a credential/config change disables the integration between those steps, the queued poll is stamped with the new revision. `HandleJob` checks only archived/revision state before resolving the new credential and contacting the provider (`internal/usecase/integrations/service.go:431-452`); the later materialization fence is too late. Persist the revisions that were admitted and compare them under lock, and atomically require `enabled` for poll creation.

3. **[P1] Most administrative mutations can commit without their audit record.** Create/update/credential/enable/archive/binding/cancellation mutate the store and only afterward call the separate `AuditLogger` (`internal/usecase/integrations/service.go:103-107`, `168-173`, `196-212`, `231-243`, `255-258`, `275-293`, `350-357`). An audit failure returns an API error but leaves the state change committed and unaudited. Move each security-relevant mutation and `appendTenantAudit` into one store transaction, as this PR already does for operation creation.

4. **[P1] Poll results can survive a successful cancellation/archive.** `executePoll` commits external runs and only then finalizes the operation in a second transaction (`internal/usecase/integrations/service.go:583-594`). Cancellation or archive can land between those calls, mark the still-running operation cancelled, and let `FinishIntegrationOperation` return that terminal row without removing the already-published runs. This contradicts the claimed cancellation publication fence. Upsert the runs and terminal operation state atomically under the same integration/revision/operation fence.

5. **[P1] Percent-encoded credentials bypass Jenkins response reflection checks.** `safeRunURL` checks the raw provider string, parses it, then reconstructs and returns the decoded path without checking the canonical result (`internal/infrastructure/integration/jenkins/client.go:487-517`, `520-530`). A provider URL containing `%74%6f%6b%65%6e` bypasses the raw `token` marker and is persisted as `/token`; this was reproduced with a focused probe. The URL is then returned by the external-runs API and rendered as a link. Reject credential markers after canonicalization too, or synthesize run URLs only from validated pipeline/build/queue identifiers.

6. **[P2] Poll resource limits are per binding rather than per operation.** The loop at `internal/usecase/integrations/service.go:521-579` performs queue/build calls and accepts up to `MaxRunsPerPoll` for every binding, with no aggregate binding/request/run/byte/deadline budget. Since bindings can be created independently, one operation can occupy a worker for thousands of network calls and accumulate a correspondingly large write set. Add operation-wide limits and cancellation/deadline enforcement.

7. **[P2] `UpdateIntegrationRequest` is unsatisfiable.** `api/openapi.yaml:2819-2826` combines `CreateIntegrationRequest`, which requires `provider` and forbids unknown properties, with a schema that sets `provider: false` and requires `version`. No payload can satisfy both branches, so generated validators/clients cannot perform updates. Define a standalone update schema with the runtime fields plus `version` and without `provider`.

8. **[P2] Detail request races can apply actions to the wrong selected integration.** `loadDetail` unconditionally publishes operations/bindings/runs from whichever request finishes last (`web/src/pages/Settings/Integrations.tsx:60-79`, `81-109`). After rapidly selecting A then B, a late A response is displayed under B; cancellation can target A, and a pipeline discovered for A can be submitted as a binding for B. Abort stale requests or gate all state publication on a request generation/selected ID.

9. **[P2] Mutation responses falsely report retained credentials as missing.** Credential presence is hydrated only by Get/List, while Update and enable/disable return the direct store object (`internal/usecase/integrations/service.go:118-135`, `168-175`, `231-245`). Those objects have `CredentialConfigured == false`, so the UI disables test/discovery and hides credential deletion immediately after otherwise successful mutations. Hydrate presence before returning or make it part of the atomic store response.

10. **[P2] Archiving neither removes credentials nor prevents later credential writes.** Archive permanently disables/hides the integration but leaves its encrypted credential row (`internal/infrastructure/persistence/postgres/integration_store.go:157-180`), and both credential mutation paths lock archived rows without rejecting them (`internal/infrastructure/persistence/postgres/integration_store.go:183-245`; memory parity at `internal/infrastructure/persistence/memory/integration_store.go:224-280`). With no unarchive path, this retains secrets unnecessarily and permits mutation of an archived aggregate. Delete credentials during archive and reject later credential mutation.

Validation performed: focused Go tests with `-race` passed for integrations, Jenkins, memory persistence, and safe HTTP; focused `go vet` passed. The percent-encoded reflection probe failed as described and was removed afterward. The pre-existing Fleet OpenAPI route-coverage debt was not attributed to this PR.

nnatuan03:
@pho-veteran @nghiadaulau — all 10 findings from the requested-changes review are addressed in commit `6f9729f`.

Review guide:

1. **Credential revision fencing:** credential PUT/DELETE now require caller-visible `version` and `connection_revision`, then compare both under the integration row lock before mutating the secret.
2. **Poll admission TOCTOU:** operation creation persists the revisions admitted by the service and compares them under lock; poll admission also atomically requires the integration to remain enabled. Workers recheck enabled/archive/revision state before resolving credentials.
3. **Transactional audit:** create/update/credential/enable/archive/binding/cancel mutations now append their audit entry in the same store transaction. Audit failures roll the mutation back.
4. **Atomic poll publication:** run upserts and terminal operation state now commit through `FinishIntegrationPoll` under the same integration/revision/operation fence. Cancelled or archived work cannot publish late runs.
5. **Jenkins URL sanitization:** `safeRunURL` now rejects credential markers after URL canonicalization as well, including percent-encoded `token` paths.
6. **Operation-wide poll budget:** one poll is capped at 100 bindings, 200 aggregate runs, 250 provider requests, 32 MiB of responses, and a 2-minute deadline. The budget is shared across all bindings.
7. **OpenAPI update schema:** `UpdateIntegrationRequest` is now standalone and satisfiable, with runtime update fields plus `version` and no `provider`.
8. **UI detail race:** detail publication is gated by selected integration ID and request generation, so late responses for A cannot overwrite B or drive actions against the wrong integration.
9. **Credential presence after mutations:** update and enable/disable responses now hydrate `credential_configured` before returning.
10. **Archive secret lifecycle:** archive deletes the credential and cancels active operations transactionally; later credential writes to archived integrations are rejected.

Suggested re-review focus:

- Store transaction boundaries and optimistic fences in the Postgres and memory implementations.
- `FinishIntegrationPoll` as the sole poll publication path.
- Jenkins canonical URL checks and shared operation budget accounting.
- OpenAPI credential concurrency fields and the UI stale-request guard.

Validation completed:

- Focused Go tests pass for integrations, memory persistence, Jenkins, HTTP API, and Postgres.
- Race-enabled tests pass for integrations, memory persistence, and Jenkins.
- Focused `go vet` passes, and `go test ./... -run '^$'` compiles the full repository.
- OpenAPI strict validation, dead-operation, and debt tests pass.
- Web build passes; all 391 web tests pass.

Known unrelated base failures remain: the full API route-coverage test reports 15 existing Fleet routes, and web typecheck reports the existing `Grade` export issue in `OverviewMetricCard.tsx`.

Please re-review the replacement commit, especially the four areas listed above.

nghiadaulau (changes requested):
Reviewed head `66eee83` against `main` (`a5880d2`) with build, vet, CGO-disabled build, unit tests, the Postgres suite on a fresh Postgres 17, the drift guards, and the web typecheck and tests. Changes requested. The 10 items from pho-veteran's review remain open (no commit since 2026-09-04 04:19); items 1, 3, 4, 5, 6, and 10 were re-verified on this head. The findings below are new; each was reproduced on this head.

## Blockers

**1. The Postgres scheduler query is a SQL type error, so no poll is ever dispatched in production.** `postgres/integration_store.go:480-481`: `COALESCE(...)::timestamptz <= $1 - make_interval(secs=>poll_interval_seconds)`. `$1` has no declared type, Postgres resolves `$1 - interval` as `interval - interval`, and there is no `timestamptz <= interval` operator. Reproduced on a fresh database: `ERROR: operator does not exist: timestamp with time zone <= interval (SQLSTATE 42883)`. Every `Scheduler.Tick` fails and logs `integration scheduler tick failed`; only manual "Poll now" works. The scheduler test uses the memory store, so nothing catches it. Fix: `$1::timestamptz - make_interval(...)` plus a Postgres test that seeds an enabled integration and asserts it is returned.

**2. The PR's own Postgres test fails and pollutes the suite, so CI goes red.** `integration_store_test.go:178-181` expects a cancelled operation's job to still be claimable, but `CancelIntegrationOperation` (`integration_store.go:458-459`) sets `jobs.status='done'`, so `Claim` returns nil: `integration_store_test.go:180: cancelled claim=<nil> err=<nil>`. The test and the implementation contradict each other. Separately, `StartIntegrationOperation` appends `audit_log` rows through `appendTenantAudit` and the cleanup at `:78-90` deletes every integration table, jobs, projects, and tenants but not `audit_log`; migration 0085's Down then raises `cannot roll back audit_log tenant chains after tenant genesis rows exist` in `TestMigration0044/0056/0058/0061/0066/0068`. On a fresh database the PR adds 7 failures over main's 4 pre-existing ones; with this one test skipped, exactly main's 4 remain. Assert that `Claim` returns nil after cancel and delete the tenant's `audit_log` rows in cleanup.

**3. The credential-reflection check rejects benign responses whenever the username is a substring of any URL, job name, class, or branch, while a percent-encoded token passes.** `jenkins/client.go:520-531` matches `adapter.username` as a substring against `job.URL` (which contains the host, `:141`, `:459`) and `branch.name` (`:268`). Reproduced: username `jenkins` with endpoint `https://jenkins.example.com` rejects every discovery; username `dev` rejects branch `develop`. The connection test has no reflection check, so the user is told the integration works, then every discovery and poll fail with `provider operation failed`. In the other direction (pho-veteran's item 5, re-verified), `safeRunURL` (`:487-517`) checks the raw string, then `url.Parse` decodes the path and `resourceURL` re-serialises it, so `https://ci.example.com/job/%61%62%63/1/` with token `abc` is stored as `https://ci.example.com/job/abc/1` and rendered as a link in `RunsCard`. Match only the token and the Basic-auth encodings, on the decoded path and the final output string.

**4. Migration numbering depends on an unenforced merge order.** 0130 with no 0129 applies on a fresh database today, but `postgres.Migrate` (`db.go:239`) calls `goose.UpContext` without `WithAllowMissing`. If this PR merges and deploys first and #779 lands `0129_*.sql` afterwards, goose v3.27.3 returns `found 1 missing migrations before current version 130` and API and worker refuse to start. Renumber at merge time or add a contiguity guard test.

## High

**5. No wall-clock budget per operation; a slow provider pins worker loops for hours.** `usecase/integrations/service.go:403-456` (`HandleJob` has no deadline), `jenkins/client.go:27-31, 104-119` (`maxDiscoveryNodes` = 2000 requests, each with a 20 s client timeout), `client.go:176-184` (2 requests per binding per poll), `service.go:261-282` (no cap on bindings per integration), `usecase/worker/worker.go:237-256` (heartbeat extends the claim while the handler runs). A tenant admin points an integration at a host they control that answers each `/api/json` after 19 s with one more folder: discovery runs 2000 requests, about 10.5 h of one worker loop. The unique index is on the full endpoint string (`0130:26-27`), so `https://evil.example/a`, `/b`, ... are distinct integrations; with `SYNAPSE_WORKER_CONCURRENCY` of them the tenant holds every loop of an `all`-profile worker and recon and SCA jobs stop being claimed. Cancel only flips the row; the handler checks state at Finish. Add a per-type deadline in `HandleJob` (test 30 s, discover and poll 5 min) under the worker visibility, a per-operation request budget in the adapter, a cap on bindings per integration, one `/queue/api/json` per poll instead of per binding, and cancellation checks between bindings and folders.

**6. Every operation leaks its idle keep-alive connection and three goroutines.** `safehttp/client.go:26-62` creates a fresh `http.Transport` with `IdleConnTimeout` 0, no `MaxIdleConns`, never closed; `client.go:77` creates one client per adapter, one adapter per `HandleJob`. Measured with a probe: 25 dropped clients kept 25 server-side connections and 75 goroutines after GC. A Jenkins that never closes idle connections adds one TLS connection per poll (minimum every 30 s per integration) until the worker restarts; 20 integrations is about 57k connections per day. Set `IdleConnTimeout`, `MaxIdleConnsPerHost`, `MaxConnsPerHost` in `safehttp`, and add a `Close()` on the adapter that calls `CloseIdleConnections()`, deferred in `HandleJob`.

**7. Tenant admins toggle control-plane egress into private ranges with no operator gate.** `integration_handler.go:51, 95` accepts `allow_private_network` from the request body with PermAdminister; `safehttp/client.go:65-71` waives `IsPrivate` when set. In a multi-tenant deployment a tenant admin gets a port-liveness and TLS-presence oracle on the operator's private ranges through the operation timeline, and any internal HTTPS service with a matching certificate receives Basic credentials chosen by the tenant. The audit metadata for `integration.created` and `integration.updated` carries only `provider`, not `endpoint` or `allow_private_network`. Add an operator env gate that Create and Update reject when the tenant flag is set and the gate is off, record endpoint and the flag in audit metadata, and block 2002::/16 and 64:ff9b::/96 in `blocked()`.

## Medium

**8. Archive retains ciphertext, and archived integrations still accept new credentials.** `postgres/integration_store.go:157-181` never touches `integration_credentials`; `:183-217` locks the row without checking `archived`; memory twin at `:191-214, 224-253`. Extends pho-veteran's item 10 with the memory parity gap.

**9. The provider-neutral domain hardcodes the Jenkins URL shape.** `domain/integration/model.go:324-334` (`CanonicalExternalKey`) requires `/job/`; `Binding.Normalize` (`:348`) and `ExternalRun.Normalize` (`:491`) both call it, so every binding and run must carry a Jenkins path. The next adapter (GitLab `group/project`, GitHub Actions `owner/repo/.github/workflows/x.yml`) cannot create a binding without a fake prefix, and the claim in `docs/guide/integrations.md` that another adapter does not change domain rules is false today. The Jenkins adapter already re-validates its keys (`client.go:172, 476`); keep the neutral checks in the domain and move the prefix to the adapter or a `ProviderDescriptor` validator.

**10. The scheduler enumerates tenants through `ports.VulnerabilityReconciliationTenantStore`.** `scheduler.go:30, 39, 76` and `cmd/synapse-worker/main.go:199`; that port (`ports.go:650`) is named for and owned by another module and satisfied by `postgres.NewEngagementRepository`. `ports/detection_provenance.go:30` already duplicates the same method. Introduce a `ports.TenantLister`.

**11. The integration store reads another aggregate's JSON payload layout.** `postgres/integration_store.go:568-603` (`MatchIntegrationAnalysis`) queries `project_analyses` by `payload->>'source_commit'` and `payload#>>'{source_revision,head}'`. No port exposes "find analysis by commit"; if project analysis moves or renames the key, its tests stay green and integration correlation silently degrades to `missing`. Implement `ports.IntegrationAnalysisMatcher` in `project_analysis_store.go` next to the writer.

**12. `integration_external_runs.analysis_id` FK is not tenant-composite** (`0130:122`), unlike the other four FKs; `job_id REFERENCES jobs(id)` will also block any future jobs retention purge.

**13. The scheduler writes one append-only audit row per dispatched poll** (`scheduler.go:102` to `service.go:328-332`, actor `system:integration-scheduler`): 2880 rows per day per integration at the 30 s minimum interval, growth controlled by tenant configuration. Audit human-initiated operations; the `integration_operations` row is the durable record for scheduled polls.

**14. Postgres `UpdateIntegration` treats every update with a non-empty config as material.** `integration_store.go:102` compares `string(current.Config)` (JSONB text with Postgres spacing) to `json.Marshal` output; they differ, so a rename alone sets `enabled=false` and bumps `connection_revision` (probe: `after rename: connection_revision=2 enabled=false`). Unreachable with the shipped Jenkins provider (config is always `{}`), and the memory store compares canonical JSON, so the two stores disagree on the core invariant. Compare `config IS DISTINCT FROM $4::jsonb`.

**15. Provider failure causes are discarded.** `service.go:608` records `provider operation failed` for every permanent failure (401, 404, invalid JSON, cross-origin job URL, reflection rejection, oversized body) although the adapter already emits safe constant strings. With blocker 3 and finding 16 this makes the feature undebuggable from the product.

**16. Same-origin job URLs with an explicit `:443`, or a Jenkins root URL that differs from the configured endpoint, abort discovery.** `client.go:469, :505` compare `parsed.Host != adapter.base.Host`. Jenkins derives the `url` field from `JENKINS_URL`, not the request host, so TLS termination in front of Jenkins (`http://ci.internal:8080` behind `https://ci.internal:8443`) fails on the first job with `cross-origin job URL`, and `:443` does not compare equal to the default port. Not documented as a prerequisite in `docs/guide/integrations.md`.

**17. Disable during a running poll leaves the operation `running` until three attempts are exhausted.** `integration_store.go:136-155` does not invalidate on disable; `service.go:583` propagates `ErrConflict` from the upsert; the worker retries twice (each re-polls Jenkins), dead-letters, and the unique active index blocks new operations meanwhile. Archive and material updates cancel properly; disable is the odd one out.

**18. A job or branch whose name contains `..` (for example `v1..v2`) aborts the whole discovery.** `model.go:330` checks `strings.Contains(cleaned, "..")` after `path.Clean` already removed traversal; `client.go:131` turns any key failure into a permanent error. Skip the job instead.

## Low

- `domain/integration/registry.go:35-39, 86` puts the concrete `Registry` with a mutex in the domain and `service.go:50` depends on it; the existing convention (`ports.VulnerabilityProviderRegistry`, concrete in `usecase/vulnerabilitymonitor/registry.go`) is the reverse.
- `service.go:372-401` is the fifth copy of the lease-loss idiom (`recon`, `cspm`, `vulnerabilitymonitor`, `vulnerabilityreconciliation`); a shared helper is due.
- `cmd/synapse-worker/main.go:220` sets the integrations-profile visibility to `cfg.ReconTimeout+time.Minute`; nothing about a Jenkins poll relates to the recon timeout.
- `List` (`service.go:118-124`) does one `IntegrationCredentialConfigured` round trip per integration; return the flag from the list query.
- No `CHANGELOG.md` Unreleased entry for a Settings tab, 18 routes, and five `SYNAPSE_*` variables; no `migration_0130_test.go` (the store test calls `Migrate`, but the documented convention is not followed).
- The hostile harness (`harness_test.go`) does not enumerate the 18 new routes; tenant isolation for them rests on the handler test and RLS.

Claims in the PR body with no test that fails when the behavior breaks: "material connection changes atomically disable the integration" (no test calls `Service.Update` or `UpdateIntegration` in any package; only the credential-replacement half is tested at `service_test.go:293-302`); "scheduler" on Postgres (blocker 1); "redirect rejection" through the Jenkins `get` path (only `safehttp` tests it in isolation); "Jenkins output URLs are reconstructed safely" (only userinfo is tested, `client_test.go:150`); the negative path of "enablement requires a successful test" on Postgres.

## Verified clean

`go build ./...`, `go vet ./...`, `CGO_ENABLED=0 go build ./...`, and the architecture tests pass; `go test` on `domain/integration`, `usecase/integrations`, `jenkins`, `safehttp`, `platform/config`, `persistence/memory`, and `httpapi -run Integration` pass; the two touched vitest files pass (8 tests) and `pnpm build` passes. `pnpm run typecheck` fails on this head and on `main` with the same pre-existing error (`OverviewMetricCard.tsx(14,26): TS2459`), not caused by this PR. Drift guards fail only on the pre-existing fleet routes and `SYNAPSE_DATA_DELETION_ENABLED`; none of the new routes or variables is reported and `integrations.md` is in the mkdocs nav. Jenkins lives entirely in infrastructure and reuses `safehttp`; HTTPS-only with userinfo, query, and fragment rejection; `CheckRedirect` returns `ErrUseLastResponse` so Basic auth never leaves the configured origin; `Proxy: nil`; resolution and address validation happen inside `DialContext` on every dial; TLS verification uses the configured host; Jenkins request URLs are rebuilt through `resourceURL` with origin and `..` checks; body capped at 4 MiB plus 20 s timeout; provider errors are fixed text. Credentials are AES-256-GCM with a tenant-and-integration-bound AAD; an endpoint change deletes credentials and bumps both revisions in one transaction under `FOR UPDATE`; enable requires a succeeded `test` at the exact revisions in the same UPDATE; `StartOperation` commits job, operation, and audit atomically; Finish and Upsert are fenced on state and revisions. RLS on all five tables with composite FKs; all routes behind `authz` (view for reads, administer for mutations); the worker runs handlers under `shared.WithTenant(job.TenantID)`; the API root registers the integration handler only in the in-memory branch, so Postgres deployments do provider egress only in `synapse-worker`. The UI uses `@untitledui/icons` and shared components with no raw colors.

Once blockers 1 through 4, pho-veteran's items, and 5 through 7 are fixed, re-request review. Merge order: after #779 lands as 0129.

nghiadaulau (changes requested):
Re-reviewed the fix commit `6f9729f5` (diff `66eee831..6f9729f5`, 21 files, +752/-259) with build, vet, CGO-disabled build, unit tests, the Postgres suite on a fresh Postgres 17, and the web tests. Changes still requested. The commit resolves 9 of the 14 items reported on the previous head; the two CI-red items and the scheduler bug are untouched.

## Resolved, with evidence

- Credential PUT and DELETE now carry `version` and `connection_revision`, compared under `FOR UPDATE` in Postgres and under the store lock in memory; stale version, stale revision, and exact replay all return 409 (Postgres probe).
- Poll admission compares caller-supplied revisions under lock and rejects disabled rows; `HandleJob` and `FinishIntegrationPoll` re-check (Postgres probe: stale admission leaves 0 `jobs` rows).
- Every administrative mutation now takes a `ports.AuditEntry` and appends it inside the same transaction (`appendTenantAudit`); the memory twin rolls back on audit failure (`TestIntegrationMutationsRejectAuditFailureAndArchivedCredentialWrites`).
- `UpsertIntegrationExternalRuns` is gone; `FinishIntegrationPoll` upserts runs and the terminal state in one transaction with a `state='running'` guard, so a late publish after cancel returns 409 and stores nothing (Postgres probe).
- Poll has a 2-minute deadline, a 250-request and 32 MiB budget, `MaxBindingsPerPoll=100`, and an aggregate run cap; cancel bumps `claim_fence` so the heartbeat cancels the request context within one interval.
- Archive deletes credentials in the same transaction; archived rows reject credential writes and binding creation (Postgres probe).
- `safeRunURL` runs the reflection check on the decoded path and the final string; single percent-encoding of the token is now rejected.
- `UpdateIntegrationRequest` is a satisfiable flat schema; `TestOpenAPI_StrictValidation` passes.
- Detail loading uses a generation counter and `selectedIdRef`; the stale-response test passes. Update and enable responses hydrate `CredentialConfigured`.

## Still blocking

**1. The Postgres scheduler query is unchanged.** `postgres/integration_store.go:566-567` still reads `<= $1 - make_interval(secs=>poll_interval_seconds)` with an untyped `$1`. Reproduced again on Postgres 17: `ERROR: operator does not exist: timestamp with time zone <= interval (SQLSTATE 42883)`. Every `Scheduler.Tick` fails on Postgres; no scheduled poll is ever dispatched. Fix `$1::timestamptz` and add a Postgres test that seeds an enabled integration and asserts `ListDueIntegrations` returns it.

**2. The Postgres suite is red with the same 7 PR-introduced failures.** `integration_store_test.go:182-184` still expects the cancelled job to be claimable while `CancelIntegrationOperation` (`:541-542`) sets `jobs.status='done'`: `cancelled claim=<nil> err=<nil>`. And the test now leaves five `hash_version=2` `audit_log` rows for its tenant (the fix writes audit for every mutation), which trips migration 0085's Down guard in `TestMigration0044/0056/0058/0061/0066/0068` (`cannot roll back audit_log tenant chains after tenant genesis rows exist`). `audit_log` is append-only by trigger, so cleanup cannot delete them; this test needs its own database, or the assertion and fixture must avoid leaving tenant genesis rows in the shared one. On a fresh database this head fails exactly these 7 tests.

**3. Username substring rejection is unchanged.** `jenkins/client.go:530-541` still matches `adapter.username` as a substring over `name`, `fullName`, `job.URL`, `job.Class`, `externalKey` (`:141`) and `build.Result`, `revision`, `branch` (`:268`), and now also over the canonical run URL, which contains the host. Probe: username `jenkins` against `https://jenkins.example.com/jenkins/job/app/` fails discovery; username `job` fails on `org.jenkinsci.plugins.workflow.job.WorkflowJob`; username `main` fails on branch `main`; username `app` fails on job `app`. Each is a permanent failure surfaced as `provider operation failed`. Match only the token and its Basic encodings.

**4. Double percent-encoding bypasses the reflection check.** With token `token`, `safeRunURL("https://jenkins.example.com/jenkins/job/release/%2574%256f%256b%2565%256e/1/")` returns the URL unchanged and without error: `url.Parse` decodes once, `resourceURL` re-escapes `%` to `%25`, and neither `parsed.Path` nor the canonical string contains the plaintext. The value is persisted in `integration_external_runs.run_url` and rendered as a link (`Integrations.tsx:477`); a viewer-role user decoding twice recovers the admin's token. Unescape until stable before matching, and reject run URLs whose path still contains `%25` after one decode.

**5. Discovery still has no budget or deadline.** `executeDiscover` (`service.go:477-500`) calls `DiscoverPipelines` with the raw execution context: no `WithTimeout`, no `WithOperationBudget`, so the new `ConsumeOperationRequest`/`ConsumeOperationBytes` are no-ops there. The bound is still `maxDiscoveryNodes=2000` at 20 s each (up to 11 h and 8 GiB per discovery) with the worker heartbeat extending the claim; `HandleJob` (`:389-451`) has no wall-clock deadline.

**6. Postgres `UpdateIntegration` treats every rename as a material change.** `integration_store.go:101` compares `string(current.Config)` (JSONB text, reordered and spaced) with `json.Marshal` output. Probe: create with `{"folder":"team-a","depth":2}`, rename only: `enabled true -> false`, `connection_revision 1 -> 2`, in-flight operations invalidated. Memory compares compact bytes, so the twins disagree. Compare `config IS DISTINCT FROM $4::jsonb` or decoded maps.

**7. `safehttp` transport leak is unchanged.** `git diff 66eee831 6f9729f5 -- internal/infrastructure/safehttp` is empty; no `IdleConnTimeout`, `MaxIdleConns`, or `CloseIdleConnections` anywhere under `internal/infrastructure/integration`.

Also unchanged from the first review: `allow_private_network` is settable by any PermAdminister principal with no operator gate and audit metadata for create/update still carries only `provider`; the scheduler still writes one `operation_started` audit row per dispatched poll (2880 per day per integration at the minimum interval), each taking the global audit advisory lock.

## New in this commit

**8. Binding cap is enforced at poll time, not at creation.** `CreateBinding` (`service.go:246-262`) has no cap while `executePoll` (`:511-513`) fails permanently above `MaxBindingsPerPoll`. Probe: 101 bindings create fine, the next poll ends `state=failed errors=[provider operation failed]`, and the scheduler requeues it every interval. `finishProviderFailure` (`:601`) also replaces "exceeds 100 bindings", "exceeds 200 runs", and the budget errors with the generic message, so the operator cannot see why. Reject the 101st binding with 400 and keep the bounded-budget reasons as sanitized error text.

**9. Matcher deadline is misclassified as lease loss.** `MatchIntegrationAnalysis(pollCtx, ...)` (`:554-557`) returns `context.DeadlineExceeded` when the 2-minute budget expires in the matcher; `executePoll` returns it raw and `retryIntegrationLeaseLoss` (`:379-387`) converts it to `ErrRetryable` because the lease context is still live. The job retries with a fresh budget each time. The `ReadRuns` branch (`:526-528`) handles this; the matcher branch does not. Traced in code, not executed.

**10. Handlers do not validate the new required fields.** OpenAPI marks `version` and `connection_revision` required with `minimum: 1`; `integration_handler.go:153-181` forwards zero values, which surface as 409 instead of 400.

**11. Memory twin records audit before a mutation that can still fail** (`memory/integration_store.go:409-415, 447-451`, `invalidateActiveOperationsLocked`), leaving an audit row for a change that did not happen. Postgres is atomic; this is dev and test fidelity only.

**12. `make lint` fails on PR files** (present at the previous head): `jenkins/client.go:417` errcheck on `response.Body.Close`, `:106,124,127` ST1005 capitalized error strings, `memory/integration_store.go:707` unused `cloneJSON`.

**13. Test gaps.** No Postgres test for stale credential PUT/DELETE, stale admission, or the `FinishIntegrationPoll` cancel fence (memory only). No HTTP test for `DELETE .../credentials` with the new body. No test for a rename-only update on Postgres (would have caught 6). No test for `ListDueIntegrations` on Postgres (would have caught 1). No test for discovery bounds. No `CHANGELOG.md` entry, and the credential API contract changed (required `version` and `connection_revision`).

## Verified clean on this head

`go build ./...`, `go vet ./...`, `CGO_ENABLED=0 go build ./cmd/...`, the architecture tests, and `go test` on `domain/integration`, `usecase/integrations`, `jenkins`, `safehttp`, `persistence/memory`, `httpapi`, `platform/config`, and both `cmd` packages pass. The hostile harness passes. The two touched vitest files pass (9 tests). Drift guards show only the two pre-existing failures from `main`. Postgres parity probes for the resolved items above all behave identically in memory and Postgres.

Blockers 1 and 2 are each a one-line and a one-test fix; 3 and 4 are one function. Once 1 through 7 are addressed, re-request review.

nnatuan03:
All remaining requested changes are addressed in `2efb500`.

Re-review guide:

1. **Postgres correctness:** `ListDueIntegrations` now casts `$1::timestamptz`; the integration suite uses an isolated database, expects cancelled jobs to be unclaimable, and covers due scheduling, semantic rename-only config comparison, stale credential PUT/DELETE, stale admission, and cancelled-poll publication fencing.
2. **Jenkins output safety:** username substrings are no longer treated as secrets; token and Basic-auth encodings are still rejected. URL paths are decoded to stability, ambiguous `%25` paths are rejected, and single/double encoded token regression tests are included. Default HTTPS port normalization is supported.
3. **Bounded execution:** test/discovery/poll have per-type deadlines, discovery shares request/byte budgets, queue data is fetched once per operation, matcher deadline/budget failures terminate with bounded diagnostics, and cancellation is checked during discovery.
4. **Transport lifecycle and SSRF:** safe HTTP transports now have connection/idle bounds and adapters close idle connections. The default-off `SYNAPSE_INTEGRATION_ALLOW_PRIVATE_NETWORK` operator gate is enforced by Create/Update; audit metadata includes endpoint + flag; 6to4 and well-known NAT64 ranges remain blocked.
5. **Admission and publication fences:** the 101st binding is rejected at both service and store boundaries; disabling cancels active work; memory mutation audit ordering no longer records a mutation when job invalidation fails; scheduled polls no longer flood the human audit chain.
6. **Architecture follow-ups from the first review:** external keys are provider-neutral while Jenkins owns `/job/`; the tenant lister is provider-neutral; commit matching now lives in `ProjectAnalysisStore`; analysis FKs are tenant-composite; job retention is no longer blocked by the operation schema; safe provider diagnostics are preserved and bounded.
7. **Contract/docs:** handlers validate required optimistic-concurrency fields as 400, credential DELETE body is exercised, hostile route denial is covered, and OpenAPI/operator docs/CHANGELOG are updated.

Validation completed:

- Focused Go tests pass for domain/integrations, service/scheduler, Jenkins, safehttp (including local TLS), memory/Postgres compilation, HTTP API, config, and both commands.
- Full repository compile, focused `go vet`, normal build, and CGO-disabled build pass (sandbox emitted only Go module stat-cache warnings).
- OpenAPI strict/dead-operation tests pass. Full route coverage still reports the same 15 pre-existing Fleet routes.
- All 391 web tests and production build pass. Typecheck still has the same pre-existing `OverviewMetricCard.tsx` `Grade` export error.
- PostgreSQL runtime execution is not available on this host (`SYNAPSE_TEST_DB_DSN` is unset); the isolated Postgres suite compiles and is ready for CI/fresh Postgres 17.
- `golangci-lint`/Semgrep binaries are not installed on this host; all previously reported lint findings were removed and `go vet` passes.

Please re-review commit `2efb500`, especially the fresh-Postgres suite and the URL decoding/transport lifecycle paths. Merge order remains after #779 so migration 0129 precedes 0130.

PR #814, opened 2026-09-04
feat: add assessment cycles and immutable snapshots

nnatuan03:
Split from #811 in response to the requested decomposition.

Depends on #779 and intentionally uses its ScanRun provenance model. Because #779 comes from a fork, this PR temporarily targets codex/pr-779-base, which is an exact mirror of the current #779 head SHA. Retarget this PR to main immediately after #779 merges. The duplicate ScanRun implementation and migration from #811 are not present in this diff.

Scope
- Assessment Cycle and Re-test lifecycle API, immutable Snapshot finalization, UI, and OpenAPI contracts.
- Resumable Cycle and Snapshot backfills plus the read-only integrity verifier.
- Default-off rollout gates and tenant allowlists.
- Migrations 0131 through 0135, following #779 migration 0129 and the reserved Integrations migration 0130.

Review fixes included
- Legacy completion remains compatible while Snapshot enforcement is disabled; enabled enforcement fails closed.
- Backfill and integrity publication use unique lease-token fencing with transactional item and business writes.
- Resume rejects schema, dry-run, or batch-size mismatches.
- Snapshot ETags return the real default-pointer CAS version and legacy-only history lists with version zero.
- Tenant ownership, forced RLS, immutable hashes, idempotency, and optimistic concurrency are preserved.

Validation
- Focused Cycle, Snapshot, backfill, integrity, HTTP, Postgres, and migration tests pass.
- Race checks for the Assessment, persistence, HTTP, and observability packages pass.

Please review this PR as the Cycle/Snapshot slice only; Finding Lineage and Comparison/Closure are stacked separately.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,19 @@ and this project aims to adhere to [Semantic Versioning](https://semver.org/spec
 
 ### Added
 
+- **Tenant-gated Assessment Cycles and immutable Assessment Snapshots.** New Cycle, Re-test,
+  archive/reopen, Snapshot finalization, bounded history, historical projection, and integrity APIs
+  preserve legacy completion while rollout is disabled and enforce native finalized evidence only for
+  explicit tenant cohorts. The `synapse-assessment-backfill`, `synapse-assessment-integrity`, and
+  `synapse-assessment-snapshot-backfill` operators provide resumable, lease-fenced migration tooling.
+  Migrations 0131–0135 add the tenant-isolated Snapshot, retained-request, backfill, and integrity
+  stores. Rollout is controlled by `SYNAPSE_ASSESSMENT_CYCLE_API_ENABLED`,
+  `SYNAPSE_ASSESSMENT_CYCLE_DUAL_WRITE_ENABLED`, `SYNAPSE_ASSESSMENT_CYCLE_DUAL_WRITE_TENANTS`,
+  `SYNAPSE_ASSESSMENT_SNAPSHOT_ENABLED`, `SYNAPSE_ASSESSMENT_SNAPSHOT_COMPLETION_ENABLED`,
+  `SYNAPSE_ASSESSMENT_SNAPSHOT_COMPLETION_TENANTS`, `SYNAPSE_ASSESSMENT_LIFECYCLE_READ_ENABLED`,
+  `SYNAPSE_ASSESSMENT_LIFECYCLE_READ_TENANTS`, `SYNAPSE_ASSESSMENT_LIFECYCLE_UI_DEFAULT_ENABLED`,
+  and `SYNAPSE_ASSESSMENT_LIFECYCLE_UI_DEFAULT_TENANTS`.
+
 - **Python Tier-2 semantic reachability and interprocedural taint analysis.** An opt-in, source-only
   tree-sitter sidecar now emits bounded semantic facts without importing or executing target Python;
   pure Go resolution proves affected-symbol call paths and precise value flow across assignments,
--- a/api/openapi.yaml
+++ b/api/openapi.yaml
@@ -29,6 +29,10 @@ tags:
 - name: health
 - name: aup
 - name: engagements
+- name: assessment-cycles
+  description: Tenant-scoped initial Assessment and Re-test lifecycle Cycles
+- name: assessment-snapshots
+  description: Immutable coverage snapshots finalized from trusted Scan Runs
 - name: findings
 - name: sca
 - name: export
@@ -128,7 +132,12 @@ paths:
         "403": {$ref: "#/components/responses/Forbidden"}
     post:
       tags: [engagements]
-      summary: Create an engagement
+      summary: Create an initial Assessment and Assessment Cycle
+      parameters:
+      - name: Idempotency-Key
+        in: header
+        required: true
+        schema: {type: string, minLength: 1, maxLength: 128}
       requestBody:
         required: true
         content:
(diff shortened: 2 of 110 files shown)


nnatuan03:
@pho-veteran @nghiadaulau — review requested for the Cycle/Snapshot slice.\n\nThis PR targets codex/pr-779-base, an exact mirror of #779 head SHA 3bf8ca2 because #779 comes from a fork. Please review only the commit above that base.\n\nFocus areas: default-off completion compatibility, Snapshot pointer CAS/ETag and legacy-only listing, lease-token/ABA fencing, transactional backfill publication, resume-mode validation, integrity fencing, and RLS/immutability.\n\nMerge guidance: merge #779, then #813; retarget this PR to main; verify the diff remains limited to this slice; then merge. Reviews can happen concurrently with the other slices.

pho-veteran (changes requested):
I reviewed the exact `3bf8ca24f274db1f9fb770ddee36e74e2f19157c..45b70ccc69ef6bc9ff9ebc6bc804fbc6ac863e12` diff. Requesting changes for the following introduced issues.

1. **[P1] The default-off Snapshot rollout blocks all Assessment completion.** `internal/usecase/engagement/service.go:204-211` rejects every transition to `completed` when the Snapshot reader is nil, while `cmd/synapse-api/main.go:695-698` installs that reader only when `SYNAPSE_ASSESSMENT_SNAPSHOT_ENABLED=true` (whose default is false). The documented legacy-compatible/default-off configuration therefore cannot complete any Assessment. Enabling the flag also applies the guard globally rather than only to a verified rollout cohort. Please inject an explicit per-tenant completion policy: preserve legacy completion while disabled and require a configured finalized default Snapshot only where enforcement is enabled.

2. **[P1] Interactive finalization accepts legacy Scan Runs as a default Snapshot.** `internal/usecase/assessmentsnapshot/service.go:216-226` synthesizes a hash for `ProvenanceLegacy`, and `Finalize` then installs that Snapshot through `CreateFinalizedCAS`. This contradicts the native-finalization contract in `docs/guide/assessment-lifecycle-operations.md:49`; the completion guard checks only `LifecycleFinalized`, so legacy evidence can satisfy completion without native manifest verification. A focused probe confirmed that a tenant-owned legacy run is accepted and becomes default version 1. Please reject non-native provenance in interactive `Finalize` and keep legacy acceptance confined to `LegacyProjector`, which intentionally does not update the default pointer.

3. **[P1] Snapshot state and its audit record are not one atomic commit.** `internal/usecase/assessmentsnapshot/service.go:158-173` and `internal/usecase/assessmentsnapshot/backfill.go:104-143` place both operations inside `TenantTransactionRunner.Run`, but the production logger starts and commits a separate transaction at `internal/infrastructure/persistence/postgres/aup_audit.go:261-285`; the memory runner at `internal/infrastructure/persistence/memory/tenant_tx.go:26-40` only serializes callbacks and cannot roll repository writes back. PostgreSQL can therefore retain an audit event if the outer commit fails, and memory mode retains the Snapshot when audit fails even though the API returns an error. A focused failing-audit probe reproduced the latter. Please make audit recording reuse the context-bound PostgreSQL transaction and add rollback/compensation semantics for the in-memory Snapshot write.

4. **[P1] New retained Cycle API and backfill mutations silently discard audit failures.** The new mutation paths call `service.record` (for example `internal/usecase/assessmentcycle/api_service.go:154-157` and `:284-287`), but `record` drops `AuditLog.Record` errors at `:661-667`; the new historical Cycle writer does the same at `internal/usecase/assessmentcycle/service.go:164-168`. These operations can commit and retain a successful response without their required audit-chain entry. Please return and propagate audit errors from these new paths and run compensation/rollback before publishing a retained response.

5. **[P1] Existing Business Asset reassignment can invalidate the new frozen Cycle/Snapshot boundary.** The new finalization path copies the Cycle boundary at `internal/usecase/assessmentsnapshot/service.go:151-154`, and Cycle membership validates the Assessment asset, but the existing `AssignEngagement` mutation at `internal/usecase/businessassetuc/service.go:199-214` remains able to change or clear that asset afterward. The source Assessment, Cycle, and immutable default Snapshot can then name different governed boundaries, while the integrity verifier and future reporting trust the frozen copies. Please reject asset reassignment once an Assessment belongs to a Cycle (with a database invariant as well as a use-case check), or atomically define a governed boundary-change workflow.

6. **[P1] The integrity verifier can publish `completed/clean` over concurrently changed state.** `snapshot_at` limits only the initial Engagement-ID query at `internal/infrastructure/persistence/postgres/assessment_cycle_integrity_repo.go:33-40`; memberships, Cycles, and Engagement boundaries are read at current state later in each page (`:67-119`). On resume, stored subjects are skipped at `internal/usecase/assessmentcycle/integrity.go:140-145`, and completion only recounts rows at `:119-131`. An already checked Cycle can therefore be reassigned/corrupted before the last page and the run still completes clean. Please fence the run with an integrity-relevant source generation, or re-read and compare deterministic source hashes for every persisted subject before marking it completed.

7. **[P2] Replaying an older finalization returns a stale, unusable ETag.** The fast replay path at `internal/usecase/assessmentsnapshot/service.go:95-99` returns the historical Snapshot object, and `internal/adapter/httpapi/assessment_snapshot_handler.go:105-109` emits its historical `DefaultVersion`. If another finalization has advanced the pointer, a client following that ETag immediately conflicts on its next request. Please load the current default pointer on replay and return its current version in both the wrapper and ETag.

8. **[P2] Re-test scope and lineage can come from different Assessments.** `internal/usecase/assessmentcycle/api_service.go:223-256` copies name/client/scope from the route Assessment, while `:260-263` permits a different explicit predecessor to be recorded by `CreateRetest`. The new Assessment can therefore claim lineage from B while inheriting governed scope from A. Please resolve the effective predecessor first and derive inherited fields from it, or reject an explicit predecessor that differs from the route Assessment.

9. **[P2] Re-test writes bypass the tenant lifecycle rollout gate.** `internal/adapter/httpapi/router.go:541-547` applies the tenant read predicate to adjacent lifecycle routes but not to `POST .../retests`; the global creation form exposes the control unconditionally at `web/src/pages/Engagements/components/CreateEngagementForm.tsx:221-229`. Enabling the global Cycle API therefore permits lifecycle mutations for tenants that have not passed backfill/integrity/read canary checks. Please enforce a tenant-scoped lifecycle write predicate on the route and hide the control unless the tenant UI capability is enabled.

10. **[P2] Initial-Assessment retries cannot reuse the retained operation's idempotency key.** `web/src/lib/api/engagements.ts:62-78` creates `newIdempotencyKey()` inside each API invocation. If the server commits but the response is lost, the user's retry generates a new key and creates another Assessment/Cycle instead of replaying the retained response. The re-test form already keeps its key in operation state; please do the same for both initial creation variants and rotate it only after success or explicit reset.

11. **[P2] Snapshot history is an unbounded N+1 endpoint.** `internal/adapter/httpapi/assessment_snapshot_handler.go:112-135` returns every immutable Snapshot, while `internal/infrastructure/persistence/postgres/assessment_snapshot_repo.go:266-292` first reads all IDs and then fully loads each aggregate separately. Unique request keys permit history to grow without bound, so a normal read can eventually cause excessive queries, memory, and response size. Please add bounded cursor pagination and enforce the limit in SQL and OpenAPI.

12. **[P2] The shipped Re-test flow tells migrated users to perform an operation this API does not expose.** Historical completed Assessments are backfilled as completed Cycles at `internal/usecase/assessmentcycle/service.go:144-148`; Re-test creation rejects them with `cycle_reopen_required`, and the UI says “Reopen the Cycle first,” but this diff has no reopen route or OpenAPI operation despite `ReopenCycle` existing only as an internal method. Please expose an authorized, reasoned, version-checked reopen operation in this slice or avoid advertising the Re-test flow until that operation exists.

Validation performed on the reviewed head:

- `go test -race ./internal/usecase/engagement ./internal/usecase/assessmentsnapshot ./internal/usecase/assessmentcycle ./internal/infrastructure/persistence/memory ./internal/adapter/httpapi` passed.
- Focused `go vet`, including PostgreSQL persistence, passed.
- The two focused probes above reproduced legacy default finalization and memory persistence after audit failure; the temporary probe file was removed.
- `TestOpenAPIRouteCoverage` fails with the same Fleet-route debt on the exact base revision, so I did not attribute it to this PR.

nghiadaulau (changes requested):
Reviewed head `45b70cc` against base `codex/pr-779-base` (`3bf8ca2`) with build, vet, CGO-disabled build, unit tests, the Postgres suite on a fresh Postgres 17, the drift guards, and the web typecheck and tests. Changes requested. The 12 items from pho-veteran's review remain open (no commit since 2026-09-04 04:21). The findings below are new or extend those items; each was reproduced on this head.

## Blockers

**1. The Postgres suite is red on this head (38 failures vs 4 environmental on the base).**

a. Eight PR tests migrate to the wrong version. `assessment_cycle_backfill_repo_test.go:23,83` runs `goose.UpTo(132)` but the table is created by 0133; `assessment_cycle_integrity_repo_test.go:20,90` uses 133 (table in 0134); `assessment_snapshot_backfill_repo_test.go:23,100,140,202` uses 134 (table in 0135). Result: `ERROR: relation "assessment_cycle_backfill_runs" does not exist (SQLSTATE 42P01)` and the same for the integrity and snapshot backfill tables. Every Postgres claim in the PR body for lease fencing, resume, concurrent start, and RLS on those tables is unverified.

b. `assessment_cycle_api_test.go:27,132` uses the shared `setupTestDB` and leaves rows in `assessment_cycle_api_requests`. The Down guard in `0132:56-60` then raises on every `goose.DownTo` in 23 pre-existing migration tests (0044, 0056, 0058, 0060, 0061, 0066, 0068, 0069, 0081, 0089, 0091, 0093, 0095 to 0099, 0101, 0107, 0123 to 0126, 0129):

```
TestMigration0129RollbackAndReapply: down to 128: ERROR 0132_assessment_cycle_api.sql:
  cannot roll back assessment cycle API idempotency while retained responses exist (SQLSTATE P0001)
```

None of these fail on the base. Clean up the rows in the test or use a per-test database.

c. Three snapshot tests fail with `validate selected lane .../sca: validation error: lane manifest hash mismatch` (`TestPostgresAssessmentSnapshotRepositoryRLSCASReplayAndImmutability`, `...ConcurrentInitialCASHasNoNumberGap`, `...ConcurrentReplayAndAuditRollback`). Root cause is in #779: `scanrun.ComputeManifestHash` hashes `time.Time` at nanosecond precision and `timestamptz` stores microseconds, so `validateSealedLane` recomputes a different hash after the round trip. The fixture `postgresNativeRun` uses `time.Now()`, which is nanoseconds on Linux and microseconds on macOS, which explains "passes locally". Reported on #779; until it is fixed any producer that seals with nanosecond stage times produces runs that can never be finalized.

**2. Completion is blocked in both flag states.** With the defaults (`SYNAPSE_ASSESSMENT_SNAPSHOT_ENABLED=false`), `engagement/service.go:204-207` returns `assessment snapshot completion guard is not configured` because `main.go:695-697` only injects the reader when the flag is on; the PR's own test at `engagement/service_test.go:294-307` asserts this regression, and no HTTP test completes an engagement. With the flag on, no scan run can become a finalized default Snapshot: the pipeline writes only legacy runs (`memory/scan_run_repo.go:55` and the Postgres `Save`), `Finalize` rejects them (`selected scan run "legacy-1" has no provenance lanes`), `CreateLegacyProjection` (`postgres/assessment_snapshot_repo.go:107-150`) never writes `assessment_snapshot_defaults`, and `usecase/scanrun` has no importer outside its tests. Treat a nil reader as the legacy path, and decide explicitly whether a legacy projection seeds the default pointer.

**3. The Snapshot backfill cannot process a completed historical Assessment on Postgres.** Trigger `synapse_assessment_snapshot_guard` (`0131:124-128`) raises on INSERT unless the engagement is draft or active. `CommitAssessmentSnapshotBackfillItem` (`postgres/assessment_snapshot_backfill_repo.go:99-142`) runs `build()` inside the same transaction with no savepoint, so the trigger error aborts the transaction and the item row is never written. Probe with one completed engagement: `state=failed processed=0 ... ERROR: current transaction is aborted (SQLSTATE 25P02)`; resume fails identically. Completed assessments are what the backfill exists for; existing tests only use draft engagements.

**4. Drift guard fails on this head.** `TestEveryGuidePageIsPublished`: `docs/guide/assessment-lifecycle-operations.md exists but is not in mkdocs.yml nav`. No `CHANGELOG.md` Unreleased entry for three binaries, eight `SYNAPSE_*` variables, and five migrations.

## High

**5. Audit and state are not one commit, and audit errors are discarded.** Root cause: `postgres/aup_audit.go:261-267` always does `l.pool.Begin(ctx)` and never consults the transaction bound by `TenantTransactionRunner.Run`. Callers inside `tx.Run` therefore commit the audit row even when the business write rolls back (probe: audit row survives a rolled-back Finalize). In the other direction, `api_service.go:661-668`, `service.go:165,299,431,545,618,697,748,798`, `backfill.go:214`, `integrity.go:357`, and `assessmentsnapshot/backfill.go:403` do `_ = audit.Record(...)`, so a retest creation (a new engagement with an authorization window) or a cycle archive can commit with no audit entry. `enguc.auditChange` and `snapshotuc.Finalize` in the same tree return the error. Fix: have `recordOnce` join `contextTenantTx` when bound (pattern in `fleet_audit_repo.go:73`) and propagate every audit error.

**6. The rollout gate covers reads but not the two writes.** `router.go:542` (`POST .../retests`) and `:550` (`POST .../snapshots/finalize`) lack `requireAssessmentLifecycleRead`, while every sibling read has it. `config.go:1172-1177` `AssessmentCycleDualWriteForTenant` returns true for every tenant when the API flag is on, ignoring `SYNAPSE_ASSESSMENT_CYCLE_DUAL_WRITE_TENANTS`. A tenant outside the allowlist can create retest engagements and move the CAS pointer while every GET returns 404.

**7. Replay returns a stale ETag and ignores `If-Match`.** `assessmentsnapshot/service.go:87-93` returns the stored snapshot before the CAS check; `assessment_snapshot_handler.go:118-121` emits the creation-time `DefaultVersion`. Probe: replay of key k1 with `If-Match: 77` after snapshot 1 was superseded returned `201`, `ETag "1"`, `default_version 1`, `lifecycle superseded`. Return the current pointer version, or 409 on a mismatched precondition.

**8. A Re-test on a completed Cycle is unreachable.** `api_service.go:462-464` returns `cycle_reopen_required`; `router.go:541-548` registers no reopen or complete route; `ReopenCycle` (`service.go:721`) is internal. `BackfillHistoricalSingleton` (`service.go:90-95`) sets historical cycles for completed assessments to `completed`, so every backfilled completed Assessment can never receive a Re-test (probe: `POST .../retests` 409, `POST .../reopen` 404).

**9. The in-memory store corrupts the default Snapshot on a second finalize when any lane has an empty scope list.** `memory/assessment_snapshot_repo.go` `cloneAssessmentSnapshot` turns `[]string{}` into nil, canonical JSON changes `[]` to `null`, `Supersede` mutates lifecycle before `Validate()` fails, and the memory transaction runner has no rollback. Probe: `second err=validation error: snapshot content hash mismatch`; first snapshot is `superseded` while the pointer still points at it; `Get` now returns `excluded_scope: null`. Existing tests always set both scope lists.

**10. The hostile harness does not exercise the eight new routes.** `harness_test.go:340-372` wires 16 services and never calls `SetAssessmentCycles` or `SetAssessmentSnapshots`, so tenant-B probes against tenant-A cycle, snapshot, `predecessor_assessment_id`, and `run_id` are not covered by the gate CONTRIBUTING relies on.

## Medium

**11. The resumable run machinery is built three times.** `ports/assessment_cycle_backfill.go:20-40` and `ports/assessment_snapshot.go:43-76` have field-for-field identical run structs; the runners in `assessmentcycle/backfill.go:77-197`, `assessmentsnapshot/backfill.go:186-360`, and `integrity.go:80-171` are copies; four repo files (884 lines), migrations 0133 and 0135 (same DDL), and two mains differ only in identifiers. #815 adds a fourth copy. One `ResumableRunStore` with a kind column and one runner taking an item processor.

**12. HTTP concerns inside the use case and the ports.** `api_service.go:82-86, 157, 288, 582, 626-637` return HTTP status codes and a `json.Marshal`ed domain entity as the retained body; `ports/assessment_cycle.go:60-66` carries `StatusCode` and `ResponseBody`. Adding JSON tags to `Engagement` silently changes what replays return. Keep encoding and status in the adapter (`vulnerability_handler.go:446` is the existing convention).

**13. Memory and Postgres list contracts diverge.** `memory/assessment_cycle_repo.go:289-291` returns an empty page for any `AssessmentStatus` or `ScanStaleness` filter; `ListMigrationPendingAssessments` at `:347-349` is a stub; memory `ListCycles` never fills `RootSnapshotID`, `CurrentSnapshotID`, `SelectedHeadScanAt`. Because the memory transaction runner cannot roll back, the use case carries about 60 lines of manual compensation (`service.go:105-107, 133-135, 201`, `api_service.go:146, 265, 628-635`, `engagement/service.go:CompensateCreate`) that only the dev twin ever exercises.

**14. Cycle invariants are re-implemented in the integrity verifier** (`integrity.go:173-315`, including `hasIntegrityGraphCycle`) while `domain/assessmentcycle/member.go` already has `IsRoot`, `DeriveBranchHeads`, `DeriveAncestors`. A new aggregate invariant will not be verified. Put `VerifyInvariants` in the domain.

**15. Lease fencing trusts the caller's clock.** `Commit`, `Advance`, `Finish`, and `Acquire` in `postgres/assessment_snapshot_backfill_repo.go` compare `lease_expires_at` against a parameter, not `now()`; clock skew lets a stale worker keep committing.

**16. `Idempotency-Key` becomes mandatory on the pre-existing `POST /api/v1/engagements`** once dual-write is on (`engagement_handler.go:199-218`, `api_service.go:848`). The web client was updated; CLI and CI clients get 400. Derive a server-side key when the header is absent.

**17. "Explicit execution authorization" is satisfied by a blackout-only RoE.** `engagement.go:139-145` accepts `len(Blackouts) > 0`, and `roe.go:40-43` treats empty `AllowedToolClasses` as every class allowed, so the new guard degrades to "any window plus any blackout" from the same actor.

**18. Unbounded snapshot history and unbounded idempotency retention.** `postgres/assessment_snapshot_repo.go:266-296` loads every snapshot with 3 plus runs queries each and `assessment_snapshot_handler.go:112-136` has no limit; `assessment_cycle_api_requests` (0132) keeps up to 2 MiB per key with no TTL.

**19. Client-controlled idempotency key is written verbatim into hash-chained audit metadata** (`api_service.go:666`). Store a hash or omit it.

**20. The default-pointer FK does not bind the snapshot to the pointer's assessment.** 0131 references `(tenant_id, snapshot_id)`; 0135 adds `UNIQUE (tenant_id, assessment_id, id)`, so tighten the FK to `(tenant_id, assessment_id, snapshot_id)` and add DELETE guards on defaults and counters.

## Low

- `domain/assessmentsnapshot/comparison.go` (132 lines) and `assessmentsnapshot.Service.Compare` have no caller; the PR body says comparison is stacked separately. `decodeAssessmentCycleJSON` (`assessment_cycle_handler.go:153`) is unused while `createAssessmentRetest` inlines the same body.
- `router.go:541-552` mixes `{assessmentId}` with use-case scoping and `{id}` with `withEngTenant` on the same resource.
- `App.tsx:116-121` and `Sidebar.tsx:102-103` each call `api.me()`; expose `features` from `useAuth()`.
- 97 files in one commit across four independently gated slices (cycle API plus 0132, snapshot plus 0131, backfill and integrity plus three binaries, UI). Blocker 2 survived "focused tests pass" because the compatibility claim was never tested at the composition boundary. Split along the flags.

## Verified clean

`go build ./...`, `go vet ./...`, `CGO_ENABLED=0 go build ./...`, the architecture test, and `go test` on the 15 touched packages pass. The hostile harness passes for the routes it covers. `pnpm run typecheck` passes; the touched vitest files pass in isolation. All 14 new tables carry `tenant_id` in the PK with composite FKs and `synapse_enable_tenant_rls`; every repository method goes through `WithTenant` with an explicit tenant predicate; `predecessor_assessment_id` and `run_id` are resolved through tenant-scoped lookups and lane keys only from the stored run. Snapshot triggers block every update other than building to finalized and finalized to superseded. The CAS pointer locks counters and defaults `FOR UPDATE`; idempotency reservation happens inside the transaction; lease-fenced commits check state, token, and expiry in the same transaction as the business write. Request bodies are bounded with `DisallowUnknownFields`.

Once blockers 1 through 4 are fixed and pho-veteran's items plus 5 through 10 are addressed, re-request review. This PR also cannot merge before #779 is fixed and merged and this PR is retargeted to `main`.

nnatuan03:
@pho-veteran @nghiadaulau — the requested Assessment Cycle/Snapshot fixes are pushed in `847ab02`.

The 12 findings from the first review are addressed:

1. Completion enforcement now has an explicit default-off, tenant-scoped policy; legacy completion remains unchanged outside the cohort.
2. Interactive finalization accepts native sealed Scan Runs only; legacy evidence remains confined to the non-default backfill projector.
3. PostgreSQL audit writes reuse the bound tenant transaction, and memory repositories now roll back Snapshot/Cycle/Engagement/request mutations on transaction failure.
4. Retained Cycle/API/backfill audit failures are propagated and roll back/compensate before a response is published.
5. Business Asset reassignment is rejected after Cycle membership in the use case, repository, and an upgrade-safe database trigger installed by migration 0131.
6. Integrity runs persist a tenant source generation and completion is serialized/fenced against concurrent source mutation.
7. Finalization replay returns the current default version/ETag and rejects an `If-Match` value that does not match the retained request precondition.
8. An explicit Re-test predecessor must equal the route Assessment.
9. Re-test, archive/reopen, and Snapshot-finalize writes require the tenant lifecycle canary; the Re-test UI is tenant-feature gated.
10. Initial create/upload retries reuse one client-generated idempotency key.
11. Snapshot history uses bounded cursor pagination (default 50, max 100) enforced in SQL and OpenAPI.
12. A permissioned, reasoned, version-checked, idempotent Cycle reopen API is now exposed.

The additional blocker/high findings from the second review are also covered: corrected `goose.UpTo` versions and retained-row cleanup; microsecond-stable manifest hashes across `timestamptz` round trips; completed legacy Snapshot backfill; MkDocs navigation and CHANGELOG coverage; empty-slice-preserving memory clones; and hostile-harness coverage for every Cycle/Snapshot route using tenant-A marker data. Further hardening includes database-clock lease checks across the new PostgreSQL run stores, a server-derived compatibility key when older create clients omit `Idempotency-Key`, 24-hour retained-response expiry with an indexed prune path, positive tool-class authorization for Re-tests, omission of raw idempotency keys from audit metadata, a three-column default-Snapshot FK plus pointer/counter delete guards, and one shared `/me` feature read in `AuthProvider`.

Validation on this head:

- `go build ./...`, `CGO_ENABLED=0 go build ./...`, and `go vet ./...`
- focused Go package tests for domain/config/Cycle/Snapshot/Engagement/memory/PostgreSQL adapters
- `go test -race` for Scan Run, Engagement, Cycle, Snapshot, and memory packages
- hostile tenant harness plus all lifecycle HTTP tests
- strict OpenAPI, Cycle/Snapshot contract, dead-operation, coverage-debt, and guide-publish guards
- web typecheck, all 384 Vitest tests, and production build

The full route-coverage test still reports the pre-existing Fleet OpenAPI debt identified on the base. A local PostgreSQL runtime was unavailable because Docker was not running, so please rerun the fresh-Postgres suite during review; the affected integration tests compile and the previously incorrect migration targets/fixtures are corrected.

Please re-review commit `847ab02`, focusing first on transaction/audit atomicity, tenant rollout gates, the source-generation fence, and migrations 0131–0135.

nghiadaulau:
Đóng vì mâu thuẫn với quyết định đã chốt trên main. PR này re-add subsystem `assessmentcycle` (internal/usecase/assessmentcycle, internal/domain/assessmentcycle, và store postgres/memory) — đúng những file mà #840 đã cố ý xóa khỏi main vì subsystem này không reachable (dead code: có domain+store nhưng không composition root/route/UI nào chạm tới), và đây là hướng user không muốn đảo lại. Thêm nữa: PR stack trên `codex/pr-779-base` (không phải main), 110 file/+14k, và 5 migration 0131–0135 đụng số (duplicate goose version = panic lúc khởi động) với migration đã ship trên main. Nếu muốn tính năng assessment snapshots/cycles thật, cần thiết kế lại theo hướng reachable end-to-end trực tiếp trên main (producer → route → store → UI), không phải bản re-add dead code này.

PR #815, opened 2026-09-04
feat: add finding lineage and fenced backfill

nnatuan03:
Split from #811 as the focused Finding Lineage slice. This PR is stacked on #814 so its diff contains only lineage work.

Scope
- Versioned Finding identities, immutable observations, deterministic producer matchers, review candidates, and override events.
- Secret-safe matcher behavior and redacted skip records.
- Resumable tenant-scoped historical lineage backfill.
- Ongoing shadow projection for Snapshot finalization plus newly created manual and offensive Findings.
- Migrations 0136 and 0137.

Review fixes included
- Manual and offensive Finding creation now projects lineage in the same tenant transaction for shadow-enabled tenants.
- Backfill business writes and item publication are atomic and fenced by a unique, unexpired lease token.
- Resume and takeover are bounded, tenant-scoped, and fail closed.
- Raw evidence, workflow, SLA, assignee, disposition, provider errors, and credential material are excluded from immutable lineage artifacts.

Validation
- Finding Lineage domain, matcher, service, backfill, memory, Postgres, migration, HTTP wiring, race, and vet checks pass.
- Tenant RLS, composite ownership, append-only behavior, and redaction have focused tests.

Base chain: #779 mirror -> #814 -> this PR. Retarget to main in order after prerequisites merge.


nnatuan03:
@pho-veteran @nghiadaulau — review requested for the Finding Lineage slice.\n\nThis PR is stacked on #814. Please review only the single commit between the base and head.\n\nFocus areas: same-transaction shadow projection for manual/offensive findings, Snapshot-finalization projection, secret matcher/redaction behavior, lease-token-fenced atomic lineage backfill, and RLS/append-only guarantees.\n\nMerge guidance: merge #814 first, retarget this PR to main, verify the diff remains limited to this slice, then merge. Reviews can happen concurrently with the rest of the stack.

pho-veteran (changes requested):
I found 13 issues introduced by this exact diff (`45b70ccc69ef6bc9ff9ebc6bc804fbc6ac863e12..ff6eb6668c48e1f0950660f5b2aac762f6d01055`). The first four can make immutable lineage disagree with the selected assessment boundary or with the result returned to the caller.

## High

1. **Snapshot finalization assigns every current Finding to the selected Snapshot.** `internal/usecase/findinglineage/shadow.go:82-94` calls `ListByEngagement` and projects every retained Finding, including scanner findings from earlier runs. Findings have no persisted ScanRun/lane link, and their repositories retain rows across rescans, so a Finding absent from the selected run is still written as an Observation for that Snapshot. A removed vulnerability can therefore appear present in immutable comparison input. Restrict this catch-up to manual/offensive Findings, and project scanner output from an immutable selected-run source or persisted source-to-run mapping.

2. **The production SCA projection uses a version-bearing component fingerprint as a stable dependency instance.** `internal/usecase/findinglineage/backfill.go:313-317` and `internal/usecase/findinglineage/shadow.go:133` pass `ComponentFingerprint` as `DependencyInstanceID`, which becomes an identity field at `internal/usecase/findinglineage/sca_matcher.go:152-153`. `internal/domain/sbom/identity.go:124-129` explicitly hashes package version and PURL into that value. Upgrading the same dependency therefore creates a different lineage Identity, even though the matcher tests expect upgrades at one dependency location to correlate. Use a version-independent dependency path/lockfile anchor; when none exists, create a provisional review candidate rather than treating `ComponentFingerprint` as stable.

3. **Concurrent first observations can create duplicate automatic identities.** `internal/usecase/findinglineage/service.go:244-260,316-321` performs a fingerprint lookup followed by an insert. Under PostgreSQL `READ COMMITTED`, concurrent transactions can both observe no match and insert different IDs. The index in `migrations/0136_finding_lineage.sql:26-27` is non-unique, and the advisory lock at `internal/infrastructure/persistence/postgres/finding_lineage_repo.go:309-311` protects candidate creation only. Identical requests can thus produce timing-dependent identity splits and later collision candidates. Lock the correlation namespace before lookup and repeat the lookup while holding the lock, reserving duplicate fingerprints for an explicit reviewed operation.

4. **A failed finalization observer leaves an unrecoverably partial Snapshot in memory mode.** The new observer runs after `CreateFinalizedCAS` at `internal/usecase/assessmentsnapshot/service.go:170-188`, but `internal/infrastructure/persistence/memory/tenant_tx.go:37-40` only serializes execution and cannot roll back. If projection fails, Finalize returns an error while the Snapshot remains persisted; retry then resolves the request-key replay before invoking the observer, so lineage is never repaired. Fail startup when shadow generation is enabled without PostgreSQL, or provide a real coordinated in-memory unit of work and retry incomplete projection.

## Medium

5. **Root CloudFormation resources cannot produce a complete fingerprint.** `internal/usecase/findinglineage/iac_matcher.go:225-236` always emits `nested_stack_path`, including an empty ordered array for a valid root-stack resource. Canonical serialization rejects empty arrays at `internal/domain/findinglineage/canonical.go:303-306`. Omit this optional field when the stack path is empty and add root-stack golden coverage.

6. **Backfill target selection accepts dimensions that match only half the namespace.** The condition at `internal/usecase/findinglineage/backfill.go:451-478` skips a dimension only when both producer and finding kind differ. A dimension with the requested producer but a different kind, or the requested kind under a different producer, is retained and can select the wrong target or create false ambiguity. Change the condition to skip when either value differs and test both half-match cases.

7. **Idempotent skip replay returns and audits an ID that was never persisted.** `internal/usecase/findinglineage/service.go:501-511` creates a fresh `SkipRecord.ID` and ignores the boolean returned by `AppendSkip`. On semantic replay, both repositories retain the old row (`internal/infrastructure/persistence/memory/finding_lineage_repo.go:514-517`, `internal/infrastructure/persistence/postgres/finding_lineage_repo.go:541-558`) while the service returns and audits the fresh ID. Return the stored record from the repository or derive a deterministic ID, then assert persisted, returned, and audited IDs are identical on replay.

8. **`CreateAttributed` no longer reports persisted partial writes when shadowing is disabled.** The base implementation wrapped post-upsert audit and attribution failures in `ports.PartialWriteError`. The new callback at `internal/usecase/findings/service.go:80-120` returns ordinary errors, while `withCreationBoundary` at `:451-465` runs without a transaction when shadowing is unconfigured or the tenant is not allowlisted. The Finding remains persisted, but callers lose its ID and cannot distinguish a repairable partial write. Preserve `PartialWriteError` outside an actual transaction and return ordinary rollback errors only for the PostgreSQL transactional path.

9. **Canonical text is hashed but the unnormalized text is used as the repository namespace.** `CanonicalizeFingerprintV1` NFC-normalizes the target, but `internal/usecase/findinglineage/service.go:245-247,540-546` looks up and stores the original target. Canonically equivalent target strings can therefore have the same fingerprint and still create separate Identities. `SourceReferenceHash` at `internal/domain/findinglineage/model.go:646-660` similarly validates normalized values but hashes the originals, and its NUL-delimited encoding admits tuple collisions because embedded NUL is accepted. Normalize once and use those values for hashing, lookup, Observation, and Identity storage; use structured encoding or reject the delimiter.

10. **PostgreSQL permits aliases and Observations to cross identity namespaces.** `ApproveAlias` at `internal/usecase/findinglineage/service.go:420-449` never checks that the selected Identity shares the alias's producer, finding kind, and target. PostgreSQL `AppendObservation` at `internal/infrastructure/persistence/postgres/finding_lineage_repo.go:76-100` also omits the namespace check enforced by the memory repository at `internal/infrastructure/persistence/memory/finding_lineage_repo.go:86-92`. An alias can therefore resolve to an unrelated Identity and permanently append mismatched evidence in PostgreSQL. Validate the alias against the Identity in the approval transaction and enforce the same invariant in PostgreSQL before insertion.

11. **Backfill can permanently record a later Snapshot as `FirstSeenSnapshotID`.** Source pagination is ordered only by random Finding ID at `internal/infrastructure/persistence/postgres/finding_lineage_backfill_repo.go:61-63`, and the first matching row irreversibly supplies `FirstSeenSnapshotID` at `internal/usecase/findinglineage/service.go:540-546`. Across member assessments in one Cycle, a later Snapshot with a lexically smaller random ID can be processed before the earlier Observation. Order by Snapshot chronology plus a stable tie-breaker, or determine the earliest matching Snapshot before creating the Identity.

12. **The 2,000-row batch limit does not bound auxiliary backfill memory.** For one source row, `internal/usecase/findinglineage/backfill.go:505-539` loads and caches every vulnerability occurrence or every Judgment for the assessment. A large assessment can exhaust memory despite the advertised batch ceiling, multiplied across up to four tenant workers. Add direct tenant-scoped lookup by occurrence/Judgment ID or fetch only IDs needed by the current source batch.

## Low

13. **The PostgreSQL tests stop before the migrations they claim to verify.** `internal/infrastructure/persistence/postgres/migration_0136_test.go:16-58` migrates only through 135 and rolls back to 134, while the lineage tables are created by 136. `internal/infrastructure/persistence/postgres/finding_lineage_backfill_repo_test.go:21,96` migrates only through 136 even though its run/item tables are created by 137. With `SYNAPSE_TEST_DB_DSN` set these tests encounter missing tables instead of exercising the intended schema; without it they skip and hide the mistake. Advance the targets to 136/135 and 137 respectively, and add an explicit 0137 round-trip/rollback test.

Focused validation completed:

- `go test -race ./internal/domain/findinglineage ./internal/usecase/findinglineage ./internal/usecase/assessmentsnapshot ./internal/usecase/findings ./internal/usecase/exploitation ./internal/infrastructure/persistence/memory`
- `go test ./internal/infrastructure/persistence/postgres -run 'FindingLineage|Migration0136'`
- Focused `go vet` across the changed domain/use-case/persistence/composition packages

The PostgreSQL integration tests were skipped locally because `SYNAPSE_TEST_DB_DSN` is not configured. Temporary regression probes reproduced the memory rollback, SCA upgrade split, CloudFormation root-stack failure, target-filter boolean error, skip replay ID mismatch, NFC namespace split/NUL collision, and lost partial-write classification; those probe files were removed before submission.

nghiadaulau (changes requested):
Reviewed head `ff6eb66` against base `codex/pr-811-cycle-snapshot` (`45b70cc`, the #814 head) with build, vet, unit tests, the Postgres suite on a fresh Postgres 17, the drift guards, and the web typecheck. Changes requested. This PR also inherits every open item on #779 and #814 and cannot merge before those land.

## Blockers

**1. The PR's Postgres tests migrate to the wrong version and never exercise 0136 and 0137.** `finding_lineage_backfill_repo_test.go:21,96` runs `goose.UpTo(136)` but `finding_lineage_backfill_runs` is created by 0137; `migration_0136_test.go:16-58` uses `UpTo(135)` and `DownTo(134)`, so `finding_identities` is never created. Actual: `relation "finding_lineage_backfill_runs" does not exist` and `relation "finding_identities" does not exist`; `TestMigration0136EmptyRoundTrip` passes without touching 0136. With the versions corrected (137/136/135) all four pass, so the code is fine and the tests are not. Lease fencing, resume, composite ownership, and the 0136 down guard have no executable Postgres coverage today, and the PR body's "Postgres, migration ... checks pass" is not backed by these tests.

**2. All 10 lineage Postgres tests fail on a fresh database with `lane manifest hash mismatch`.** Root cause is #779 (`scanrun.ComputeManifestHash` hashes nanosecond `time.Time`; `timestamptz` stores microseconds; `assessmentsnapshot/snapshot.go:291` re-verifies after the round trip). Reported there. Until it is fixed, Snapshot finalization from native runs, and therefore this PR's finalization shadow projection, is unreachable on Postgres.

**3. Two divergent planners turn the same legacy Finding into lineage, and the documented rollout runs both.** `shadow.go:98-184` (`project`) and `backfill.go:292-425` (`prepareFinding`) derive identities from different inputs. For an SCA Finding, the backfill builds `SCAFingerprintInputV1` from `occurrence.Ecosystem/Package/ComponentFingerprint` (`backfill.go:313-318`); the next Snapshot finalization re-projects every Finding (`shadow.go:86-94`, confirmed by `shadow_test.go:32`) from `AdvisoryID/ComponentFingerprint/DedupKey` only (`shadow.go:133`), `scaPURLFromPackage("", name)` fails (`sca_matcher.go:387`), the plan turns provisional, and `Correlate` creates a second identity plus a review candidate for every SCA Finding. SAST loses the judgment anchor the same way (`backfill.go:326-338` has it, shadow does not). The doc comment on `ShadowProjector` (`shadow.go:16-19`) says manual and offensive only; the code projects all kinds and hard-codes `OwnershipValid: true` (`shadow.go:102-108`). One planner with occurrence and judgment readers injected, used by both paths.

**4. Shadow projection runs inside the Snapshot finalization transaction and any single Finding aborts it.** `shadow.go:151-152` returns `ErrSecretMaterialRejected`, `:133-168` returns matcher `Build` errors, `:182` returns `Correlate` validation errors, and `assessmentsnapshot/service.go:186-190` aborts the whole `Finalize`. Reproduced: a secret Finding whose dedup key is `secret:<rule>:config/password=prod.yaml:1` (the shape `sca/findings.go:281` writes) surfaces `ErrSensitiveInput` from `SASTMatcherV1.Build`; a file path over about 505 bytes produces a `location` over 512 bytes that `Observation.Validate` rejects. Either one makes every `POST .../snapshots/finalize` for that assessment fail until the flag is turned off. The backfill degrades `prepareFinding` errors to a skip (`backfill.go:251`) but not `Correlate` errors (`:270`), so the tenant run terminates `failed` at that Finding on every resume. Separately, each `Correlate` calls `audit.Record`, which opens a second pool connection while the outer transaction pins one; 5,000 Findings means 5,000 lookups plus advisory locks inside one HTTP request, and on timeout the Snapshot rolls back while 5,000 audit rows stay committed. Map non-retryable per-Finding errors to skip records with `SkipRedactionRequired` or `SkipInvalidTrust`, and run projection as its own fenced pass per Snapshot.

## High

**5. Audit does not join the bound transaction, and the new paths create attestations for state that never committed.** `postgres/aup_audit.go:261-265` always calls `pool.Begin`. Inside `transactions.Run` at `service.go:193,321,356,375,409,444,461,494,511` and `backfill.go:260-269`, a `finding_lineage.correlated` entry naming identity X commits even when the item insert or the outer commit fails. In `withCreationBoundary` (`findings/service.go:62-116, 444-451`, `exploitation/service.go:82-127`), `finding.created` is committed even when `ProjectCreatedFinding` fails and the Finding rolls back (probe: 1 `audit_log` row after rollback). The retry loop at `backfill.go:260` retries `Correlate` three times inside an already-aborted transaction and appends one audit row per attempt. `fleet_audit_repo.go:73` already shows the transaction-joining pattern.

**6. A Cycle/default-Snapshot mismatch turns manual and offensive Finding creation into a hard 400 for shadow tenants.** `shadow.go:60-80, 99-100` returns `validation error: shadow Finding/Snapshot ownership mismatch` inside the creation transaction; `ErrNotFound` is tolerated (`:66, :74`) but this check is not. Postgres `GetCycleByAssessment` (`assessment_cycle_repo.go:106-118`) does not filter `archived_at`, so after a re-home the row chosen is arbitrary. Skip projection instead of failing the Finding.

**7. `selectBackfillTarget` admits any dimension whose producer OR finding kind matches.** `backfill.go:454` reads `if dimension.Producer != producer && dimension.FindingKind != findingKind { continue }`; the intended skip is `||`. Probe: dimensions `{sca, vulnerability}` and `{dast, vulnerability}` turn an SCA Finding into `ambiguous_target_identity`, a provisional identity, and a review candidate. Affects both the backfill and `ShadowProjector.project` (`shadow.go:110`), and the rows are immutable.

**8. Keyset cursor collation mismatch can skip Findings while the run reports `completed`.** `finding_lineage_backfill_repo.go:61-63` orders by `f.id COLLATE "C"` but filters `f.id > $2` in the database collation; `:223` compares the checkpoint the same way. On a non-C `lc_collate` database, ids differing in punctuation or case land on different sides of the boundary and never get lineage. `assessment_cycle_integrity_repo.go:39,312` collates both sides. Fix both comparisons and add a test with ids such as `a-b`, `ab`, `A1`.

**9. `findings.Service` behavior changed without tests.** `CreateAttributed` used to return `*ports.PartialWriteError` when the audit write failed after the Finding was persisted; it now returns a bare error, so callers can no longer distinguish "persisted, audit missing" from "nothing persisted". `Create` now goes through `withCreationBoundary` and resolves the engagement on every write. `findings/service_test.go` is byte-identical to the base; nothing asserts that a projection failure rolls the Finding back, that a non-allowlisted tenant skips projection, or that `ProjectCreatedFinding` is not called on the dedup-replay path. `ShadowProjector.ProjectCreatedFinding` has zero tests.

## Medium

**10. No uniqueness on identities.** `0136:26-27` is a plain index and `service.go:244-260` looks up then `:316` creates with no lock. Two concurrent finalizations (or a backfill during a finalization) insert two identities for one fingerprint; rows are trigger-immutable, so every later observation becomes a `fingerprint_collision` candidate forever. Take `pg_advisory_xact_lock` on the fingerprint or add a partial unique index excluding provisional identities.

**11. The review workflow is unreachable.** `ResolveCandidate` (`service.go:330`), `AppendOverride` (`:386`), `ApproveAlias` (`:420`), and `ConfirmManualCrossAssessmentLink` (`manual_workflow.go:192`) have no adapter, `api/openapi.yaml` is untouched, and the PR body says "HTTP wiring" passes while the diff has no HTTP change. Wire it or state that the surface is dormant.

**12. Dead `Matcher` interface.** `domain/findinglineage/matcher.go:35-76` (`Matcher`, `MatcherRequest`, `MatcherResult`, `VerifyConformance`) has zero implementations; `Descriptor()` exists only on SCA and SAST and is never called. The real contract is five `Build` structs dispatched by `switch kind` in two places.

**13. Producer-kind knowledge lives in five places, one of them SQL.** `backfill.go:602-615`, `:557-589`, the two dispatch switches, and the `producer_filters <@ ARRAY[...]` CHECK in `0137:27-29`. A new producer needs a migration.

**14. Postgres `AppendObservation` does not enforce the identity namespace; the memory twin does** (`postgres/finding_lineage_repo.go:76-101` vs `memory/finding_lineage_repo.go:86-92`). An override can link an observation whose producer, kind, or target differ from the identity's.

**15. `Correlate` replay is not idempotent on the trusted-producer-id path with incomplete identity fields** (`service.go:180` looks up replay only when `canonicalErr == nil`; `:452-462` always appends a fresh-ID observation). Probe: first call `matched/producer_id`, second call `conflict: source observation already exists in snapshot`.

**16. Backfill run lifecycle audit is best-effort** (`backfill.go:723-727` `_ = runner.audit.Record`, `:699-716` audit after `Finish` commits).

**17. Third copy of the fenced-run machinery** (`ports/finding_lineage_backfill.go:91-97` mirrors `ports/assessment_snapshot.go:100-106` and `ports/assessment_cycle_integrity.go:98-100`; runner `backfill.go:116-202` mirrors `assessmentsnapshot/backfill.go:186-260`). Raised on #814 as well; a generic `FencedRunStore` is due before the fourth copy in #816.

**18. Override and resolution hashes are self-integrity only** (`model.go:772-818` excludes the prior event's hash), so the "chain" rests on the FK and triggers alone.

## Low

- `0137:4-5` adds `findings_tenant_id_id_unique UNIQUE (tenant_id, id)` that no FK references (the item FK targets `findings(tenant_id, engagement_id, id)` from 0066); it builds a full unique index on the largest table under `ACCESS EXCLUSIVE` for nothing.
- Three sensitive-marker lists (`sast_matcher.go:498-506`, `sca_matcher.go:471-484`, `canonical.go` `sensitiveCanonicalKeys`) diverge inside one feature.
- `// ponytail:` marker at `backfill.go:473`; `_ = created` at `service.go:497` discards the idempotency signal.
- `AdvanceFindingLineageBackfillRun` recomputes four `count(*)` over all run items per batch, quadratic over a run.
- No `CHANGELOG.md` Unreleased entry for a new binary, two migrations, three metrics, and two `SYNAPSE_*` variables.

## Verified clean

`go build ./...`, `go vet ./...`, and `go test` on `domain/findinglineage`, `usecase/findinglineage`, `findings`, `exploitation`, `assessmentsnapshot`, `persistence/memory`, `platform/config`, `observability`, and both `cmd` packages pass. The hostile harness passes. `pnpm run typecheck` passes (no web files in this diff). Drift guards: same two pre-existing failures as the base; this PR fixes the mkdocs nav one. Dependency direction holds and consumer-side interfaces avoid a cycle between `findings`, `assessmentsnapshot`, and `findinglineage`. All 10 new tables carry `tenant_id` in the PK with composite FKs and `synapse_enable_tenant_rls`; every query filters `tenant_id` under `WithTenant`; the runner re-checks `source.TenantID`. Update, delete, and truncate triggers cover all 8 lineage tables; Down migrations refuse while rows exist. Lease fencing probes pass: stale token on commit, advance, and finish return `ErrConflict`, commit at exact expiry is refused, takeover rotates the token. Override CAS uses an advisory lock per source stream plus `expected_version` and `prior_event_id`. `RedactSecretProducerInputV1` rejects raw fields before `Build`; skip records hold only a hash and an enum code; no new HTTP routes.

Once 1 through 4 are fixed and 5 through 9 are addressed, and #779 and #814 have landed, re-request review.

nghiadaulau:
Đóng theo #814. PR này stack trên #814 và shadow-projection/backfill của nó phụ thuộc trực tiếp vào domain/ports/stores `assessmentcycle` + package `assessmentsnapshot` mà #840 đã xóa khỏi main, nên không thể rebase lên main bằng cách renumber migration đơn thuần (migration 0136 cũng đụng số với 0136_project_analyses_branch đã ship). Phần 'finding lineage' bản thân có thể có giá trị, nhưng nó đang bị buộc chặt vào assessmentcycle; nếu muốn, mình làm lại finding-lineage như một feature standalone reachable trên main thay vì cứu bản stacked này.

PR #816, opened 2026-09-04
feat: add assessment comparison relationship and closure workflows

nnatuan03:
Split from #811 as the focused Comparison, Relationship, and Closure slice. This PR is stacked on #815, so the diff excludes #779, Cycle/Snapshot, and Finding Lineage implementation details.

Scope
- Immutable lifecycle and neutral Snapshot Comparisons, stable item paging, manual match review, and failed-generation repair.
- Conservative historical relationship candidates plus review-only blocked repair plans.
- Preview and commit workflows for relationship changes, Cycle closure, reopen, immutable manifests, and deterministic reports.
- Ongoing SCA-to-Snapshot-to-Lineage-to-Comparison shadow coordination behind tenant gates.
- Rollout evaluator, metrics, UI, OpenAPI, Helm, Docker packaging, and operations runbook.
- Migrations 0138 through 0141.

Review fixes included
- Comparison verification readers are mandatory and concrete in API, worker, and backfill roots; effective retest decisions bind production inputs.
- Closure loads authoritative Finding, SLA, accepted-risk, waiver, and verification references, binds them into previews/manifests, and resolves every reference during report rendering.
- Closure text uses the shared strong sensitive-text validator with token, URL-userinfo, Basic-auth, and PEM tests.
- Shadow mode now creates ongoing artifacts while lifecycle reads remain disabled.
- Concurrent relationship CAS failures map to relationship_preview_stale and have a real Postgres contract test.
- ScanRun persistence uses #779 native lanes and fails closed on ID conflicts.

Validation
- Focused Go tests pass for all Assessment domains/use cases, SCA, memory/Postgres persistence, HTTP, observability, commands, API, and worker.
- Race detector passes across Assessment domains/use cases, memory persistence, HTTP, and observability.
- Assessment OpenAPI contract tests pass.
- Frontend: 397 tests pass; typecheck and production build pass.

Known base failures, reproduced outside this diff
- 15 Fleet routes are registered but absent from OpenAPI coverage.
- SYNAPSE_DATA_DELETION_ENABLED is missing from the existing configuration guide.
- Local Helm render test cannot run because helm is not installed.
- The egress-broker Unix-socket readiness test fails in this macOS environment; this PR does not modify that package.

Base chain: #779 mirror -> #814 -> #815 -> this PR. Retarget to main in order after prerequisites merge.


nnatuan03:
@pho-veteran @nghiadaulau — review requested for the Comparison/Closure slice.\n\nThis PR is stacked on #815. Please review only the single commit between the base and head.\n\nFocus areas: mandatory verification readers in API/worker/backfill paths, ongoing SCA shadow projection, authoritative closure references and report re-resolution, sensitive-text validation, relationship CAS for stale mappings, immutable reports, and rollout gates.\n\nMerge guidance: merge #815 first, retarget this PR to main, verify the diff remains limited to this slice, then merge. Known base-branch failures are separated from focused validation in the PR body.

nghiadaulau (changes requested):
Reviewed head `1b1a752` against base `codex/pr-811-finding-lineage` (`ff6eb66`, the #815 head) with build, vet, unit tests on 19 packages, the Postgres suite on a fresh Postgres 17, the drift guards, and the web typecheck, tests, and build. Changes requested. This PR inherits every open item on #779, #814, and #815 and cannot merge before those land.

## Blockers

**1. The PR's own Postgres migration tests are off by one and never exercise 0138 through 0141.** `migration_0139_test.go:20,73,200` runs `goose.UpTo(138)` then asserts `assessment_cycle_closure_manifests` exists (`exists=false`); `migration_0140_test.go:35,69` uses `UpTo(139)`; `migration_0141_test.go:35` uses `UpTo(140)`; `migration_0138_test.go:12-15` does `UpTo(137)`, `DownTo(136)` and passes without ever applying 0138. `TestMigration0139ClosureManifestGuards`, `TestAssessmentCycleScaleTargets`, and the misnamed `TestMigration0143RollbackGuard` (`assessment_comparison_repo_test.go:303`) fail with `column "active_closure_manifest_id" of relation "assessment_cycles" does not exist`. The rollback guards and FORCE RLS assertions for all four migrations have zero passing coverage. Fix each `UpTo` to the migration's own number and `DownTo` to N-1.

**2. Shadow coordination runs synchronously on the scan critical path and fails the scan after results are committed.** `sca/service.go:3217-3219` and `:2274-2276` return the error from `notifyAssessmentScanRun`, which runs after `SealScanRun`, the findings save, the evidence link, and `recordScannedImage` have all committed. `assessmentlifecycle/shadow.go:606-678` does Finalize (3 retries on conflict), snapshot listing, and comparison queueing inline. In the job path (`service.go:2015-2050`) the error marks the ScanJob `ScanFailed` and the idempotency guard forbids a re-run. For a shadow-enabled tenant, one queue insert failure or an RLS misconfiguration produces a failed scan whose findings and sealed run already exist, with no retry. Also ungated for every tenant: `persistAssessmentScanRun` (`:3153-3156`, `:2235-2238`) now returns errors that were previously discarded, so a lane failing `lane.Validate()` aborts the scan. `service_test.go` adds no test for `persistAssessmentScanRun`, the lane builder, or the observer error path. Make the post-seal projection best effort or enqueue it as a job the worker owns.

**3. Closure report GET re-resolves every manifest reference from scratch, once per reference, before checking the stored artifact.** `assessmentcycle/report.go:148-163` calls `ResolveAssessmentClosureReference` per reference; `closure_references.go:50-174` rebuilds the full artifact map each time (2 snapshot reads, 2 observation lists, 3 queries per source finding). The cache lookup at `report.go:164` runs only after the loop. A `read_only` user in a rollout tenant calling `GET .../closure-manifests/{id}/report` in a loop costs O(N²) round trips per request over N observations; the worker path (`report.go:271`) runs the same. Build the artifact map once per `Render`, and serve the stored artifact first when its `content_hash` is present.

**4. Closure commit binds references as of the preview time, so a decision recorded between preview and commit is neither bound nor detected as stale.** `closure.go:304` calls `buildClosurePreviewState(..., token.AsOfAt)` and `closure_references.go:127-131, 149-153` drop items with `At > asOf`. Reproduced with the real `ClosureDecisionReader` and a memory SLA store: preview at T0, `accepted_risk` transition at T0+1m, commit at T0+2m with the T0 token succeeds and the manifest omits the new event. `closure_test.go:23-31` wires a stub reader returning nil, so no test covers reference binding or drift. At commit, rebuild the reference set with `now`, compare to the token's binding hash, and fail with `closure_preview_stale` on any difference.

## High

**5. The "shared strong sensitive-text validator" does not exist; four substring denylists guard text that lands in immutable artifacts.** `closure.go:846-867` (`safeClosureText`), `relationship_change.go:479-490`, `assessmentcomparison/api.go:467-478`, `assessmentrelationship/service.go:547-558`. Probe: `safeClosureText` accepts `AKIAIOSFODNN7EXAMPLE ...`, `token=eyJhbGci...`, `ghp_0123...`, `xoxb-...`, `secret: hunter2`, and a Cyrillic-homoglyph `pаssword=hunter2`; the comparison and relationship variants additionally accept URL userinfo and PEM. At the same time it rejects ordinary reasons (`Basic retest completed`, `Password policy finding fixed`, `Client credentials rotation verified`) and surfaces them as `closure_reason_required`. Accepted text is written verbatim to `assessment_cycle_closure_manifests.reason` (trigger-immutable, `0139:201`), copied into the immutable report (`report.go:187`), and returned to every PermView user. One validator in `internal/platform/redact` with NFKC folding, key-value and prefix detectors, `redact.URLCreds`, and PEM headers; consider storing only a hash of the reason in the manifest so a leak can be scrubbed without breaking the chain.

**6. Audit does not join the bound transaction; closure and reopen audit entries commit before the state change, and relationship commit discards audit errors.** `postgres/aup_audit.go:262-263` uses `pool.Begin`; `closure.go:342` and `:446` record inside `executeRetained` before `Enqueue`, `CompleteAssessmentCycleRequest`, and COMMIT, so a deferred FK failure or a dropped connection leaves the cycle `open` while the chain records `assessment_cycle.closed` with a non-existent `manifest_id`. `api_service.go:692-699` does `_ = service.audit.Record(...)` for `CommitRelationshipChange`, so a reparent can commit with no audit row. Same root cause as #814 and #815.

**7. Relationship candidate routes and service bypass every rollout gate.** `router.go:583-586` registers four routes (including `POST .../generate` and `POST .../decisions`) without `requireAssessmentLifecycleRead`; `main.go:1546-1556` constructs `relationshipuc.NewService` unconditionally; `web/src/App.tsx` mounts the relationships route outside `AssessmentLifecycleRoute`. `assessmentrelationship/service.go:409-423` issues one `GetIdentity` query per observation of both snapshots before dedup, so a PermReview user outside the allowlist can generate candidates repeatedly at 2N queries each. Gate the routes, mount only when enabled, and batch-load identities.

**8. Manual match review is accepted on a superseded comparison.** `assessmentcomparison/api.go:232-244` has no status check and `service.go:347-349` `Replace` returns the successor without re-hashing. Reproduced: reviewer A confirms X (C1 superseded, C2 queued); reviewer B confirms Y on C1 with `expected_version=4` and gets 202; B's override is active; C2 generation then fails with `input_changed`. No comparison reflects both reviews. Reject review unless status is `complete` or `needs_review`, and recurse `Replace` on the successor.

**9. `input_changed` failures are unrepairable and one such record aborts the whole `--repair-failed` run.** `service.go:182-195` re-hashes input that can never match again; `backfill.go:102-107` returns on the first `Repair` error before any cycle is processed. Reproduced: three runs return `processed=0 repaired=0` while `attempts` climbs 2, 3, 4. Requeue a replacement for the current input and `continue`.

**10. The rollout evaluator accepts states the runbook forbids.** `assessmentrollout/gate.go:89-175`: `PhaseUIDefault` with `LifecycleUIDefaultEnabled=true, LifecycleReadEnabled=false` is allowed (contradicts `docs/guide/assessment-lifecycle-operations.md:133` and `config.ValidateAssessmentLifecycleRollout`); canary phases with reads and UI already enabled are allowed; 999 semantic mismatches out of 999 comparable items is a warning; dead-letter growth 500 with 1 repair is allowed. Thresholds 500/1000 are defined in `config.go:22-25`, `comparisonuc` constants, `gate.go:139-143`, and the backfill flag parser.

## Medium

**11. Re-queueing a dead-lettered pair replays the failed record and enqueues nothing** (`service.go:119-122`, `api.go:161-165`): HTTP 200, `status=failed attempts=2`, no job. Return 409 with a repair code or transition back to `queued`.

**12. `tenant_id` becomes a Prometheus label** (`observability/prometheus.go:83, 87, 92`), against the comment at `:33` that pool metrics deliberately carry no tenant labels. Enumerates tenant identifiers on `/metrics`; the generation histogram is tenants x modes x statuses x versions x bands x buckets.

**13. Preview-token signing is implemented twice in the use case and key derivation twice in `main`.** `relationship_change.go:444-463` and `closure.go:764-804` each carry an HMAC sign/verify pair with nonce and TTL while `internal/platform/worksign` exists; `main.go:1508-1514, 1524-1530` derive keys by hand and fall back to `rand.Read` per process when `SYNAPSE_MEASURE_CURSOR_SECRET` is blank, so two replicas outside Helm answer `relationship_preview_invalid` or `closure_preview_stale` to each other. Fail startup when the closure feature is on and the secret is blank.

**14. Cycle list rule lives in SQL and the memory twin does not implement it.** `postgres/assessment_cycle_repo.go:195-245` decides which comparison is the cycle's and derives `review_state` inline; `memory/assessment_cycle_repo.go:296` returns zero rows for any of the five new filters and never fills `ComparisonID`, `ComparisonStatus`, `ComparisonSummary`, `ActiveManifestID`. Put the classification in `domain/assessmentcomparison` and have both stores apply it.

**15. `referenceError` writes a fresh audit row on every failed report GET** (`report.go:275-281`, `time.Now()`, no idempotency key). Once a reference stops resolving, each PermView GET appends a chain link. Use `RecordOnce` keyed on manifest and reason.

**16. `assessment_comparisons` and `assessment_comparison_items` lack TRUNCATE guards** (`0138:129, 145`) while every other new table in 0139 to 0141 has one.

**17. Fourth copy of the tenant-backfill binary skeleton** (`cmd/synapse-assessment-comparison-backfill/main.go`, 225 lines, vs the three in #814 and #815). A shared runner in `internal/composition` taking a per-tenant closure.

**18. The hostile harness has no closure, comparison, or relationship cases** for 18 new routes (foreign cycle, comparison, candidate, manifest ids should return 404).

**19. Worker registers closure-report and comparison handlers and builds `ClosureDecisionReader` unconditionally** (`cmd/synapse-worker/main.go:171-205, 410-413`) while the API gates them on `AssessmentClosureEnabled`; `main.go:1282-1287` builds a second `snapshotuc.Service` that differs from `snapshotService` only by not having `SetScanJobStore`.

## Low

- Helm values named `api.assessmentCycleApi.*` render into `synapse.runtimeEnv`, which `worker.yaml` and `migration.yaml` also include (`_helpers.tpl:89-118`).
- Migration 0139 adds a self-referential CHECK (`active_closure_cycle_version = version`, `0139:237-247`); a rolled-back pre-0139 binary bumping `version` on a completed cycle violates it. Belongs in the runbook's rollback section.
- `time.Now()` at `report.go:279` and `assessmentcomparison/service.go:163` while the services take `ports.Clock`; the closure report is the one path advertised as deterministic.
- No `CHANGELOG.md` Unreleased entry for 18 routes, two binaries, four migrations, five `SYNAPSE_*` variables, and a UI page.
- 122 files and 20,717 lines in one commit across six independent slices (comparison plus 0138; closure plus 0139 and 0141; relationship plus 0140; shadow coordinator plus the SCA hook; gate, backfill, Helm, Docker, docs; web). Blocker 2 deserves its own review. Split along those seams.

## Verified clean

`go build ./...`, `go vet ./...`, and `go test` on 19 touched packages pass, including `sca`, all assessment domains and use cases, `httpapi`, `observability`, memory persistence, and four `cmd` packages. The hostile harness passes for the routes it covers. `pnpm run typecheck`, the touched vitest files (one load-induced timeout passes in isolation), and `pnpm build` pass. Drift guards show the same two pre-existing failures as the base. Layering holds and the cross-use-case orchestration sits in its own `assessmentlifecycle` package behind a one-method `ScanRunObserver`. All nine new tables get `synapse_enable_tenant_rls`; every new query runs under `WithTenant` with an explicit tenant predicate; request-body ids are resolved through tenant-scoped repositories or matched against the cycle's own members; `DecidePair` rejects cross-cycle pairs; closure re-checks tenant, cycle, ids, and hashes at render. All new routes sit behind `rt.authz` with PermView for reads and PermReview for closure, reopen, relationship commit, item review, and candidate decisions. Preview tokens are HMAC-SHA256 with domain-separated keys, `hmac.Equal`, and bound to tenant, actor, cycle, versions, binding hash, nonce, and expiry. Comparisons, items, manifests, references, candidates, plans, decisions, and reports have BEFORE UPDATE/DELETE triggers; manifest transitions are restricted. Report content is built from manifest fields plus the hash-bound summary with `GeneratedAt = SealedAt`. Paging is position-based over write-once items. Body limits and `DisallowUnknownFields` are in place. Re-closure after reopen produces manifest v2 and rejects the stale first token.

Once 1 through 4 are fixed and 5 through 10 are addressed, and #779, #814, and #815 have landed, re-request review.

pho-veteran (changes requested):
Reviewed exact diff `ff6eb6668c48e1f0950660f5b2aac762f6d01055..1b1a752de45dd87dca2067898ac87bfea8da244f`. Changes requested.

## High

**1. Shadow projection can turn a durably completed scan into an unretryable failure.** `internal/usecase/sca/service.go:2235-2275,3153-3218` seals the evidence, persists the ScanRun/findings/result, and only then synchronously calls `notifyAssessmentScanRun`; `internal/usecase/assessmentlifecycle/shadow.go:34-106` can fail while finalizing a Snapshot or queueing a Comparison. For queued scans, `internal/usecase/sca/service.go:2025-2068` records `ScanFailed` and returns nil to the durable worker, while its terminal idempotency guard prevents redelivery. A transient shadow-store or queue error therefore reports a failed scan although its primary artifacts committed, with no automatic way to finish the projection. Make the post-seal projection best-effort or place it behind a durable outbox/job, and test both direct and queued observer failures.

**2. Closure accepts a Comparison without proving that it still represents current governance inputs.** `internal/usecase/assessmentcycle/closure.go:585-605` selects the newest non-superseded lifecycle Comparison using only Snapshot IDs, mode, and status. The Comparison input built at `internal/usecase/assessmentcomparison/service.go:413-498` also binds relationship versions, active overrides, open match candidates, and verification decisions. A later retest/review/override can change those inputs without changing the default Snapshot IDs, yet preview and commit still accept the obsolete output. Recompute the current `GenerationInput` at both closure boundaries and require its hash to match `comparison.InputHash`.

**3. The Comparison input hash does not bind the observations that produce its items.** `internal/domain/assessmentcomparison/comparison.go:532-603` includes Snapshot hashes and governance IDs, but no observation IDs, content hashes, or observation-set revision; `internal/usecase/assessmentcomparison/service.go:554-587` reads the mutable observation membership only after the hash checks. Appending an observation to a finalized Snapshot therefore leaves the input hash unchanged, so `CreateQueued` replays an old Comparison that may omit a new high/critical finding. Bind a canonical observation set or immutable observation-set revision and fence concurrent appends.

**4. Stored Comparison output integrity is trusted rather than verified.** `internal/domain/assessmentcomparison/artifact.go:75-106,384-396` computes a content hash on completion, but `Validate` later checks only that the stored digest is syntactically valid. `internal/infrastructure/persistence/postgres/assessment_comparison_repo.go:307-373` loads items and summary without recomputing either the summary or output hash, and `internal/usecase/assessmentcycle/report.go:298-302` compares only the unchanged stored digest to the manifest. A probe that incremented `Summary.CurrentCount` while retaining `ContentHash` still passed `Validate`. Recompute the canonical summary/output digest after full loads and before closure/report consumption.

**5. Closure reporting bypasses the evidence-chain publishability gate.** `internal/usecase/assessmentcycle/report.go:34-48,114-221` has no evidence-verifier dependency and can generate or return an immutable report after the assessment evidence ledger has been truncated or tampered with. Existing reports fail closed on evidence integrity at `internal/usecase/report/service.go:208-234`; the new constructor wiring at `cmd/synapse-api/main.go:1297-1304` and `cmd/synapse-worker/main.go:196-202` omits an equivalent gate. Verify the relevant evidence chains before serving cached or newly rendered closure reports.

**6. Report reference validation is quadratic and runs even for cached artifacts.** For every manifest reference, `internal/usecase/assessmentcycle/report.go:148-164` calls `ResolveAssessmentClosureReference`; each call rebuilds the complete artifact map by rereading both Snapshot observation sets and all associated retest/SLA histories at `internal/usecase/assessmentcycle/closure_references.go:50-174`. The cache lookup occurs only afterward. A tenant viewer can repeatedly issue GETs against a naturally large manifest and force roughly O(N^2) database work per request. Build and validate the reference map once per render while preserving integrity validation before return.

**7. The new governed multi-store writes are not atomic in either supported persistence mode.** The in-memory `TenantTransactionRunner` at `internal/infrastructure/persistence/memory/tenant_tx.go:26-40` is only a mutex, so closure/reopen (`internal/usecase/assessmentcycle/closure.go:339-359,442-453`), Comparison replacement (`internal/usecase/assessmentcomparison/service.go:369-392`), and relationship decisions (`internal/usecase/assessmentrelationship/service.go:133-146,198-224`) retain already-written state when a later audit, queue, or retained-response operation fails. In PostgreSQL, `internal/infrastructure/persistence/postgres/aup_audit.go:261-287` always starts and commits a separate transaction instead of joining the transaction bound by `internal/infrastructure/persistence/postgres/tenant.go:32-39`; an audit can therefore survive a rolled-back business mutation, while relationship commit discards audit failures through `internal/usecase/assessmentcycle/api_service.go:692-699`. Add rollback-capable memory transactions/compensation and make audit participate in the bound PostgreSQL transaction.

**8. Relationship commits have no shared freshness fence with scans and Comparison completion.** `internal/usecase/assessmentcycle/relationship_change.go:366-423` excludes queued/generating Comparisons from its impact set and computes that set before the cycle row is locked later by `internal/usecase/assessmentcycle/service.go:470-473`. A generating worker can prepare against the old graph, the relationship commit can miss it, and the worker can then complete without ever being superseded; scan admission is likewise unfenced. Closure can consume that apparently current artifact. Lock/fence graph-dependent producers before impact calculation and include active Comparisons in invalidation semantics.

**9. Human-review boundaries are not enforced in the use cases.** The repository already defines the fail-closed `shared.IsMachineActor` predicate at `internal/domain/shared/principal.go:5-20`, but Comparison review (`internal/usecase/assessmentcomparison/api.go:210-300`), relationship decisions (`internal/usecase/assessmentrelationship/service.go:173-224`), relationship commits (`internal/usecase/assessmentcycle/relationship_change.go:174-290`), and close/reopen commits (`internal/usecase/assessmentcycle/closure.go:267-359,390-453`) accept reserved `agent:`, `llm:`, `mcp:`, `system:`, `bot:`, and service identities. Adapter RBAC is not an execution-layer separation-of-duties check. In addition, `assessmentrelationship/service.go:81-112,173-224` lets one reviewer provide any lowercase 64-hex imported-reference hash, create the resulting candidate, and confirm it themselves; a probe reproduced the self-confirmed artifact. Reject machine actors for human decisions and enforce independent review where caller-supplied evidence is the qualifying signal.

**10. Four divergent substring denylists do not keep secrets out of immutable review artifacts.** `internal/usecase/assessmentcycle/closure.go:846-867`, `internal/usecase/assessmentcycle/relationship_change.go:479-490`, `internal/usecase/assessmentcomparison/api.go:467-477`, and `internal/usecase/assessmentrelationship/service.go:547-557` recognize different credential forms. The Comparison validator accepts URL userinfo and generic `token=...` values; a probe reproduced the URL case. Closure accepts common token/key prefixes while also rejecting benign reasons containing phrases such as `Basic retest` or `Password policy`. Accepted reasons are persisted verbatim in immutable lineage events or closure manifests and reports. Use one shared redaction-aware validator and avoid storing raw free text where only attribution/integrity requires a digest.

## Medium

**11. Manual review can mutate a superseded Comparison and strand its successor.** `internal/usecase/assessmentcomparison/api.go:232-300` checks version but not status. After reviewer A supersedes C1 and queues C2, reviewer B can use C1's current ETag to resolve another candidate and append an override. `Replace` at `internal/usecase/assessmentcomparison/service.go:347-349` merely returns C2 without rebuilding it, so C2 later fails permanently with `input_changed` and no Comparison reflects both reviews. Require `needs_review` before any lineage mutation and disable review controls for superseded items.

**12. Failed Comparison recovery has two permanent-stranding paths.** Re-queueing an existing failed pair returns from `internal/usecase/assessmentcomparison/service.go:119-122` without enqueueing work, and `internal/usecase/assessmentcomparison/api.go:161-165` reports HTTP 200. Separately, an `input_changed` failure can never be repaired against its immutable old input hash, but `internal/usecase/assessmentcomparison/backfill.go:91-116` sends every failed record to `Repair` and aborts the whole tenant batch on the first repeated mismatch. A probe reproduced `Repair` returning `comparison generation input changed` after the failure. Requeue retryable failed records and replace/skip stale-input records without blocking later work.

**13. Closure cannot resolve a valid verification sourced only from an intermediate assessment.** Lifecycle generation considers every path Snapshot and can select such a decision at `internal/usecase/assessmentcomparison/service.go:502-542` and `internal/usecase/assessmentcomparison/retest_verification.go:40-79`. Closure then asks for that verification while `internal/usecase/assessmentcycle/closure.go:503-537` supplies only initial/final Snapshot IDs; `internal/usecase/assessmentcycle/closure_references.go:80-170` has no source finding from the intermediate assessment and returns not found. A valid multi-hop remediation can therefore never close. Resolve references across the frozen path or add a tenant-owned verification-by-ID lookup.

**14. Feature rollout wiring is inconsistent in both directions.** Relationship review is constructed and registered unconditionally at `cmd/synapse-api/main.go:1546-1556`; its routes at `internal/adapter/httpapi/router.go:582-587` omit the lifecycle-read tenant gate, so an off-by-default deployment still exposes generate/decision endpoints. Conversely, closure enablement is valid without Cycle API/dual-write according to `internal/platform/config/config.go:1211-1213`, but `cmd/synapse-api/main.go:1497-1540` only constructs/registers the Cycle API and closure endpoints inside those other flags. Enabling `SYNAPSE_ASSESSMENT_CLOSURE_REPORT_ENABLED` alone can initialize report services while exposing no closure paths. Give each surface an explicit feature and tenant gate and validate required wiring combinations.

**15. The default in-memory adapter does not implement the new Cycle list contract.** `internal/infrastructure/persistence/memory/assessment_cycle_repo.go:271-351` silently drops every row whenever one of the new assessment/comparison/staleness filters is present and never populates `RootSnapshotID`, `CurrentSnapshotID`, Comparison, active-manifest, or scan-staleness fields. PostgreSQL derives all of those at `internal/infrastructure/persistence/postgres/assessment_cycle_repo.go:191-249`. With no `SYNAPSE_DB_DSN`, valid list requests therefore return empty or incomplete results rather than matching production behavior. Implement the same projection/filter semantics in memory or fail closed when the feature requires PostgreSQL.

**16. Migration tests apply the preceding migration instead of the migration under test.** `migration_0138_test.go:12-16` never applies 0138; `migration_0139_test.go:20-48,63-74`, `migration_0140_test.go:35-49,60-70`, and `migration_0141_test.go:35-58` apply 138/139/140 while asserting objects from 139/140/141. `assessment_comparison_repo_test.go:301-326` is also mislabeled 0143 and seeds 0138 tables after stopping at 0137. With a PostgreSQL DSN these tests either fail before exercising the guards or pass vacuously. Apply N, then roll back to N-1 for each migration.

## Low

**17. Whitespace-normalized relationship commands return success without executing.** `internal/usecase/assessmentcycle/relationship_change.go:293-296` validates a trimmed local copy, while preview signs the raw command at `:153-166` and commit switches on the raw command without a default at `:201-256`. A probe using `" select_head "` returned 200 with the old selected head and no mutation. Normalize once before signing/execution or reject noncanonical command text.

**18. Added comments violate the repository's explicit `ponytail:` prohibition.** The prefix appears in exact-diff additions at `internal/domain/assessmentclosure/closure.go:451`, `internal/usecase/assessmentcomparison/api.go:503`, `internal/usecase/assessmentcomparison/service.go:780`, `internal/usecase/assessmentcycle/relationship_change.go:282`, and `internal/usecase/assessmentcycle/api_service.go:743`. Remove only the prohibited prefix.

## Verification

`go build ./...` passed. Focused `go vet` passed. `go test -race` passed for the assessment closure/comparison/relationship domains, their use cases, lifecycle/SCA integration, and memory persistence. Focused PostgreSQL tests compile but integration cases skip because `SYNAPSE_TEST_DB_DSN` is unset. `go test ./...` reached only failures reproduced on the exact base (Fleet OpenAPI debt, two agent telemetry tests, and config-doc drift) plus the local missing `helm` tool before timing out; the focused exact-diff packages pass. Frontend checks could not run because `pnpm` is not installed. All temporary probes were removed and the reviewed worktree is clean.

nghiadaulau:
Đóng theo #814/#815. Đây là phần đuôi của epic assessment-lifecycle (được tách từ #811 thành #814/#815/#816), base `codex/pr-811-finding-lineage` (không phải main), 122 file/+20k. Nó re-add domain/usecase/repo/port/handler `assessmentcycle` — đúng những file #840 đã xóa khi gỡ subsystem dead code — nên mâu thuẫn với quyết định đã chốt và không thể land độc lập lên main. Nếu muốn assessment comparison/closure, cần thiết kế lại reachable trên main.

PR #818, opened 2026-09-05
fix: close the operational review of the product (#817)

nghiadaulau:
Closes the operational review in #817. Every item was reproduced before it was fixed, and every fix ships with a test that fails without it.

Seven themes, developed on separate branches and merged here, plus three rounds of independent review. 71 commits, 288 files.

## Security of the product

- **Cross-tenant user management.** `POST /api/v1/users` took `tenant_id` from the request body and never compared it with the caller's tenant, and `GET /api/v1/users` had no tenant predicate, so a tenant admin could mint an admin API key in another tenant and enumerate every tenant's users. User reads and writes now carry the tenant explicitly in the port, and cross-tenant provisioning is confined to the bootstrap platform principal, which is the only identity that legitimately seeds a new tenant's first admin. Verified live: a tenant admin now gets `403 forbidden: user management is confined to the caller's own tenant`.
- **The global vulnerability-source registry** has no tenant column and decides which advisories every tenant's detection reads. Mutating it now requires that same platform principal on top of the administer permission. The check reads the authenticated principal directly rather than through `PrincipalFrom`, whose operator fallback keeps attribution coherent but would fail open as an authorization test.
- **`allow_private_network`** turned the control plane into a probe of the operator's own network. It now needs a deployment opt-in, `SYNAPSE_VULNERABILITY_SOURCE_ALLOW_PRIVATE_NETWORK`, default off. Verified live: `400 allow_private_network requires the deployment to enable private-network vulnerability sources`.
- **Request bodies** on the human plane were unbounded in fourteen handlers. A middleware applies a 1 MiB ceiling by route pattern, with larger explicit ceilings for every route that accepts an upload: source publish, engagement and project creation, coverage upload, bundle import, SARIF, SBOM, evidence and threat model. An unknown pattern takes the default, so a new route is bounded the moment it is registered. Verified live: a 2 MiB body on a 1 MiB route is rejected.
- **Connection lifetimes.** Listeners set write and idle timeouts alongside the existing header timeout. The write deadline is generous because in Go it also covers reading the request body, and the largest body the API accepts is a 600 MiB source archive.
- **Audit atomicity.** The audit log opened its own transaction even inside a caller's, so a rolled-back write left a committed row claiming it happened. It now appends on the caller's bound transaction. The VEX apply and approval decision paths propagate the audit error instead of discarding it. The regression test finds one orphan row without the fix and none with it.
- `golang.org/x/image` moves to v0.45.0, clearing the one vulnerability govulncheck reported as reachable.

## Database-enforced tenancy

Thirteen tables carried a `tenant_id` but had no row level security, and none of their repositories used `WithTenant`: `projects`, `project_analyses`, `project_analysis_hotspots`, `project_hotspots`, `project_hotspot_review_events`, `project_issues`, `project_issue_review_events`, `quality_gates`, `quality_profiles`, `threat_models`, `agent_sessions`, `agent_approvals`, `agent_plans`. Every one now routes through a tenant-bound transaction while keeping an explicit predicate. Two stores that dropped the predicate when the tenant was empty now fail closed, and an approval decision can no longer be applied across tenants.

**Migration `0129` inverts this project's usual order and says so in its header.** Forced RLS takes effect on commit and the runtime role does not own the tables, so a previous-release binary on the raw pool would see zero rows. Deploy the binaries first, confirm every replica and worker is on that build, then migrate. The phase-1 code is correct with or without the migration, so holding `0129` for the following release is a valid choice. Isolation is proved under a real `NOSUPERUSER NOBYPASSRLS` role, and a round-trip test per converted store proves normal operation still works.

## Scanner accuracy

The engine discarded its truncation flag and applied one global 500-finding cap, so vendored JavaScript consumed the whole budget before first-party code was reached.

| Corpus app | findings in vendored paths, before → after | previously missed, now detected |
|---|---|---|
| railsgoat | 500 → 0 | SQL injection `users_controller.rb:29`, MD5 `user.rb:45`, Marshal `password_resets_controller.rb:6` |
| Vulnerable-Flask-App | 485 → 0 | SQL injection `app.py:265` |
| dvpwa | 496 → 0 | SQL injection `student.py:45` |
| WebGoat | 30 → 0 | budget freed for first-party code |

A later review round corrected the skip that produced those numbers. Taking every `.js` under `static/`, `assets/` or `public/` also dropped the application's own scripts, silently, which is worse than the noise it removed. A web asset in those trees is now skipped only when the file itself declares it third-party, and files excluded by policy are counted separately from a scan cut short by a budget. railsgoat settles at 44 findings with 19 files excluded, and every named detection above still fires.

Behind those numbers: a per-file cap before the global one, vendored, minified and generated-file skipping (the old `.min.js` entry never matched because the lookup used `filepath.Ext`), a bounded ten-line look-back so a query built on the previous line still resolves at its sink, thirteen tightened rules, and widened credential key shapes so `cookieSecret:` and `config['SECRET_KEY']` are found. `python-assert-string-literal`, reported by a user as firing on `assert "X" in y`, now matches only a whole-condition literal.

## CLI

`synapse-cli quality` crashed on any repository containing HTML or CSS, because the use case rejected analyzer kinds it does not emit itself. Root cause was wider than the issue recorded: the CSS pack emits `maintainability` as well as the HTML pack's `security`. All five kinds now map, an unknown kind degrades instead of aborting, and a table test covers the full vocabulary. `--offline` now disables every egress-capable resolver rather than only the live advisory source.

## First install

The README's full-stack command failed on a clean checkout. The two compose stacks shared a project name and a volume with different superusers, so running one then the other failed in `postgres-init`. The dashboard container never started, because corepack downloaded a different pnpm as the unprivileged user and then hit `EACCES` on the bind mount; the full stack now serves the built dashboard through nginx, which also drops the image from 693 MB to 58 MB. The dev stack handed the API a superuser, which it correctly refuses to serve with, so it now creates the app role the way the full stack does. A dev values file makes `helm lint` and `helm template` pass without weakening the production defaults.

## Product surface

`GET /api/v1/capabilities` reports each optional subsystem, whether it is enabled, and the `SYNAPSE_*` switch that controls it. User management gains update, disable, enable and key rotation, all audited, with a guard against disabling or demoting a tenant's last enabled admin. `PUT /api/v1/engagements/{id}/status` used to return 404 and now works. The fifteen fleet operations that were registered but undocumented are described in the spec, so the coverage ratchet is clean without growing the debt file.

**Breaking:** engagements and projects now serialize snake_case through explicit view types. The web client is updated in the same change, and a contract test pins the fixtures against `resource_view.go` so the mocks cannot drift from the server again. Before that fix the engagements page did not merely render blanks, it crashed.

## Dashboard

Irreversible transitions are confirmed in a modal. A disabled subsystem renders as disabled in the sidebar and as an explanatory state in the page body, naming the switch, instead of a raw `HTTP 404` with a Retry button that 404s again. The findings table gained page sizes, sortable columns, URL-backed filters, and a search that actually indexes the package, dependency path, CVE and priority it advertises. The engagement hero collapses once a scan finishes and the tab bar sticks, so content no longer starts below the fold at 1024 px. Severity and scope badges now pass AA in light as well as dark. There is a real 404 page, a toast and live-region utility, arrow-key tab navigation and a skip link, and the hooks-order bug that logged a React internal error on every engagement load is gone.

## Test posture

`go test ./...` passes on 318 packages against a fresh Postgres. The only failure is `internal/infrastructure/sandbox`, which needs bubblewrap privileges the CI container does not grant and which fails identically on `main`. Web: typecheck, lint, 471 vitest tests and the production build all pass.

Four suites were red or flaky on `main` for reasons unrelated to what they test, and are fixed here: the telemetry shipper fixtures predated a strict content-type check; the acquire and gitdiff suites shelled out to git and inherited the developer's global hooks; and four Postgres tests created cluster-global probe roles under fixed names, so one aborted run poisoned every later run in every other database on the server.

Three new inventory tests guard the class of defect this epic found rather than the instance: every route must pass the authorization chokepoint, the public-route exemption list may not go stale, and the hostile harness coverage of tenant-scoped routes is ratcheted at its current 94 of 139.

The inventory is parsed with `go/ast`, so a registration written across several lines, one written as `mux.Handle`, or a comment that merely mentions the guard are all read correctly, and a pattern the parser cannot model is an error rather than a silent skip. Coverage is matched on the whole pattern with its parameters filled in; an earlier draft compared the fixed prefix before the first parameter, which counted every route under an existing prefix as covered on registration and reported 115 where the real figure was 33. A generated cross-tenant sweep over every tenant-scoped GET route closes most of that gap.

## Verification performed

Full Go suite on Postgres 17; the documented API flow end to end on the in-memory store, on Postgres with a separate migration and runtime role, and in the full compose stack; the four security fixes exercised against a running server; the corpus rescanned and compared against the pre-change baselines; Helm rendered and linted; the dashboard driven with Playwright in both themes at three viewports.


nghiadaulau:
## Review round: three independent reviewers, nine fixes

Two subagent reviewers and Codex reviewed this branch. Every finding was reproduced against the code before it was acted on. Nine were confirmed and fixed here; each fix ships with a test that fails without it.

### Blocking, now fixed

**1. A default-tenant admin could seize the platform principal.** The bootstrap admin is stored with an empty `tenant_id`, and both the Postgres predicate and the in-memory twin normalize an empty tenant to `default`. The row was therefore a member of that tenant's roster and reachable through the ordinary tenant-scoped lookup, so its admins could rotate the operator's API key, read the plaintext from the response, and present it. That key authenticates as `operator`, which is the platform principal the vulnerability-source guard added in this very PR tests for. Disabling or demoting the same row locked the deployment operator out.

Update, disable and key rotation now refuse a target of `operator` for anyone but the bootstrap principal. The credential is owned by `SYNAPSE_API_TOKEN`. `TestDefaultTenantAdminCannotSeizeTheBootstrapPrincipal` reproduces all four mutations and fails without the guard.

**2. The new body limits broke every real upload.** An `http.MaxBytesReader` nested inside another one can only tighten the bound. Three routes read a multipart upload behind their own larger reader while sitting on the 1 MiB default, so engagement creation with a source package, project creation from an archive, and analysis start with a coverage report were all capped at 1 MiB. Those routes now carry the ceiling their handlers were written for, the JSON branch of engagement creation states its own bound, and two tests keep the class closed.

**3. The private-network gate had three ways past it.** It fired on create and update only, so a draft connection test reached an RFC1918 endpoint with the switch closed, and a row stored while the switch was on kept working through test, re-enable and the sync scheduler. The gate moved into the provider registry, the one point every caller resolves a source through. Disabling and archiving a grandfathered source still work.

### Also fixed

**4. Event streams were cut off at the listener write timeout.** Go sets `WriteTimeout` once at the start of the response. Both stream handlers bound themselves at 30 minutes against a 10-minute listener, and they ignore write errors, so the failure was silent. Each now clears its write deadline; a control test proves the deadline really truncates without it.

**5. The harness coverage ratchet measured nothing.** It truncated each pattern at its first parameter and asked whether the harness contained that prefix, so two prefixes accounted for all 116 "covered" routes and any new engagement route counted as covered on registration. The honest figure was 33. Coverage is now matched on the whole pattern, and a generated cross-tenant sweep over every tenant-scoped GET route raises it to **94 of 139**. The sweep found a `500` on `GET /api/v1/projects/{key}/measures` and a nil dereference in the agent handlers, both fixture gaps, now wired.

**6. The route inventory was a regex.** It is parsed with `go/ast` now. A multi-line registration, a `mux.Handle`, and a comment mentioning the guard are all read correctly, and a pattern the parser cannot model is an error rather than a silent skip. Fixture tests cover each escape.

**7. The bound audit append could not retry.** A unique violation aborts the enclosing transaction, so the retry hit `25P02` instead of the `23505` it dispatches on: the fork guard was dead and the caller's write was lost. The append runs inside a savepoint now. Seven assessment-cycle sites that discarded the audit error propagate it, which is only safe because of the savepoint.

**8. VEX apply and approval decide claimed atomicity they did not have.** Each VEX status change committed separately, and an approval committed before its audit, so a failed append returned an error for a decision that already stood. Both run inside one tenant transaction now, and both keep working without one.

**9. The Compose proxy capped uploads at 1 MiB and streams at five minutes.** nginx now allows 600 MiB, waits past the stream ceiling, passes events unbuffered, and sets `X-Content-Type-Options`, `X-Frame-Options`, `Referrer-Policy` and a strict content security policy.

### Also in this round

The advisory chain lock is one transaction-scoped key per deployment, so a bound append holds it until the caller commits. The rule is to audit last in a bound transaction, and a new guard walks each transaction closure and fails when a repository call follows the audit call. `GET /api/v1/capabilities` staying readable by the read-only role is now a documented decision in the security guide rather than an unexamined default.

A concurrency flake also surfaced: three test packages migrate the same database, and `synapse_enable_tenant_rls` is not idempotent under concurrency, so a full-suite run failed with a duplicate policy while the package passed alone. The tests use `MigrateLocked`, which production already uses.

### Verification

`go test ./...` on a fresh Postgres 17: **318 packages pass**. The only failure is `internal/infrastructure/sandbox`, which needs bubblewrap privileges this container does not grant and fails identically on `main`. Web typecheck, 471 vitest tests and the production build pass. `go vet ./...` is clean. `golangci-lint` reports six issues, all in files this branch does not touch and all present on `main`.

### Not fixed, deliberately

`GET /api/v1/capabilities` stays at the view permission, because the dashboard uses it to explain a disabled feature to every role. Migration `0129` has still only been exercised against a database migrated from empty; the phasing note in its header stands, and it wants a run against a restored production snapshot before it ships.

nghiadaulau:
## Second review round: the QA pass found a P1 in the scanner

The independent QA reviewer ran mutation testing against the branch and found five defects that survived the first round. All were reproduced before being acted on, and all are fixed here.

### The serious one

**Static analysis was dropping first-party code and calling the scan complete.** The vendored-asset skip took every `.js` under `static/`, `assets/` or `public/`. That is exactly where a Flask or Django application keeps its own scripts. A hand-written `static/js/app.js` was dropped along with the copied jQuery, and the minified probe judged any file of three or more lines by average line length, so a short file holding one long constant went too. On a fixture, seven findings per asset file and two per short file vanished with `Truncated` still false, which is the worst shape a scanner failure can take: a clean tree and an unscanned one read identically, and `--fail-on` passes either way.

A web asset in those trees is now skipped only when the file itself says third-party: a distributed-library banner in its head, a vendor or build directory in its path, or a line long enough to be build output. The banner test reads inside block comments, which is where a Rails asset-pipeline copy declares its licence.

The trade holds. On railsgoat the vendored noise stays at 44 findings, against 491 for a name-only narrowing, and all five previously-missed corpus vulnerabilities still fire.

| corpus app | findings | files skipped as vendored | truncated |
|---|---|---|---|
| railsgoat | 44 | 19 | no |
| Vulnerable-Flask-App | 73 | 0 | yes |
| dvpwa | 6 | 2 | no |

**Completeness is now honest in both directions.** A file excluded by policy is counted in a new `SkippedFiles` field rather than folded into `Truncated`, which would be true for every real repository and therefore mean nothing. `Truncated` is no longer set when a scan lands exactly on the finding budget with the last file fully read. The SCA scan reads the reporting form and turns both into a source warning. Nothing consumed that flag before, which the reviewer noted separately.

### Rule regressions and false positives

Two rules had stopped matching their highest-signal shape, because the bare-argument pattern accepted a plain identifier only. `exec.Command(args[0], args[1:]...)`, where the binary itself is attacker-controlled, and `w.Write([]byte(s))`, the idiomatic Go response write, both match again.

Three cross-line false positives are closed. `||` is logical-or everywhere but SQL and PL/SQL, so `opts.sql || DEFAULT_SQL` was reported as SQL injection at high severity. The ten-line look-back reached over a function boundary and attributed one function's assignment to another's sink. And `redirect_to <model>`, the canonical safe Rails idiom, was reported as an open redirect.

### The last-admin guard raced

It counted a tenant's other enabled admins and then wrote, with nothing in between. Two concurrent demotions each saw the other admin still enabled, both passed, and the tenant was left with nobody who could administer it. A two-goroutine test reproduces it within a hundred iterations on every run.

The count and the write now run in one tenant transaction, the roster read takes a row lock through a new capability interface, and an in-process mutex covers the stores that have no transactions. Both halves are needed: the mutex alone does not serialize two replicas, the row lock alone does nothing for the in-memory store.

### Two guards were passing on things they did not check

The body-limit tests all set `r.Pattern` by hand, so they said nothing about the middleware order that makes the ceiling work. Reversing the two middlewares silently caps every upload route at 1 MiB again and left the whole package green. The new test drives the real handler end to end with a 2 MiB multipart body and fails on the reversed order.

The engagement handler test enumerated nine response keys, so renaming `authorized_from` broke the web client and passed. The wire contract now pins the complete key set of every view type against the struct tags, in both directions.

### Test hygiene

The probe-role cleanup borrowed the test's pool, and Go runs a test body's deferred pool close before its cleanup callbacks, so every drop ran against a closed pool and the error was discarded. Giving each role a per-run name had removed the self-healing drop that a fixed name gave the next run, so the leak became permanent at six cluster-global roles per run. The helper owns the drop now and opens its own connection for it. The promotion tests used fixed fingerprints, which are unique for the life of the database, so a second run against the same database failed on a conflict unrelated to the test.

### Verification

`go test ./...` on a fresh Postgres 17: **318 packages pass**, and the role count is unchanged before and after. The persistence package passes twice in a row against the same database. The only failure is `internal/infrastructure/sandbox`, which needs bubblewrap privileges this container does not grant and fails identically on `main`. Web typecheck and 471 tests pass. `go vet ./...` is clean.

Checked against a running server: a default-tenant admin gets `403 forbidden: the bootstrap operator is managed through SYNAPSE_API_TOKEN` on rotate, disable and demote, while rotating an ordinary user in the same tenant still returns 200 and the environment token still authenticates. A 4 MiB multipart source package on engagement creation returns 201, and a 1.5 MiB JSON body on the same route is still refused. The command-line scanner on a fixture now reports the `static/js/app.js` findings and the indexed `exec.Command` that both used to be missed.

### Not fixed

`GET /api/v1/capabilities` stays at the view permission, documented as a decision in the security guide. There is still no drift guard tying a new optional subsystem to a capability-catalog entry. Migration `0129` has only been exercised against a database migrated from empty and wants a rehearsal against a production snapshot before it ships.

nghiadaulau:
## Three scanner commits were missing from this branch

The QA reviewer noticed that the write-up named the secret scanner as changed while the diff touched only the pattern engine and the git-diff helper. That was accurate: three commits from the scanner slice never made it into the epic merge, so this pull request claimed a widening it did not ship.

They are merged now. The credential rules matched their keyword as a whole word, which missed the two shapes real configuration uses: camelCase, as in `cookieSecret`, and a bracketed config key, as in `app.config['SECRET_KEY_HMAC']`. Both the pattern engine and the secret scanner are affected, and both are fixed, with the value guards untouched so precision is unchanged. The SSRF and credential rules also stop firing on lookups and version pins.

One conflict needed a decision. The slice dropped the double-pipe operator from the cross-line evidence set outright; this branch splits it by language instead, because the operator really is string concatenation in SQL and PL/SQL and only logical-or elsewhere. The split is kept, so neither language loses.

Every slice branch is now an ancestor of this one, which was verified for all nine.

`go test ./...` on a fresh Postgres 17 after the merge: **318 packages pass**, and the cluster role count is unchanged. The only failure remains `internal/infrastructure/sandbox`, on bubblewrap privileges this container does not grant, identically to `main`. `go vet ./...` is clean.

PR #855, opened 2026-09-08
feat: complete event-time correlation and governed response

### LINKED ISSUE #676: C2 · Event-time correlation engine (detections + timeline → incidents)
> Part of #594 · Phase **C** · sub-item **C2**. Depends on C1 (Incident) + B (endpoint State Timeline).

## Goal
An **event-time** correlation engine that folds detections + endpoint transitions into incidents: watermark + allowed-lateness, out-of-order + late-event revision, a stable dedupe key, bounded correlation graph, and anti-storm — never ingest-order.

## Scope
Pure-domain correlation over the State Timeline + detections producing `IncidentEvent`s (Created/DetectionAttached/...); windowing by event time; dedupe; storm suppression with a visible coverage note.

## Acceptance
Out-of-order and late events correlate deterministically into the same incidents; a storm is suppressed without silently dropping (coverage-honest); dedupe is stable.

### LINKED ISSUE #680: C6 · Governed response end-to-end (telemetry-verified post-condition)
> Part of #594 · Phase **C** · sub-item **C6**. Depends on C1 + the response seam (#425) + B (telemetry to verify against).
## Goal
≥1 **governed response end-to-end** with the full saga (Proposed→AwaitingApproval→Approved→Issued→Claimed→Executing→CommandApplied→Verifying→VerifiedSucceeded|...→RolledBack) — CommandApplied != VerifiedSucceeded: the post-condition is **verified via telemetry**, with target-fingerprint (never a bare PID) and an agent execution journal persisted BEFORE the side effect.
## Scope
Extend the response saga with the verify-via-telemetry post-condition + ResponseAttempt journal + ReversibilityClass; wire ResponseRequested/ResponseVerified onto the incident (C1).
## Acceptance
One governed response runs e2e with idempotent execution and a telemetry-verified post-condition; a reversal has its own verified post-condition; VerificationUnknown on insufficient coverage is representable.

pho-veteran:
## Summary

Completes the two remaining EPIC 594 control-plane capabilities from #676 and #680: deterministic event-time correlation into canonical incidents, and governed endpoint response whose success is established from independently accepted target telemetry rather than command output.

### Event-time correlation and canonical incidents

- Adds a two-phase correlation pipeline. A bounded source page is first materialized into an immutable snapshot ordered by detection recording position; a second phase consumes staged signals globally by `(occurred_at, signal_id)`.
- Persists snapshot high-water, source and staged cursors, policy digest, provisional maximum event time, finalized watermark, assignments, and active sessions. Intermediate pages retain the finalized watermark and active sessions; finalization computes `MaxObservedAt - AllowedLateness`, advances the completed source frontier, prunes finalized sessions, and deletes only the completed staged snapshot.
- Makes allowed-late revision, too-late signals, and anti-storm suppression explicit outcomes. Exact deduplication and deterministic bridge selection make results independent of ingest order and page size.
- Adds immutable merge edges. Physical incident logs and assignments stay append-only, while logical reads and later writes resolve aliases to a canonical root and expose merge provenance.
- Bounds page size, causal timeline fanout, active sessions, and transaction work. Saturation fails visibly without advancing correlation state.
- Commits incident events, merge provenance, assignments, active state, staged deletion, and checkpoint movement through the same tenant transaction.

### Governed endpoint response

- Implements approval, signed dispatch, durable endpoint journaling before side effects, command-applied state, independent telemetry verification, and separately governed reversal.
- Addresses process actions by stable asset/process fingerprints. The Linux actuator uses pidfds and descriptor-pinned executable metadata; duplicate delivery replays durable journal state instead of repeating an effect.
- Adds immutable target-evidence receipts built from accepted target-primary telemetry. Each receipt binds the action, attempt, challenge, source agent/session, bounded timeline, coverage, completeness, and digest.
- Keeps the independent observer on its own host identity. The observer signs the immutable receipt binding and does not relabel observer-local telemetry as target telemetry. Verification reloads the receipt and derives the post-condition from target timeline and coverage.
- Maps missing, stale, ambiguous, gapped, sampled, truncated, saturated, or incomplete telemetry to unknown. `CommandApplied` never implies `VerifiedSucceeded`.
- Extracts observer signing/outbox/transport orchestration from the agent command into composition and infrastructure packages.
- Adds a distinct `SYNAPSE_DB_HALT_WRITER_DSN` database capability. Runtime DML on halt fences and dispatches is revoked; the dedicated writer atomically advances the fence, inserts dispatch, and persists mandatory audit intent.

### Persistence and deployment

- Adds migrations 0138-0143 for attempt invariants, execution runtime state, correlation checkpoints/staging, incident merge provenance, target-evidence receipts, and halt-writer privileges.
- Implements memory/PostgreSQL parity for replay validation, tenant isolation, append-only facts, bounded reads, and transaction rollback.
- Wires API, agent, migration command, Helm values/schema/templates, systemd, Kubernetes, and configuration documentation.

## Verification

- Focused correlation suites pass after rebasing onto current upstream `main`, including out-of-order/page-size equivalence, dedupe, allowed-late/too-late outcomes, deterministic bridges, source high-water isolation, and saturation with zero state advance.
- Focused response, receipt, verification-ingest, telemetry-ingest, observer-runtime, and native actuator suites pass after the rebase.
- PostgreSQL integration covered migrations from zero, correlation atomic rollback, target-receipt immutability, tenant isolation, and distinct admin/runtime/halt permissions.
- Helm lint/render/schema/chart tests, changed command builds including `CGO_ENABLED=0`, web checks, and focused race/native tests passed during implementation.
- Final disposable AWS acceptance ran in account `116378167742`, region `us-east-1`, on Amazon Linux. Correlation domain, correlation use case, governed-response E2E, and the root-only native stop/reversal saga all returned zero. Uploaded binaries and result logs were checksum-verified.
- Ownership tags were revalidated before teardown. The run-owned EC2 instance/volume, security group, IAM role/profile, versioned artifact/backend buckets, and DynamoDB lock table were removed; residual tagged compute, volumes, ENIs, snapshots, security groups, IAM objects, buckets, and lock table were verified absent.

## Operational notes

- The final AWS run used an issue-specific compiled acceptance harness on a real AWS Linux host; it was not a full live API + RDS + enrolled-agent topology. PostgreSQL behavior was exercised separately against a real PostgreSQL instance with distinct roles.
- Hardened Linux response fails closed when pidfd or required process metadata is unavailable. Non-Linux builds expose the unsupported adapter rather than approximating process control.

Closes #676
Closes #680


nghiadaulau:
## Deep review — verdict: blocked (4 must-fix), CI-green notwithstanding

The change is large and well-structured, and CI is green, but green here only means the existing tests do not reach these paths. Four issues are must-fix before merge; each was verified against this PR's revision.

### Blockers

1. **Migration `0138` is not backward-compatible (breaks the still-running old binary during migrate-before-deploy).**
   `migrations/0138_response_attempt_invariants.sql:12` adds `reversibility_class TEXT` with no default, then `:28` `SET NOT NULL` + `:29` a `kind`-dependent CHECK. The old binary's INSERT omits the column (`internal/infrastructure/persistence/postgres/response_repo.go:52` at current main), so once `synapse-migrate` runs before the binary rollout (the repo's phased-deploy rule), every governed-response apply on the old `synapse-api` fails a not-null violation. A single scalar DEFAULT can't satisfy the kind-dependent CHECK. The `0138` test only migrates the new schema, so CI misses it.
   **Fix:** expand/contract. In `0138` keep `ADD COLUMN reversibility_class TEXT` (nullable) + the backfill only; drop `SET NOT NULL` and the CHECK (or make the CHECK `reversibility_class IS NULL OR ...`). Add the `SET NOT NULL` + tightened CHECK in a later migration after the new binary is fully rolled out, with its own `migration_00NN_test.go`, and a regression test inserting with the old column list against the `0138` schema.

2. **Watermark is poisoned by an unclamped, agent-controlled future event time.**
   `internal/domain/correlation/incremental.go:237-241` advances `Watermark = MaxObservedAt - AllowedLateness` with no clamp against processing time, and `MaxObservedAt` comes verbatim from `detection.Observed` (validated only for `IsZero`). The watermark is monotonic + persisted, so one detection with a far-future `Observed` (agent clock skew or a malicious enrolled agent) pushes the watermark forward permanently and every later real detection is classified `too_late` into its own one-signal incident. `responseverificationingest/service.go:84-85` already rejects future observations, confirming the missing boundary.
   **Fix:** reject a future `Observed` at ingest (mirror `responseverificationingest`), and clamp in the domain: `candidate := minTime(MaxObservedAt, processedAt).Add(-AllowedLateness)`. Add tests for both.

3. **Correlation-on-ingest advances only one pipeline step per sealed batch (no drain).**
   `correlationuc/service.go` performs one bounded step; reaching a first incident from idle needs 3+ calls. `detectledger` `correlateAfterIngest` runs `correlate` once and ignores `HasMore`, and the run only re-schedules on a coalesced pending batch, never on `HasMore`. So a burst that then goes quiet strands in `PhaseSource` with zero incidents until another batch is sealed or an operator re-POSTs `/correlate` — contradicting `detectledger/service.go:104` ("an incident exists as soon as detections are sealed"). The full-drain contract exists only client-side in the SLO test.
   **Fix:** drain on ingest — loop `CorrelateEngagement` while `res.HasMore || res.Phase != ""` (with an iteration cap, treating `ErrConflict` as retry). Add a `detectledger` test with a correlator that needs 3+ steps.

4. **Active-session saturation wedges a snapshot permanently.**
   `correlationuc/service.go:258` returns `ErrSaturated` before the commit once `len(ActiveSessions) > MaxActiveSessions` (default 500), and `incremental.go:249` prunes nothing on non-final pages, so distinct `(asset,host)` sessions accumulate across the whole snapshot. `ErrSaturated` isn't retried, there's no abort/reset on the store port, and a mid-snapshot policy change hard-errors, so raising the limit doesn't recover. Any engagement whose snapshot references >500 distinct hosts halts incident production permanently.
   **Fix:** evict least-recently-active sessions when the bound would be exceeded (emitting a coverage/suppression event) so the page commits and the checkpoint advances, or add `AbortCorrelationSnapshot` to reset `Phase`. Add a multi-page test that exceeds the limit and asserts forward progress.

### Notable non-blocking
- Attach-event timestamp is non-deterministic across consume page boundaries (`incremental.go:377-380`).
- Fleet host-isolation control is inert in production (`fleet_handler.go:126-139,461-471` gate on setters nothing calls).
- `CHANGELOG.md` Unreleased not updated despite new operator routes; migrations `0139/0140/0141/0143` ship without a dedicated `migration_00NN_test.go` (only `0138`/`0142` have one).
- `SameSignal` uses `==` on `*Timeline` (compares monotonic clock + `*Location`); incident-scoped apply binds a client-supplied engagement/target not the incident's; `approval.Service.Decide` changed 400→403 with no service test.

Blocker 1 is an unambiguous phased-migration rule violation; blockers 2-4 make the headline correlation feature triggerable-broken by untrusted input or unreliable in normal bursts. Recommend addressing 1-4 (or splitting the correlation drain/saturation work into a follow-up with the feature gated) before merge.

pho-veteran:
Addressed all findings from the deep review and rebased the branch onto current upstream `main` (`1c1c36a3`).

Must-fix blockers:

1. **0138 rolling compatibility** — kept `reversibility_class` nullable in the expand phase, backfilled existing rows, and added a compatibility trigger for the exact pre-0138 INSERT shape. A real PostgreSQL regression test writes all supported action kinds through the legacy column list and reads them through the new repository.
2. **Future event-time poisoning** — V1/V2 detection ingest now rejects signed observations after trusted receipt time before evidence/provenance/record effects, audits the rejection, and fails closed if audit persistence fails. Watermark calculation also clamps to trusted processing time while retaining the actual maximum observation for diagnostics.
3. **Correlation-on-ingest draining** — the callback now carries phase/progress and drains bounded source/consume steps to terminal state, retries conflicts within a strict budget, and audits failures/exhaustion while preserving per-engagement coalescing.
4. **Active-session saturation** — replaced the pre-commit wedge with deterministic per-signal top-K retention. Every signal still receives an exact immutable assignment; evictions produce stable visible incident coverage events and audit counts. Multipage/page-size/replay/forward-progress tests cover the new behavior. The state-bound hot path was also optimized so the 20k-signal pipeline SLO passes.

Other findings:

- Attachment events now use stable signal event time; page-order tests compare complete events including timestamps.
- Timeline signal equality is semantic and uses `time.Time.Equal`.
- Production fleet transport now requires and wires distinct client-certificate and enrollment hosts plus the trusted certificate header; config and Helm fail closed, with render/config tests and docs.
- Incident-scoped response no longer accepts client-selected `engagement_id`. The coordinator loads canonical incident provenance and binds process/host targets before preparation; incident-scoped file quarantine fails closed until authoritative asset provenance exists. Mismatch tests prove no action or incident-event side effects.
- Blank/machine approval actors return `ErrForbidden` without persisting a decision or audit event.
- Added dedicated PostgreSQL migration coverage for 0139, 0140, 0141, and 0143, including RLS/policies/constraints/triggers/rollback-reapply behavior. Also fixed the 0139 rollback history guard to inspect rows despite FORCE RLS.
- Added an Unreleased changelog entry and updated OpenAPI, including authoritative incident engagement provenance and correlation eviction counts.

Validation after rebase:

- Focused correlation/detectledger/incident/response/approval/HTTP/config/OpenAPI/PostgreSQL suites: pass.
- Real disposable PostgreSQL tests for migrations 0138/0139/0140/0141/0143: pass.
- `make build` and `make vet`: pass.
- `CGO_ENABLED=0 go build ./cmd/...`: pass.
- Web: 105 files / 618 tests pass; TypeScript typecheck and production build pass.
- Helm 4 lint/render and Go chart wrappers: pass.
- Full `go test ./...`: all packages pass except the five known live bubblewrap integration tests in this restricted host (`bwrap: Failed to make / slave: Permission denied`); fail-closed sandbox behavior was not weakened.
- `git diff --check`: pass; no local env, Terraform state/plan, TLS key, or generated secret artifacts are in the diff.

PR remains open and unmerged for human review.

nghiadaulau (approved):
**Reviewed in depth (CI disabled) — approving.** This is the most carefully-defended path in the platform and it holds up.

**Build/test:** `go build ./...` + `go vet` clean; the changed backend packages (response/responsesaga/correlation/incident/detection/telemetry/fleetagent + their usecases + response infra + httpapi + synapse-agent) — 53 packages — all pass. No web changes.

**Security review — the governed host-execution path is SAFE TO SHIP. No critical/high/medium.** Every golden rule is enforced *structurally*, not by convention:
- **Rule 1 (argv, no shell):** `Action.Argv` is a fixed catalogued template; `Validate` + `containsShell` reject anything else; the actuator never execs the display argv — it maps a validated command to a direct kernel API (`PidfdSendSignal` on a telemetry-pinned pidfd; restart via `exec.Command(exe)` with empty env, O_PATH-pinned fd, ELF/file-capability/owner checks, and refuses any process that had cmdline args). No interpolation anywhere.
- **Rule 2 (scope+authz server-side before dispatch):** every apply routes through the same `safety.Gate` as exploitation, requires a non-machine human approver with SoD, and the command is ed25519-signed by the control plane and re-verified agent-side (signature, key validity, addressing, authorization window, halt fence) before any side effect — a compromised agent can't run an unsigned/unaddressed command. Signed kill-switch (`ResponseHaltCommand`) with a latched, fail-closed fence. RBAC: apply/revert=PermOperate, decide=PermReview, observer-assign=PermAdminister.
- **Rule 3:** observed cmdline+env are deliberately dropped; signing keys `0600`, never logged.
- **Rule 5/6:** command-applied ≠ verified-succeeded is structural (`StateVerifiedSucceeded` reachable only via `StateVerifying`); verification requires an *independent* verifier (executor≠verifier enforced), from a purpose-signed telemetry receipt sealed as hash-chained evidence; HITL gate; append-only audit with actor id.
- **Tenant isolation** on all new correlation/response stores + handlers.
- **Event-time correlation:** a pre-watermark signal is isolated into a deterministic per-signal incident id with a coverage note — it does **not** re-mint/duplicate an existing incident. No reminting hazard.

**Architecture review — PASS.** Dependency rule holds across all 10 domain packages; new usecases import only domain+ports; adapters don't reach infra internals; `cmd/synapse-agent/response_runtime.go` is a composition root only; port assertions present.

**Non-blocking follow-ups** (fine to ship, address later):
1. Optional deps on `correlationuc`/`detectledger` are injected via post-construction setters (`SetNotifier`/`SetCorrelator`/…) read unguarded — harmless as wired at composition time, but fold them into the constructor (or guard) to remove the latent set-after-serving race.
2. `isolate_host`/`quarantine_file` are accepted at the DTO layer but only `stop_process` has a signed target-resolution path; they fail closed today — keep them gated until their fingerprint→target binding gets the same asset-binding review the process path received.

I resolved the merge conflicts with current `main` (CHANGELOG union + the additive router/`main.go` wiring collision with #779's `scanRunHistory`, gofmt'd, compiler-verified) and pushed to the branch. Excellent, rigorous work. LGTM.

PR #907, opened 2026-09-09
feat(sca): preserve dependency scope per edge

VietSory:
## Summary

Preserves dependency scope on each SBOM relationship instead of flattening it onto a component.

* Adds per-edge `Scope` and `Optional` metadata to `sbom.Dependency`; owned lockfile parsers split mixed edge groups deterministically.
* Computes effective scope through the dependency graph: dev/test/background paths stay non-shipping transitively, while any production path wins.
* Uses effective scope in SCA vulnerability classification and the project dependency graph.
* Handles CycloneDX synthetic project roots so an explicit root-edge scope is not overridden by a production-shaped target component.
* Adds regression coverage for dev-only transitive paths, mixed prod/dev diamonds, optional edges, and synthetic-root propagation.

## Validation

Rebased as one commit (`94d88b49`) directly on current `main` (`03c8271d`).

Passed:

* Go lint
* Frontend test, typecheck, and build
* Windows Go build, vet, and full test suite
* Linux D3.3 packages: `internal/domain/sbom` and `internal/usecase/sca`

Linux `go test ./...` remains blocked by the pre-existing, untouched `internal/usecase/response.TestIncidentCoordinatorPreparesPostgresActionBeforeRequestProjection` Postgres conflict. It reproduces consistently and is outside this PR’s diff.

### DIFF
--- a/internal/infrastructure/tools/ownsbom/cargo.go
+++ b/internal/infrastructure/tools/ownsbom/cargo.go
@@ -119,7 +119,9 @@ func (Cargo) Parse(ctx context.Context, in ParseInput) ([]sbom.Component, []sbom
 		return "", false // unresolvable (not in the lock, or ambiguous without a version) -> no edge
 	}
 
-	// Pass 2: emit components + resolve edges.
+	// Pass 2: emit components + resolve edges. Cargo.lock does not encode dependency kinds per edge, so the
+	// relationship inherits the lockfile's path-derived base scope; direct dev roots then carry development
+	// through ReachableScopes to their transitive descendants.
 	set := newComponentSet()
 	var deps []sbom.Dependency
 	for _, p := range pkgs {
@@ -142,7 +144,7 @@ func (Cargo) Parse(ctx context.Context, in ParseInput) ([]sbom.Component, []sbom
 			}
 		}
 		if len(on) > 0 {
-			deps = append(deps, sbom.Dependency{Ref: ref, DependsOn: on})
+			deps = append(deps, sbom.Dependency{Ref: ref, DependsOn: on, Scope: scope})
 		}
 	}
 	return set.components(), deps, nil
@@ -151,8 +153,8 @@ func (Cargo) Parse(ctx context.Context, in ParseInput) ([]sbom.Component, []sbom
 // cargoDevDeps reads the companion Cargo.toml beside Cargo.lock (if present) and returns the set of DIRECT
 // [dev-dependencies] names, so the lock's resolved crates can be scoped dev vs prod. Best-effort – no
 // manifest, or an unreadable one, yields no dev refinement (everything stays prod/path-scoped). Their
-// transitive dev-deps stay prod (Cargo.lock is a flat resolved set; precise dev-transitivity needs the
-// graph) – same limitation as the Syft path, which Cargo.lock's lack of a dev flag forces.
+// transitive dev-deps are later classified by graph scope propagation, which preserves development along
+// every path rooted at a direct dev dependency.
 func cargoDevDeps(dir string) map[string]bool {
 	dev := map[string]bool{}
 	if dir == "" {
--- a/internal/infrastructure/tools/ownsbom/conan.go
+++ b/internal/infrastructure/tools/ownsbom/conan.go
@@ -160,8 +160,10 @@ func (Conan) Parse(ctx context.Context, in ParseInput) ([]sbom.Component, []sbom
 		}
 	}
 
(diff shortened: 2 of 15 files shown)


nghiadaulau:
**Reviewed strictly (CI disabled). Solid, well-built work — go-arch PASS, build + the new tests green, `ReachableScopes` is a clean, cycle-safe, deterministic pure-domain function. But it now overlaps a PR that merged while this was open (#918), and needs a rebase + one reconciliation before it can land.**

### What changed under it
`main` merged **#918 "maven transitive tree with per-edge scope"**, which already added `Dependency.Scope`/`Optional` (hence the `sbom.go` conflict — both PRs add the same fields; main's has the `json:",omitempty"` tags) **and** a scope-restriction model in `sbom.go`: `scopeRank`/`scopeLabel`, which maps **both Maven scopes and the domain `Scope*` labels** (compile/runtime = production, **provided/system = "present at build, not shipped"**, test = never shipped).

### The reconciliation (blocking)
This PR introduces a **second, independent** scope-propagation path — `ReachableScopes` + `normalizedScope`/`isNonShippingScope` — with a **narrower vocabulary**. `isNonShippingScope` is `scope == ScopeDevelopment || IsBackgroundScope(scope)`, and `IsBackgroundScope` does **not** include Maven `provided`/`system`. So `ReachableScopes` classifies a `provided`/`system` edge as **shipping/production**, while #918's `scopeRank` classifies it as **not-shipped**. Because `ReachableScopes` is wired into `sca/service.go`'s gate scope and runs over *every* SBOM — including Maven graphs, whose edges now carry `provided`/`system` from #918 — a **provided-only (never-shipped) Maven dependency's vuln could stay in the actionable gate** that #918's model would deprioritize. Two divergent scope models in one gate will drift.

**Ask:** rebase onto current `main` (take main's `Dependency` struct with the omitempty tags) and have `ReachableScopes` classify scopes through #918's existing `scopeRank`/`scopeLabel` (or at minimum make `isNonShippingScope` agree with it on `provided`/`system`) so the gate and the dependency-evidence path can't disagree. The genuinely-additive value here — extending the 8 non-Maven parsers (npm/pnpm/poetry/yarn/uv/cargo/conan/nuget) to emit per-edge scope, plus graph propagation into the gate — is worth keeping; it just needs to sit on top of #918's model rather than beside it.

### Also fold in (from the arch review)
- **Monotonic clamp:** the graph override is unconditional in the demotion direction, so a shipping component reached only via a background-scoped ancestor could be demoted into a gate-exempt bucket and drop out of the actionable gate. Low likelihood with the current parsers (they emit only production/development), but in a security tool the gate should never *lower* a component out of scope — clamp refinement so it never moves a shipping component into an `IsBackgroundScope` bucket.
- **Tests:** `dependency_scope_test.go` and `npm_edge_scope_test.go` are single-scenario; make them table-driven across the multi-path / demotion / cycle cases like `reachable_scope_test.go` already is.

Happy to re-review promptly once rebased — the core is good, this is about not shipping two scope models.

VietSory:
I ran a control CI on the untouched `af9a0ee7` tree, with none of the changes from #907 included.

Control CI #82 reproduces the same failure classes currently seen on #907:

* **Lint:** typo `clas` in `internal/infrastructure/tools/taintrules/load_test.go`
* **Linux:** `TestIncidentCoordinatorPreparesPostgresActionBeforeRequestProjection` conflict in `internal/usecase/response`
* **Windows:** failures in `blob`, `sourceupload`, and the IaC test under `internal/usecase/sca`
* **Frontend:** passes on the control run as well

These failures therefore reproduce on the baseline `main` tree independently of #907.

#907 is currently rebuilt on `main`, is `mergeable=true`, and its diff is limited to the intended 16 files. The targeted D3.3 tests and `go vet` also pass.

nghiadaulau (approved):
**Approved — the reconciliation I asked for is done. Merging.** You dropped the divergent `ReachableScopes`/`isNonShippingScope` and now emit `sbom.ScopeProduction`/`ScopeDevelopment` on edges, deferring scope propagation to #918's model (its `scopeRank`/`ReachableScopes` in `sbom.go`). That removes the two-scope-model / Maven-provided-treated-as-shipping hazard from my earlier review; the PR is now correctly scoped to the 8 non-Maven parsers' per-edge emission + the projectuc wiring.

Verified on current main: `go build ./...` + `go vet` clean; ownsbom / sca / projectuc / domain-sbom tests green; diff limited to the intended 16 files; rebased and MERGEABLE. (Re: the CI classes you noted — agreed those are pre-existing on the baseline tree, unrelated to this diff.) Nice, clean turnaround. LGTM.

PR #942, opened 2026-09-10
feat(secrets): expand credential detector coverage

VietSory:
## Summary

Expands Synapse's owned secret detector catalog from **78 to 123 detectors** as part of **#860 Dimension 6.2**.

This adds 45 provider/config-specific credential detectors while keeping the existing owned scanning engine and its current detection behavior intact.

### What changed

* Add 45 new secret detector rules covering additional:

  * cloud and identity credentials
  * CI/CD tokens
  * observability/service API keys
  * package registry credentials
  * HTTP authorization credentials
  * Kubernetes/config credentials
  * additional SaaS/API tokens
* Split the existing base secret rules from the parity expansion so the catalog remains maintainable.
* Preserve the existing:

  * keyword prefiltering
  * secret redaction
  * placeholder allow-list behavior
  * offline/read-only scanning posture
* Add positive detection fixtures for the new rules.
* Add placeholder-negative coverage to avoid treating common example credentials as findings.
* Update the rule catalog/golden parity data for the expanded detector set.
* Document the user-visible detector expansion in `CHANGELOG.md`.

This change is intentionally scoped to the D6.2 detector catalog expansion and does not introduce a third-party secret scanning engine.

## Validation

Validated on a clean branch based directly on the current upstream `main`.

* Owned secret detector count: **78 → 123**
* Secret scanner targeted tests: pass
* Rule catalog/parity tests: pass
* Relevant `go vet`: pass
* Linux/Windows build and vet: pass
* Frontend CI: pass

Fork validation also surfaced existing failures outside this PR's 8-file diff, including Windows integration failures and an existing lint typo on `main`; the secret scanner and rule catalog portions pass.

Part of #860


nghiadaulau (approved):
Reviewed locally (CI disabled) — strict pass. Adds ~11 provider secret detectors (GCP OAuth client-secret/refresh-token, kubeconfig bearer, Authorization Bearer/Basic, .npmrc/.pypirc/.netrc/htpasswd, Elastic API key, GitLab deploy token), appended to the mature set so existing rules stay stable.

**FP guards (the key concern for a secret scanner):** every new rule is keyword-prefiltered + length-bounded + entropy-gated (minEnt 3.0) + subject to the framework allow-list + inline `synapse:allow` suppression, and the catalog-parity/count tests keep rules ↔ catalog ↔ rule_keys.txt in lockstep. The broad `authorization-bearer`/`authorization-basic` patterns are the most FP-prone but require a 20+/16+ char, entropy≥3.0 token after the `authorization: bearer/basic` keyword — a low-entropy doc placeholder (`bearer YOUR_TOKEN`) is excluded, a real token flagged; consistent with how gitleaks handles these. `go build` + secretscan tests green.

Minor (non-blocking): if the generic `authorization-*` pair proves noisy on realistic-looking fixtures in docs/tests, consider nudging their entropy floor up a touch — but the keyword+length+entropy+allow-list stack is a sound default. LGTM.

Issue #974, opened 2026-09-11
EPIC VI-09: Monitoring security, tenant isolation, and concurrency correctness


nnatuan03:
Parent epic: #965

## Objective

Validate the security and consistency boundaries of continuous vulnerability monitoring before rollout.

## Product contract

Synapse creates and updates asset inventory on its own scan schedule. Vulnerability Intelligence synchronizes advisory data and correlates it with persisted inventory. It does not rescan assets daily, integrate Trivy, or maintain a duplicate vulnerability database. This clarification takes precedence over scanner-specific requirements in `VULNERABILITY_INTELLIGENCE_MODULE.md`.

## Existing implementation to reuse

Reuse WithTenant and RLS, tenant-aware foreign keys, transactions, authentication, RBAC, CSRF protection, safe HTTP clients, audit logging, and the hostile test harness.

## Scope

- [ ] Review the boundary between global advisory reference data and tenant-owned assets, inventory, evidence, and jobs. Verify that asynchronous execution preserves tenant context.
- [ ] Review provider and credential trust boundaries, including network policy, redirects, DNS and private addresses, body and decompression limits, cancellation, and secret-safe errors and logs.
- [ ] Test transaction and generation fencing across concurrent inventory changes, advisory changes, manual reconciliation, and recovery. Prevent stale evidence retirement.
- [ ] Verify authorization for inventory imports, source management, and manual execution. Missing authentication dependencies must fail closed.
- [ ] Prevent cross-tenant disclosure through impact queries and shared provider-run links. Enforce concurrent workflow deduplication through database constraints.
- [ ] Review production configuration, non-superuser database access, RLS enforcement, and least privilege. Data-only workers must not weaken Synapse scanner sandbox requirements.

## Dependencies

#966, #967, #968, #969, #970, #971.

## Acceptance criteria

- [ ] Hostile multi-tenant, CSRF, RBAC, and authentication failure tests pass.
- [ ] Race, lease-loss, and out-of-order execution tests preserve exposure correctness and human decisions.
- [ ] Hostile provider fixtures demonstrate bounded processing and credential protection, with no unsafe rendering of provider content.
- [ ] Scanner privileges and unrelated modules remain unchanged. Identified defects are resolved or explicitly tracked before rollout.

## Implementation status as of September 11, 2026

Existing safeguards and selected hardening changes are available. A consolidated review of the clarified monitoring workflow is still required.

[...]

Issue #968, opened 2026-09-11
EPIC VI-03: Incremental correlation and durable recovery without asset rescanning


nnatuan03:
Parent epic: #965

## Objective

Identify assets affected by new or changed advisories using their existing inventory, including changes imported through the CLI or committed before a worker failure.

## Product contract

Synapse creates and updates asset inventory on its own scan schedule. Vulnerability Intelligence synchronizes advisory data and correlates it with persisted inventory. It does not rescan assets daily, integrate Trivy, or maintain a duplicate vulnerability database. This clarification takes precedence over scanner-specific requirements in `VULNERABILITY_INTELLIGENCE_MODULE.md`.

## Existing implementation to reuse

Reuse `vulnerabilityruntime`, `vulnerabilitycorrelation`, indexed package and CPE inventory queries, advisory revisions, and evaluation checkpoints. Reference #533 and #535.

## Scope

- [ ] Correlate advisory changes against current persisted inventory for every eligible tenant and asset. Use package and CPE indexes to select candidates.
- [ ] Reuse the revision and checkpoint backlog or existing durable queue and outbox mechanisms to recover changes committed before dispatch, including CLI imports and worker restarts.
- [ ] Paginate advisory, asset scope, and component traversal. Bound each unit of work and ensure that repeatedly failing records do not block unrelated advisories or assets.
- [ ] Bind evaluations to advisory revisions and inventory generations. Enforce consistency and fencing at publication so that stale workers cannot create, retire, or overwrite current evidence.
- [ ] Use shared matching and version semantics for inventory changes and advisory changes. Monitoring must not call a scanner, generate an SBOM, save a new scan snapshot, or duplicate components.

## Dependencies

#966, #967.

## Acceptance criteria

- [ ] A single Synapse scan followed by a new advisory creates a finding for the affected asset without another scanner invocation.
- [ ] Tests cover more than 100 assets and more than 100 advisories, with tenant and inventory scope isolation.
- [ ] Recovery handles CLI imports, retries, and failure after materialization. Evaluation checkpoints advance only after complete processing.
- [ ] Unsupported or indeterminate matching retains prior evidence, and invalid or non-advancing cursors are rejected.
- [ ] Inventory identifiers remain unchanged, and concurrent tests prevent mixed-generation publication and stale writes.

## Implementation status as of September 11, 2026

[...]

Issue #966, opened 2026-09-11
EPIC VI-01: Architecture, inventory identity, and removal of Trivy integration


nnatuan03:
Parent epic: #965

## Objective

Establish the implementation contract for continuous monitoring against Synapse inventory and remove the Trivy integration introduced during earlier work.

## Product contract

Synapse creates and updates asset inventory on its own scan schedule. Vulnerability Intelligence synchronizes advisory data and correlates it with persisted inventory. It does not rescan assets daily, integrate Trivy, or maintain a duplicate vulnerability database. This clarification takes precedence over scanner-specific requirements in `VULNERABILITY_INTELLIGENCE_MODULE.md`.

## Existing implementation to reuse

Reuse `domain/asset`, engagement project and host bindings, persisted SBOMs and components, and native scan provenance. Reference #514, #531, #534, and the scanner boundaries in #860.

## Scope

- [ ] Map each requirement in VULNERABILITY_INTELLIGENCE_MODULE.md to existing implementation, required changes, and acceptance tests. Record the product clarification that supersedes scanner-specific requirements.
- [ ] Review existing work in progress and remove unreleased code, configuration, and migrations introduced solely for Trivy. Preserve unrelated legacy fixtures and integrations. Use forward migrations for any schema changes already deployed.
- [ ] Define identity using tenant, technical asset or target, and current inventory scope. Treat BusinessAsset as a grouping construct, not a substitute for technical target identity.
- [ ] Define inventory snapshot, completeness, and generation contracts. Distinguish missing, empty, partial, and historical inventory, including project, host, branch, image, and module boundaries.
- [ ] Update the specification adaptation and implementation plan so that daily monitoring does not invoke a scanner, generate an SBOM, or perform license scanning.

## Dependencies

None. This epic establishes the architecture and scope required by the implementation epics.

## Acceptance criteria

- [ ] No newly introduced Trivy runtime, configuration, or API dependency remains, and existing Synapse scanning behavior is preserved.
- [ ] Multiple correlation evaluations can use one inventory snapshot without changing SBOM or component identifiers.
- [ ] Different assets and inventory scopes remain isolated. Missing or partial inventory is never treated as authoritative empty inventory.

## Implementation status as of September 11, 2026

The recently added Trivy wiring has been removed from the local working tree. Documentation still requires alignment, and migration compatibility and implementation review are not complete.

[...]

Issue #965, opened 2026-09-11
EPIC: Continuous Vulnerability Intelligence for Synapse Asset Inventory


nnatuan03:
## Objective

Identify which assets are affected by newly published or changed vulnerabilities without waiting for another asset scan.

Synapse is the scanner and SCA engine responsible for creating and updating asset inventory. Vulnerability Intelligence continuously correlates the existing inventory with changes in the shared advisory database and maintains the resulting exposure and remediation workflow.

## Operating model

1. Synapse scans an asset on an independently configured schedule, such as weekly, and persists the asset, component, installed version, and inventory snapshot.
2. Vulnerability feeds synchronize daily or according to their configured source schedules.
3. New or changed canonical advisory revisions are correlated with current persisted inventory.
4. Matching evidence creates or updates occurrences, findings, risk assessments, SLA assessments, and actions.
5. Subsequent advisory changes and asset inventory updates drive the exposure lifecycle.

A feed update must not cause target acquisition, repository cloning, application builds, SBOM generation, or scanner execution.

## Requirements and scope

`VULNERABILITY_INTELLIGENCE_MODULE.md` remains the transfer specification. The product clarification dated September 11, 2026 takes precedence wherever that document describes scanner-specific behavior.

- Do not integrate Trivy binaries, adapters, caches, database download jobs, report importers, or endpoints.
- Reuse the existing Synapse advisory database, observations, aliases, revisions, indexes, and APIs. Do not create a second vulnerability database.
- Keep asset scan schedules independent of feed synchronization schedules. Daily monitoring means synchronization and correlation, not daily asset scanning.
- Treat KEV, EPSS, and public exploit metadata as enrichment, not as independent proof that an asset is affected.
- Treat scan results as evidence for a particular inventory snapshot. Do not convert observed scan detections into unrestricted global affected ranges.
- Preserve the separation between automated exposure state and human remediation decisions.

The transfer specification and the English implementation plan at `docs/guide/vulnerability-intelligence-implementation-plan.md` merged to main through #979 on September 12, 2026. #966 owns requirement mapping and inventory safety; #975 owns final requirement-to-test evidence. The detailed execution readiness handoff is published for review in #1082.

## Existing implementation to reuse

Build on #514 and its completed child epics #515, #516, #517, #518, and #519. Reuse canonical advisory storage, source management, durable jobs, inventory, correlation, occurrences, risk evaluation, findings, SLA services, tenant isolation, and the existing user interface.

[...]

PR #979, opened 2026-09-11
fix(vulnintel): establish review contract and fence advisory revision writes

nnatuan03:
## Purpose

First independently reviewable delivery responding to #965 and https://github.com/KKloudTarus/synapse-ce/issues/965#issuecomment-5629450359.

## Included

* Commit the transfer specification and the English Synapse adaptation plan, with explicit ownership and a requirement traceability matrix.
* Preserve the static inventory and dynamic intelligence product contract. No Trivy integration or additional vulnerability database is introduced.
* Reject older advisory revisions in occurrence Apply and in the PostgreSQL upsert predicate.
* Serialize concurrent first publication by logical occurrence identity so state and event history agree.
* Correct empty reconciliation-run persistence to match the non-null schema contract. The new database test exposed this existing error.
* Add real PostgreSQL tests for 16 concurrent initial writers, monotonic revision events, direct stale SQL rejection, and stale domain replay.

## Verification

* Focused PostgreSQL test passed three times against an isolated PostgreSQL 17 database, including fresh migrations.
* The committed snapshot was tested separately from the existing dirty development workspace.
* Occurrence domain race tests passed. Repository build verification is recorded in the follow-up review response.

## Scope remaining

This PR does not complete #966, #968, #974, or #965. Inventory completeness and generation fencing remain outstanding. Other review findings, including workflow identity, bounded correlation, scheduler ownership, cooldown, impact views and retention, retain their existing child owners.

Additional local work on skipped outcomes, withdrawal semantics, authorization, providers, UI and recovery is deliberately excluded from this initial PR. No issue should be automatically closed by this change.

### DIFF
--- a/VULNERABILITY_INTELLIGENCE_MODULE.md
+++ b/VULNERABILITY_INTELLIGENCE_MODULE.md
@@ -1,0 +1,834 @@
+# Vulnerability Intelligence Module
+
+> Synapse adaptation, September 11, 2026: the source-system specification below is
+> retained for traceability. For this repository, the product contract in
+> [the implementation plan](docs/guide/vulnerability-intelligence-implementation-plan.md)
+> takes precedence. Synapse alone scans assets and persists inventory. Vulnerability
+> Intelligence synchronizes the existing advisory database and correlates changes
+> against that inventory without rescanning. Do not implement Trivy, another scanner,
+> or another vulnerability database. Source-system verification below is not Synapse
+> verification evidence.
+
+> Current implementation snapshot: 2026-09-10  
+> Source system: `vuln-assets-management`  
+> Intended consumer: an AI implementing this module in another codebase with a similar architecture
+
+## 1. How to use this document
+
+This is a self-contained implementation package for the current **Vulnerability Intelligence** module. It describes the architecture, data contracts, provider behavior, persistence rules, API/UI surface, configuration, failure semantics, security requirements, known defects, and a recommended porting sequence.
+
+Normative labels:
+
+- **AS-IS**: behavior implemented in the current repository.
+- **PORTING REQUIREMENT**: behavior that should be preserved.
+- **KNOWN GAP**: current behavior that should not be copied blindly.
+- **RECOMMENDED FIX**: intended improvement for a new implementation.
+
+Relative source paths are included only for traceability. The target codebase does not need the same directory layout or language.
+
+## 2. Objective and boundaries
+
+The module turns external vulnerability intelligence and local software inventory into actionable asset findings. It owns:
+
+1. Ingestion and normalization from NVD, OSV, CISA KEV, VulnCheck KEV, FIRST EPSS, ExploitDB, and Trivy.
+2. Matching vulnerabilities to asset components using CPE and ecosystem-aware version ranges.
+3. Component-level evidence with active, resolved, and reopened lifecycle states.
+4. One remediation workflow finding per `(asset, CVE)`.
+5. CVE browsing, finding impact, ingest health, manual execution, and administration through API and UI.
+
+### 2.1 Keep these domains separate
+
+| Domain | Primary storage | Purpose |
+|---|---|---|
+| Canonical CVE intelligence | `cves` | Normalized advisory records for browsing and matching |
+| Asset remediation workflow | `findings`, `asset_vulnerability_occurrences` | Evidence and workflow state for asset vulnerabilities |
+| Scanner database catalogue | `vulnerability_database_entries` | Separate “Vulnerability Database” snapshot/search feature |
+
+This document covers the first two domains and their Trivy integration. Do not conflate the scanner catalogue with the canonical CVE store.
+
+## 3. Architecture
+
+```mermaid
+flowchart LR
+    A[Asset inventory\nasset_versions] --> D[Daily orchestration]
+    NVD[NVD] --> I[Canonical CVE service]
+    OSV[OSV] --> I
+    KEV[CISA KEV] --> I
+    VCK[VulnCheck KEV] --> I
+    EPSS[FIRST EPSS] --> I
+    EDB[ExploitDB] --> I
+    TDB[Trivy DB] --> TS[Trivy scanner]
+    D --> NVD
+    D --> OSV
+    D --> TS
+    A --> TS
+    TS --> I
+    I --> C[(cves)]
+    TS --> O[(occurrences)]
+    NVD --> O
+    OSV --> O
+    O --> F[(findings)]
+    C --> F
+    F --> SLA[SLA assessments]
+    C --> API[REST API]
+    F --> API
+    API --> UI[React UI]
+    M[Retention maintenance] --> C
+    M --> O
+```
+
+Two matching paths coexist:
+
+- **Occurrence-based:** daily Trivy and targeted NVD/OSV write source-scoped evidence, then reconcile findings.
+- **Legacy correlation:** scans candidate CVEs against asset versions and directly upserts findings. It remains in import and manual-correlation flows.
+
+## 4. Runtime stack and external services
+
+- Go services/repositories.
+- PostgreSQL using JSONB, expression indexes, transactions, and `FOR UPDATE SKIP LOCKED`.
+- Trivy CLI with a shared on-disk database cache.
+- React, TypeScript, React Router, React Query, and Vite.
+- Cookie JWT authentication supplied by the wider application.
+
+| Provider | Endpoint | Authentication |
+|---|---|---|
+| NVD CVE API 2.0 | `https://services.nvd.nist.gov/rest/json/cves/2.0` | Optional `apiKey` header |
+| OSV | `https://api.osv.dev/v1` | None |
+| CISA KEV | CISA JSON, GitHub mirror fallback | None |
+| VulnCheck KEV | `https://api.vulncheck.com/v3/backup/vulncheck-kev` | Bearer token |
+| FIRST EPSS | `https://api.first.org/data/v1/epss` | None |
+| ExploitDB | GitLab raw `files_exploits.csv` | None |
+
+## 5. Current source map
+
+| Area | Current paths |
+|---|---|
+| Canonical model/service/repository | `internal/intel/cve.go`, `service.go`, `postgres.go` |
+| Providers | `internal/intel/nvd.go`, `osv.go`, `kev.go`, `vulncheck_kev.go`, `epss.go`, `exploitdb.go` |
+| Targeted refresh/scheduling/retention | `internal/intel/targeted.go`, `schedule.go`, `maintenance.go` |
+| Trivy | `internal/trivy/cli.go`, `database.go`, `trivy.go`, `postgres.go` |
+| Matching/version semantics | `internal/correlation/service.go`, `matcher.go`, `cpe.go`, `postgres.go`, `version/*` |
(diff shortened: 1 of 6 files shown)


pho-veteran (changes requested):
Reviewed against `main` at `fddad63d`. The direction is right and the diagnosis behind it is correct — my comments are one blocking issue, one sequencing risk, and two scope clarifications.

## What this gets right

The three-layer approach is the correct shape. Domain `Apply` rejecting a lower revision protects every store including the in-memory one; the `WHERE EXCLUDED.advisory_revision >= vulnerability_occurrences.advisory_revision` predicate makes the guarantee hold at the database rather than only in a process that read a fresh predecessor; and `pg_advisory_xact_lock` on the logical identity addresses something `FOR UPDATE` structurally cannot — you cannot row-lock a row that does not exist yet, so concurrent first publication was genuinely unserialized. It also matches the existing convention in `advisory_materializer.go:100` and `vulnerability_reconcile_run_store.go:44`.

The `reconciliation_run_id` correction is right, and per migration `0097` (`TEXT NOT NULL DEFAULT ''`) the empty string is the schema's contract, so `item.ReconciliationRunID.String()` is the correct replacement for `nullableID`. The new fence test writes a base occurrence with no `ReconciliationRunID`, so it covers that path.

## The NOT NULL defect is broader than the PR describes

The body calls this "writes without a reconciliation run". On my reading it is every continuous-correlation write, which makes this fix load-bearing for the whole EPIC rather than incidental.

`vulnerabilitycorrelation` sets `ReconciliationRunID` in exactly one occurrence literal, `service.go:400`, on the advisory path. Two callers reach these paths with no run id at all:

- `SBOMReconcileRequest` (`service.go`) has no `ReconciliationRunID` field. `coordinator.go:50` constructs it with `TenantID`, `EngagementID`, `Keys`, `CPEKeys` only. So every SBOM-path detection, and the retirement sweep at `service.go:268-283`, writes a zero id.
- `coordinator.go:86` builds `ReconcileRequest` with `TenantID`, `EngagementID`, `AdvisoryID`, `Revision`, `Cursor` — no run id — so line 400 also writes a zero id.

Only `vulnerabilityreconciliation/service.go:316` passes a real `run.ID`.

With `SYNAPSE_VULNERABILITY_OCCURRENCE_WRITES_ENABLED=true`, the zero id becomes `nil` via `nullableID` and violates `0097`'s NOT NULL. So on PostgreSQL the continuous correlation path could not persist occurrences at all — only the durable reconciliation service could. It stayed hidden because the flag defaults to `false`, no PostgreSQL test performs an occurrence `Upsert`, and the `vulnerabilitycorrelation` tests use `memory.NewVulnerabilityOccurrenceStore()`. Worth restating in the PR body, because it is the difference between a tidy-up and unblocking the EPIC's primary write path.

## Blocking: the conflict return halts intelligence fleet-wide

```go
if tag.RowsAffected() != 1 {
    return fmt.Errorf("%w: occurrence advisory revision is superseded", shared.ErrConflict)
}
```

A superseded revision is the expected, benign outcome of a race — the fence working. Returning an error makes it fatal well beyond this transaction:

`saveOccurrence` → `Upsert`'s `WithTenant` closure → `upsertAndEvaluate` → `vulnerabilitycorrelation` (`return ReconcileResult{}, fmt.Errorf("upsert SBOM occurrence for %s: %w", ...)`) → `coordinator.go:90-92`, which returns on the first engagement error → no later engagement or tenant is processed, no checkpoint is written for any of them, and the sync run fails. The next retry re-hits the same record.

Nothing handles it: `grep ErrConflict` across `vulnerabilitycorrelation/`, `vulnerabilityruntime/` and `vulnerabilityevaluation/` returns no callers. This is the failure-isolation problem in #968 ("repeatedly failing records do not block unrelated advisories or assets"), reached through a new code path.

Suggested change: treat a superseded revision as a successful no-op — roll back the evaluation, return a result with `Changed: false`, and count it (a `Superseded`/`Skipped` counter alongside `Unmatchable` in `vulnerabilityreconcile.Counts` would fit the existing shape). If it must surface as an error, callers need to classify it with `errors.Is` and continue, and that classification belongs in this PR rather than being left for #968.

Also worth a test: the current fence test asserts the stale write loses, but not that the caller survives it.

Note the interaction with the new lock: because the advisory lock serializes `Upsert` on the identity, the second writer reads the committed predecessor and the domain fence catches it, so `RowsAffected() != 1` should be close to unreachable in practice. That makes it a defensive branch — which is exactly why it should not be the one that stops the fleet.

## Sequencing risk: this activates a write path whose inventory hazards are still unfenced

Fixing the NOT NULL violation makes SBOM-path occurrence writes functional on PostgreSQL for the first time. Two unfenced behaviours sit directly behind it:

- `sca/service.go:2654-2670` degrades to an empty SBOM on generator failure (`sbomGenErr`), that document is persisted as the newest SBOM, and the retirement sweep at `vulnerabilitycorrelation/service.go:268-283` has an empty `seen` set and moves every `detected` occurrence to `no_longer_detected`. Completeness is still not persisted — `scan_repo.go` `SaveScan` writes `tool_versions`, `vuln_db_snapshot`, `grype_database_version` and nothing about completeness.
- `component_inventory_store.go:82-102` still recomputes `latest_sbom` per page while the cursor belongs to the previous SBOM.

Retirement of occurrences created by the durable path already works today (those rows carry a real run id, and the sweep preserves the loaded value), so this PR does not create F1 — but it substantially widens the set of occurrences the sweep can act on. Recommend #966's completeness/generation gate lands with or before this path is enabled anywhere real, and that the PR body notes the dependency rather than leaving the ordering implicit.

## Scope clarification: revision fencing is not generation fencing

The PR body says generation fencing remains outstanding, which is correct — flagging it so #968's acceptance criterion is not read as satisfied by revision fencing alone. The SBOM path reads `s.materializer.CurrentRevision(ctx, value.ID)` at write time (`service.go:178`, `:234`), so a delayed or orphaned SBOM worker always carries the newest advisory revision and passes the new fence cleanly. The stale-snapshot half of "stale workers and stale snapshots must not overwrite newer state" needs the inventory generation, and there is still no generation field on `SBOMReconcileRequest`, `ReconcileRequest`, or `Occurrence` to fence on.

## Worth a test: the `''` sentinel makes continuous-path rows permanent retirement candidates

`ListUnreconciled` (`vulnerability_occurrence_store.go:293-306`) selects `state='detected' AND created_at<=$3 AND reconciliation_run_id<>$4`, and it is the retirement stage of the durable reconciliation service (`vulnerabilityreconciliation/service.go:353`). Switching the stored value from NULL to `''` changes how those rows behave in that predicate: `NULL <> 'run-1'` is NULL, so a NULL row was never a candidate, whereas `'' <> 'run-1'` is true, so every continuous-path row now is one.

In the reachable state today this is a change from "impossible" rather than from "excluded" — the NOT NULL violation meant such rows could not be inserted at all — so it is not a regression. And it looks self-consistent for an advisory-scoped run: the correlation stage rewrites still-detected occurrences with `run.ID` before the retirement stage runs, which excludes them. What I could not convince myself of by reading is the full-scope case (`run.AdvisoryID` empty, so the `($5='' OR advisory_id=$5)` filter is open) combined with the correlation stage stopping early on its cursor or page limit — there the sweep's candidate set is every detected occurrence the run has not yet reached, and the ones written by continuous correlation no longer have NULL protecting them.

I am flagging this as unverified rather than asserting a defect. It is worth a test that runs a full-scope reconciliation over occurrences created by the SBOM path and asserts none are retired while still present in inventory, because if it does bite, the symptom is false resolution — the invariant #965 puts first.

## Minor: lock scope covers evaluation

`upsertAndEvaluate` wraps `Upsert` and `EvaluateOccurrence` in one tenant transaction via `transactions.Run`, and `WithTenant` joins that ambient transaction (`tenantTransactionKey`) rather than opening its own. So the advisory lock is held until the outer transaction commits — across SLA assessment, action building, and the notification outbox, not just the occurrence write. Functionally correct, but it serializes far more than the identity write. Taking the lock in a narrower scope, or documenting the intent, would keep contention predictable once bounded correlation lands.

## Summary

Requesting changes on the conflict-handling path. The fence itself, the lock, and the `reconciliation_run_id` correction are all sound, and the last is more significant than the PR claims. Reviewed at `fddad63d`; build and `go vet` pass on the branch in a clean worktree, and the occurrence domain tests pass. I did not run your PostgreSQL fence test.

pho-veteran:
Accepting both corrections. The `Changed: false` point is right, and it lands on the current diff rather than only on my suggestion.

`upsertAndEvaluate` has no `Changed` guard — it calls `EvaluateOccurrence(workCtx, result)` unconditionally after a successful `Upsert`. So for the domain fence this PR already adds (`occurrence.go`: `return UpsertResult{Occurrence: *previous}, nil`, with `Changed` false and `Event` nil), a stale replay proceeds into evaluation today, carrying the *previous*, newer occurrence: `GetCanonicalAtRevision` at the newer revision, `assessments.Upsert`, then `projection.Project` and the SLA reassessment. Actions and the outbox are skipped, but only incidentally — the guard at the end, `!stored.Created && upserted.Event == nil`, trips because the assessment already exists and there is no event. So the rejected attempt does have downstream effects, reachable from the domain fence alone without the SQL branch. That is a stronger statement of your point than I made, and it means the contract is needed in this PR for code already in it.

It is also why `Changed: false` cannot carry the signal, exactly as you say. `!Changed` is already load-bearing for legitimate re-evaluation: the store returns early with `Changed: false` on a no-op upsert and evaluation still runs deliberately, since advisory metadata can move under a stable occurrence. `UpsertResult` currently carries `Occurrence`, `Event`, `Created`, `Changed`, so separating the two cases needs its own signal on the result — a `Superseded` field or equivalent — with `upsertAndEvaluate` bypassing evaluation only on that. One value then covers both entry points, the domain fence and the SQL rejection, and gives the counter something to count.

Agreed the tests belong here, covering both paths. Worth noting the second needs a deliberate seam: with the advisory lock in place the natural path reaches the domain fence first, so the SQL rejection is not reachable by ordinary concurrency and has to be injected.

On lock scope — your position is better than my suggestion. Holding the identity lock across evaluation is what makes evidence and workflow publication atomic, and releasing it early would trade a real guarantee for contention nobody has measured yet. Documenting the intent at the lock site and putting lock-wait visibility in #968/#971 is the right call; I withdraw the narrower-scope suggestion.

The empty-string retirement scenario staying an unverified risk, folded into #975's full-scope multi-page case, is the resolution I was after.

nghiadaulau (approved):
Reviewed locally (CI disabled) — strict pass. A monotonic advisory-revision fence so a delayed provider/recovery job can't overwrite newer vuln intelligence, enforced at **both layers** (defense in depth):
- **Domain** (`vulnerabilityoccurrence.Apply`): returns the previous occurrence unchanged when `next.AdvisoryRevision < previous.AdvisoryRevision` — pure, no infra imports.
- **DB** (`vulnerability_occurrence_store` UPSERT): `… WHERE EXCLUDED.advisory_revision >= vulnerability_occurrences.advisory_revision`, surfacing `ErrConflict: advisory revision is superseded` on a fenced stale write.

`go build` + domain tests + the postgres occurrence/fence tests pass. No migration. Well-scoped integrity fix.

One thing to confirm at your leisure: the switch from `nullableID(item.ReconciliationRunID)` to `item.ReconciliationRunID.String()` writes `""` rather than NULL when the run id is empty — fine if the column is NOT NULL / not an FK, worth a glance if it is. LGTM.

PR #1001, opened 2026-09-11
feat(secrets): add active secret verification

VietSory:
## Checklist

- [ ] `make build vet test typecheck` passes
- [ ] `cd web && pnpm build` passes (if the dashboard changed)
- [ ] Tests added/updated for new behavior
- [ ] Preserves the safety invariants (argv exec, server-side scope enforcement, secrets

### Documentation

Drift guards run in `make test`, so these usually fail loudly rather than silently. Confirm anyway:

- [ ] New or changed `SYNAPSE_*` variables are in `docs/guide/configuration.md`, and
      security-relevant ones state the risk of a careless value
- [ ] New or changed CLI flags, subcommands, and exit codes are in `docs/guide/cli.md`
- [ ] New HTTP routes are described in `api/openapi.yaml` (do not grow
      `api/testdata/openapi-coverage-debt.txt`)
- [ ] New guides are in `mkdocs.yml` nav, so the site actually publishes them
- [ ] Claims match shipped behavior: no unbuilt artifact, unimplemented setting, or
      ingest-only format described as an export
- [ ] Screenshots retaken if the documented UI changed, with no tokens, keys, or real
      customer data captured


nghiadaulau (approved):
**Approved — merging.** This is the highest-risk kind of feature (taking detected secrets and calling providers to check they're live), and it's built correctly. Reviewed locally (CI disabled); `go build` + secretverify/secretscan/sca/config tests pass.

**Security review — CLEAN, no blocking findings.** Traced every concern:
- **Rule 3 (secret never leaks):** the raw secret is a transient callback value (`VisitRawSecret`→`materialize`), never stored on a struct, in an error, or a DTO. Its *only* egress is the legitimate provider request (GitHub `Authorization: Bearer`, Vault `X-Vault-Token`, AWS `X-Amz-Security-Token`). The AWS secret access key is used **only** to derive the SigV4 HMAC signing key — never transmitted or logged (the header carries the public access-key id + signature). No `slog`/`log`/`fmt` touches token material; the `ports.SecretVerification` DTO has no credential field and constrains provider/reason to `[A-Za-z0-9._-]`; response bodies are discarded on 200 / fixed-substring-matched on error (account/ARN never captured).
- **SSRF:** endpoints are fixed constants (`api.github.com`, `sts.amazonaws.com`); finding data only selects *which* check runs, never a URL. Everything goes through `safehttp` (blocks loopback/link-local incl. 169.254 metadata/private/CGNAT, dials the resolved IP → no DNS-rebinding, proxy disabled, **redirects not followed** so headers can't replay to another host). Vault addr is operator-config, validated absolute-HTTPS-no-userinfo with the path forced to `/v1/auth/token/lookup-self`.
- **Opt-in / scope:** default **OFF** (`SYNAPSE_SECRET_VERIFY_ENABLED=false`); wired in all three roots only when enabled. Scoped to the already-authorized workspace worktree — git history and image-rootfs use detection-only and `VisitRawSecret` hard-refuses `FromHistory`; traversal-safe via `os.OpenRoot`. The gate errors out (not degrades) if enabled without the scanner extension.
- **Bounds:** ≤32 provider calls, 8s timeout, 64 KiB body cap, dedup maps prevent fan-out.

**Minor (non-blocking):** `sca/secret_verification.go` loops the verifications twice over slightly different slices (re-checks `Valid()`, so safe) — a maintainer could unify on the filtered `valid` slice. LGTM — excellent, careful work.

nghiadaulau:
**Follow-up on my approval — please read before merging: this now overlaps #1003, which merged while it was open, so it needs a rebase-to-extend rather than a straight merge.**

My security review above stands — the AWS-SigV4 + Vault verification code here is genuinely well-built (no secret leak, SSRF-safe via safehttp, opt-in, workspace-scoped, bounded). The problem is packaging: **#1003 "add opt-in active secret verification" (merged 2026-09-11) already shipped** the GitHub/GitLab half and established the provider framework in `internal/infrastructure/secretverify/` — a `provider{name, baseURL, rejectOn401, check}` registry wired by `register(...)`. Its `verifier.go` even documents the gap this PR fills:

> AWS needs the access-key-id + secret-access-key PAIR (SigV4) … and HashiCorp Vault needs an operator-supplied address. Adding pair-aware AWS and configured-address Vault verification is a documented follow-up; returning SecretUnknown …

So this PR's contribution (pair-aware AWS + configured-address Vault) is real and wanted — but it currently reimplements the whole package as a parallel `verifier.go`, which conflicts with #1003 on 5 files (`secretverify/verifier.go`, `verifier_test.go`, `ports/secretverify.go`, `usecase/sca/service.go`, `cmd/synapse-cli/main.go`) and would ship a second, divergent verifier.

**Ask:** rebase onto current `main` (which has #1003) and re-express this as **two new providers in #1003's registry** — an AWS provider that does the SigV4 GetCallerIdentity check (with the access-key-id/secret pairing) and a Vault provider keyed on `SYNAPSE_SECRET_VERIFY_VAULT_ADDR` → `/v1/auth/token/lookup-self` — plus the `maxChecks`/pairing bounds, rather than a replacement package. Keep the SigV4 signing + Vault-addr validation exactly as reviewed; they're solid. Then the diff shrinks to the delta and merges cleanly.

Holding merge on this reconciliation only. Nice work on the crypto/SSRF details — I want to land it, just on top of #1003.

VietSory:
Reconciled this PR with the active secret-verification framework merged in #1003 and force-updated the branch.

* Rebased onto current `main` (`1eacf614`).
* Kept #1003’s `SecretVerifier`, `VerifyingSecretScanner`, SCA, config, and CLI architecture as the source of truth.
* Removed the duplicate verifier/scanner/DTO/service implementation from this PR.
* Ported AWS as a grouped-verifier extension. The scanner only pairs mutually unambiguous nearby access key and secret key findings, plus a session token for `ASIA` credentials, before making one SigV4 `STS GetCallerIdentity` request. Ambiguous or incomplete pairs remain `unknown` without network egress.
* Ported Vault into the existing provider registry using `SYNAPSE_SECRET_VERIFY_VAULT_ADDR`, absolute HTTPS/no-userinfo validation, and the existing SSRF-hardened HTTP client.
* Preserved the existing per-scan deduplication, 100-check cap, rate limiter, default-off/`--offline` behavior, redaction, and retain-the-finding semantics.

The PR now contains exactly one commit, changes 16 files (`+634/-29`), and GitHub reports `mergeable: true`.

Verification completed:

* Targeted tests for `secretverify`, `secretscan`, config, composition, CLI, API, worker, and configuration-documentation drift
* `go test -race` for `secretverify`, `secretscan`, config, and composition
* `go vet` for the affected dependency cone
* `git diff --check` and conflict-marker scan

The repository-wide test run only encountered unrelated existing/environment failures: the kernel-event timing integration test, Helm tests because `helm` is unavailable, and an existing guide on `main` missing from the MkDocs navigation. None of those paths are changed by this PR.

The reconciliation requested in the previous review is now complete.

nghiadaulau (approved):
**Re-reviewed the rebased version — the reconciliation I asked for is done. Approving + merging.**

You dropped the parallel package and now **extend #1003's provider registry**: Vault is registered as `v.byRule["vault-token"] = &provider{name:"vault", check: checkVault}` and AWS via `aws.go`, rebased onto main-with-#1003 (hence mergeable, and the diff is now the AWS+Vault delta rather than a second verifier).

Verified the reviewed security properties survived the rebase:
- **Secret non-leak (rule 3):** `ports.SecretVerdict` is a string enum and the DTO carries no credential field; `checkVault` puts the secret only in the `X-Vault-Token` header of the legitimate request and returns a verdict enum; errors are static (`build request`/`request`) with no secret; the code documents it never logs the request or headers. AWS still uses the secret access key only to derive the SigV4 signature (never transmitted/logged).
- **SSRF:** delivery still routes through `safehttp` (5 refs in verifier.go); Vault addr is operator-config, HTTPS-validated.
- **Opt-in:** default-off preserved (`!c.SecretVerifyEnabled`), Vault registered only when a validated address is supplied.

`go build` + secretverify/secretscan/sca/config tests pass. Clean turnaround — LGTM.

PR #1023, opened 2026-09-12
fix(ospkg): match Debian source advisories and resolve Fedora/SLES

nghiadaulau:
## What

Two owned OS-package cataloger (`ospkg`) correctness fixes for image scans under `SYNAPSE_SBOM_PRODUCER=ownsbom`, found while benchmarking the owned engine against Trivy.

### 1. Debian/Ubuntu source-keyed advisory matching

A Debian/Ubuntu security advisory is keyed by the **source** package (one `openssl` advisory covers the `libssl3`, `libcrypto3`, … binaries built from it). The advisory matcher already matched a binary against its source via the deb PURL `upstream=` qualifier that Syft records, but `ospkg` never read the dpkg `Source:` field, so under the owned SBOM producer its components carried no `upstream=` and a binary whose source name differs (`gpgv` from `gnupg2`, `libc6` from `glibc`) never matched its source-keyed advisory.

`ospkg` now parses `Source:` (`<name>` or `<name> (<source-version>)`) and emits `upstream=<source>[@<source-version>]`, with the whole value percent-encoded so a hostile `Source` value cannot inject a qualifier, and only when the source name differs from the binary (dpkg omits the field when they are equal).

**Measured:** on `python:3.9.18-slim` (Debian 12), the owned engine's Debian CVE recall against Trivy rose from **29% (78/265) to 98% (259/265)** (the remaining 6 are Debian-tracker `DLA`/`TEMP` identifiers, not CVEs). PyPI stayed 12/12.

### 2. Fedora / SLES resolved-flag drift

The resolved-flag gate (`rpmMatchableIDs`) had drifted behind the Fedora updateinfo feed (`Fedora:<major>`) and the SLE OVAL feed (`SUSE:<major.minor>` per service pack) that already landed, so a Fedora or SLES image warned "distro unresolved" while it in fact matched. `fedora` and `sles` are now in the matchable set; SLES resolves only with a `major.minor` VERSION_ID (bare `SUSE:15` would conflate service packs). CentOS stays deliberately unmapped.

## Soundness

The `upstream=` value is percent-encoded through an allowlist encoder (`@`→`%40`, `&`→`%26`, `=`→`%3D`, `?`→`%3F`), so an untrusted image's `Source` field cannot inject a second qualifier or cross-key to another release. Emitting the source package is exactly what Debian security tracking intends (advisories are source-keyed), so this raises recall without introducing false positives. rpm and apk paths are unchanged.

## Tests

- `TestDpkgSourceQualifier` (table): Source parsing edge cases.
- `TestCatalogDebianSourceUpstream`: binary→source `upstream=` emission and round-trip.
- `TestCatalogDebianSourceQualifierInjectionSafe`: a hostile `Source: evil&distro=lies` cannot inject a `distro=` qualifier.
- `TestCatalogRPMFedoraResolved`, `TestCatalogRPMSLESResolution`: Fedora/SLES resolved-flag.

An internal multi-agent review (correctness / security / soundness lenses, adversarially verified) surfaced one low-severity gap: no regression test pinned the qualifier-injection safety of the `Source` field. That test was added (`TestCatalogDebianSourceQualifierInjectionSafe`) before opening this PR.

`go build ./cmd/... ./internal/...` (CGO off), `go vet`, and the `ospkg` + `ownadvisory` suites pass.

Refs: #860


nghiadaulau:
Addressed the Codex review: the **High** finding (a crafted `Source: openssl@1.0` could smuggle the matcher's `@` delimiter past percent-encoding and attribute openssl CVEs to a benign binary) is fixed by validating the parsed name/version against the Debian `Source` grammar and rejecting malformed fields — an invalid Source now emits no `upstream=` qualifier. Added table cases for `@`, qualifier separators, unclosed/trailing-junk parens, and uppercase. The **Medium** (same-name binNMU source version dropped) is a false-negative-only edge, not a false positive: dpkg orders a `+bN` binNMU between the source revision and the next source version, so matching at the binary version is affected-status-equivalent; addressing it fully needs matcher support for a same-name source version and is left as a follow-up.

PR #1024, opened 2026-09-12
fix(advisory): scope affected symbols to version-matching block

nghiadaulau:
## What

A soundness fix for advisory-provided affected symbols (the input to symbol-level reachability), on both the owned offline matcher and the live OSV scanner.

### The defect

OSV permits the same package in several `affected[]` blocks with different version ranges, each carrying its own `ecosystem_specific` symbols. The owned matcher determined a version match against one block but then collected symbols across **all** name-matching blocks (`Match` + `AffectedSymbolsFor`), so a finding for version A could carry a symbol that only applies to version B. That would seed a false "reachable vulnerable symbol", which the raise-only reachability path uses to prioritize a finding, and which feeds evidence.

### The fix

- `advisory.MatchDetails(ecosystem, name, version)` returns the matched fixed version together with symbols drawn **only** from the blocks whose range actually includes the component version. `Match` now delegates to it; `AffectedSymbolsFor` is kept but documented as version-agnostic and not for the finding path.
- The owned detection source (`ownadvisory/source.go`) uses `MatchDetails`.
- The live OSV scanner (`osv/scanner.go`) had the same defect (it unioned symbols across name-matched blocks) and additionally read only the Go vuln DB's `imports[].symbols`, not RustSec's `ecosystem_specific.affects.functions`. It now version-scopes its symbols (`versionScopedAffected`) and reads both symbol forms, matching the owned offline ingester so the live and offline paths no longer drift.

## Soundness

Strict version-scoping is the safe direction: a symbol dropped by it only under-drives raise-only reachability (a missed prioritization), never a false suppression. A block that declares no ranges and no explicit versions is kept (OSV: no constraint means all versions).

## Tests

- `advisory.TestMatchDetailsSymbolsAreVersionScoped`: a 1.x component gets only the 1.x block's symbol, a 2.x only the 2.x, a gap version none; `AffectedSymbolsFor` stays version-agnostic.
- `osv.TestOsvToRawSymbolsAreVersionScoped`: the live scanner attaches only the version-matching block's symbol.
- `osv.TestOsvToRawReadsRustSecFunctions`: the live scanner reads `affects.functions`.

`go build ./cmd/... ./internal/...` (CGO off), `go vet`, and the `advisory` + `osv` + `ownadvisory` suites pass.

Refs: #860


nghiadaulau:
Addressed the Codex review: (1) **High** — the OSV `limit` event was ignored, so a limit-capped range matched every later version and could keep its symbols out of range; `limit` is now modeled as an exclusive upper bound in the owned matcher (offline + live). (2) **Medium** — dropped the "no ranges and no versions implies all versions" branch (OSV uses an introduced-0 range for that), so an unconstrained block no longer contributes symbols. (3) **Medium-low** — dedup now unions affected symbols so a RustSec entry's functions survive when a symbol-less GHSA alias wins. The RustSec-OSV-export version-req lossiness (a function listed in a block matching broader versions than its native `^0.1` requirement) is an inherent limitation of the OSV export; it is bounded by raise-only reachability (over-prioritize, never suppress) and applies equally to the already-merged D4.6. Added tests for limit, dedup-union, and the no-constraint drop.

PR #1026, opened 2026-09-12
feat(sca): fail-closed detection readiness guard for owned-only scans

nghiadaulau:
## What

A fail-closed **detection readiness guard**: an empty vulnerability database no longer reads as a clean scan. When every configured detection source reports an empty DB after a scan (an unsynced owned advisory corpus, a missing Grype binary/DB, or no source configured) over an SBOM with components, no detection actually ran, so a zero-vulnerability result is a false clean.

The guard, captured once from the just-finished scan's source provenance and carried to result build (so a concurrent advisory sync cannot flip the verdict), then:
- marks the scan completeness **not-confident** and surfaces a `SourceWarning` (sync advisories / provision the DB), and
- under `SYNAPSE_STRICT_SOURCES=true`, **fails closed**.

A source that does not report provenance is assumed to carry its own coverage, so the guarantee is scoped: an owned-only / all-provenance deployment fails closed on an empty corpus; a mixed deployment trusts an opaque source.

This is the safety prerequisite for running with no Syft/Grype installed (`SYNAPSE_SBOM_PRODUCER=ownsbom`, `SYNAPSE_DETECTION_SOURCES=advisory-store`): an unpopulated store fails or warns rather than shipping a misleading clean result. `SYNAPSE_STRICT_SOURCES` is already wired end to end (config → scacompose/CLI → `SetStrictSources`).

## Review

A Codex review of the first draft flagged four issues, all addressed here: (1) an empty source list is now treated as unready (not a false pass); (2) the verdict is snapshotted once and carried to result build rather than recomputed, so a concurrent sync cannot flip it; (3) the opaque-source branch is kept but its scoped guarantee is documented; (4) the strict path returns the warning alongside the error.

## Tests

`TestDetectionReadiness` (all-empty non-strict/strict, one-populated, opaque, no-components, no-sources, strict-returns-warning) and `TestApplyDetectionReadiness` (snapshot mutation). Full `sca` suite, `go vet`, and `CGO_ENABLED=0 go build ./cmd/... ./internal/...` pass.

Refs: #860


nghiadaulau:
Addressed the Codex confirm-review finding: detection coverage now requires a non-empty **DB** marker, not merely a tool version. A source that reports its binary/tool version while its vulnerability DB or advisory corpus is empty (a Grype binary present with an unbuilt DB) is no longer counted as coverage, so it cannot mask an empty corpus, and strict mode fails closed as intended. Added a version-only-source test case. The other three earlier findings (empty source list, single-read snapshot, strict returns warning) were already resolved in the revision.

PR #1027, opened 2026-09-12
fix(pythonprogram): resolve input/eval/exec/compile builtins for taint

nghiadaulau:
## What

The owned Python taint catalog models `input()` as a taint source and `eval`/`exec`/`compile` as CWE-94 code-execution sinks, but a bare call resolves to `builtins.<name>` only when the name is in the resolver's `builtinCallables` allowlist. Those four names were missing, so the source and sink models were **dead** (unreachable via a bare call): `x = input(); os.system(x)` and `eval(tainted)` produced no finding.

Add `input`, `eval`, `exec`, `compile` to `builtinCallables`.

## Soundness

A local or imported definition of the same spelling is resolved **before** the builtin fallback (`lookupLexical` / `lookupImports` run first in `resolveReference`), so a user-defined `eval()` is never mistaken for the builtin sink. No false positive is introduced. Verified by a shadow test case.

## Verified separately (no code change)

While scoping this, I verified end to end through the tree-sitter fact extractor that taint **already** propagates through f-strings, `%`-formatting, `str.format`, `+` concatenation, `str.join`, and comprehensions (via the generic expression value-flow in `valueFor`), so a source carried through any of those into a sink is detected. The remaining gap was purely the bare-builtin source/sink resolution fixed here.

## Tests

`TestPythonBuiltinSourceSinkResolution` (cgo) drives the extractor end to end: `input()`→`os.system` is flagged CWE-78; `eval`/`exec` on tainted input flagged CWE-94; a locally-shadowed `eval()` is NOT flagged. `pythonprogram`, `taint`, `taintscan`, and `astwalk` (cgo) suites and `CGO_ENABLED=0 go build ./cmd/... ./internal/...` pass.

Refs: #860


nghiadaulau:
Addressed the Codex review Medium: `compile` is removed from the CWE-94 sink (it was a dead sink before this change anyway). Compiling untrusted text does not by itself execute it, so flagging `compile(tainted)` would be a benign-syntax-check false positive; the execution risk surfaces at the `eval`/`exec` the code object is passed into, which the sink covers. Added a negative test asserting `compile` is not a code-execution sink. Codex confirmed `input` (source) and `eval`/`exec` (sinks) sound with no resolver false positive for shadowed/attribute cases and no resolution regression.

PR #1028, opened 2026-09-12
feat(compliance): per-framework rollup with failed/not-assessed status

nghiadaulau:
## What

Turns the per-framework compliance rollup (EPIC #860 D6.6) into a **real status model** wired into the scan result. Previously `compliance.Rollup` only counted the controls findings touched and was called only from unit tests (the review's "chỉ đếm, chưa pass/fail, chưa wired").

For each framework a finding maps to (CIS AWS 3.0, CIS Kubernetes 1.10, OWASP 2021, PCI DSS 4.0, ISO 27001), `FrameworkCoverage` now lists **every control Synapse can assess** in that framework with a status:
- `failed` — a finding mapped to it (a deterministic curated CWE+rule lookup, never an inference), or
- `not_assessed` — mapped but no finding.

It **never asserts `pass`**: a clean finding set is not proof a control's requirement is met, only that Synapse's checks for it did not fire, so a `not_assessed` control is never misread as a clean pass (the review's "không có finding không được suy ra PASS"). Each framework carries an `assessable_controls` **denominator** (the controls Synapse maps, not the framework's full catalogue), so `failed of assessable` is never mistaken for full-framework compliance.

## Reachability

The rollup is attached to the scan result's compliance report (`compliance.frameworks`), so it reaches the **HTTP API and the CLI `--json`** output. `NOT_APPLICABLE` (e.g. a Kubernetes framework against a scan with no K8s manifests) is reserved for a future scan-scope signal, and a dashboard panel is a follow-up (the scan-result compliance report is not yet surfaced on a UI-consumed route).

## Tests

`TestComplianceRollup` verifies per-control `failed`/`not_assessed` status, the assessable denominator, and that a mapped-but-untouched control is `not_assessed` (never a pass). `compliance` and `sca` suites and `go build ./internal/...` pass.

Refs: #860


nghiadaulau:
Addressed the Codex review: (1) `controlsByFramework` dedup is now deterministic (lexicographically-smaller title wins) so emitted control metadata never depends on map iteration order; (2) a per-finding `(framework,control)` guard ensures one finding increments a control's failed-count at most once even if it maps through multiple paths. On finding #3 (zero-finding frameworks omitted): this is the intended contract (only frameworks a finding maps to are emitted; the never-assert-PASS property means absence is 'not emitted', never 'clean pass') and is documented on `FrameworkCoverage`/`Report.Frameworks`; a dashboard consuming it must treat an absent framework as not-assessed, not clean. The 'no PASS' guarantee is scoped to `Report.Frameworks` (the existing `Evaluate` Passed/Results model is unchanged).

PR #1029, opened 2026-09-12
ci(owasp): gate the SAST scorecard with a content-pinned corpus

nghiadaulau:
## What

Turns the OWASP BenchmarkJava SAST recall ratchet (EPIC #860 D5.8) into a **real CI gate with a content-pinned corpus**. The gated test existed, but no workflow provisioned the corpus or `synapse-ast`, so the earlier "CI provisions both" claim was aspirational and the ratchet never ran in CI.

- **New workflow** `.github/workflows/owasp-benchmark.yml`: checks out `OWASP-Benchmark/BenchmarkJava` at a pinned commit (`20cbf3d1…`), verifies the answer key's sha256, builds a java-facts-capable `synapse-ast` (CGO), runs `TestOWASPBenchmarkScorecard`, and uploads the scorecard artifact.
- **Content pin, not filename**: the test verifies `expectedresults-1.2.csv` sha256 against a pinned value (`1809f6a6…`, overridable via `SYNAPSE_OWASP_BENCHMARK_SHA256` for a deliberately re-pinned corpus with re-calibrated floors), so a modified or wrong-version answer key can no longer pass the ratchet falsely.
- **Explicit not-covered reporting**: the scorecard now logs the categories the engine does NOT model (crypto, hash, ldapi, securecookie, trustbound, weakrand, xpathi, xss), so the three scored classes are never mistaken for full OWASP coverage.

## Validation

Ran end to end locally against the pinned corpus (checksum verified, 2740 test files): recall **0.437 (cmdi) / 0.301 (sqli) / 0.436 (pathtraver)**, above the committed floors; ratchet green. So CI will be green on merge.

Broader corpora (Juliet, securibench-micro) and cross-category false-positive accounting remain future work (standard OWASP scoring is per-category, which this preserves).

Refs: #860


nghiadaulau:
Addressed the Codex review: (1) removed the `SYNAPSE_OWASP_BENCHMARK_SHA256` override from the workflow so the test verifies the answer key against its checked-in compiled constant, independent of the workflow's own checksum value — true belt-and-suspenders, and re-pinning now requires editing both the workflow and the source constant (reviewed with recalibrated floors); (2) the answer-key loader now tolerates exactly one header row and then fails loud on any further non-true/false row or an empty name/category, so a corrupt or re-pinned corpus cannot silently move the ratchet denominator. Re-validated end to end against the pinned v1.2 corpus (recall 0.437/0.301/0.436, ratchet green).

PR #1030, opened 2026-09-12
feat(ospkg): report CentOS as a structured unsupported-coverage gap

nghiadaulau:
## What

CentOS is deliberately unsupported (CentOS Stream runs ahead of RHEL, so applying a RHEL fixed version would be a false match). Its packages were already cataloged for inventory, flagged unresolved, and never aliased to RHEL, but the pipeline warning conflated this **by-design coverage gap** with a **parse failure** ("release could not be resolved").

The owned OS-package cataloger now recognizes CentOS as a known-unsupported rpm distro (`OSPackageResult.UnsupportedDistro`), and the scan surfaces a distinct `coverage=unsupported` warning naming it as deliberately unmatched. So a CentOS image reads as an explicit unsupported-coverage gap: never a clean posture, never aliased to another distro's advisories.

## Soundness

Both hard requirements were already met (no false clean, no RHEL alias) and remain; this only distinguishes the honest reason. The cataloger sets the flag only for the recognized `centos` id; every other unresolved/garbled release keeps the existing generic warning.

## Tests

`TestCatalogRPMCentOSUnsupported`: CentOS packages cataloged, `DistroResolved=false`, `UnsupportedDistro="centos"`, and the PURL keeps a `centos` namespace (never aliased to rhel). `ospkg` and `sca` suites and `go build ./internal/...` pass.

Refs: #860


nghiadaulau:
Codex review: no findings, sound. Added the suggested guard test (`TestUnsupportedAndMatchableAreDisjoint`): `knownUnsupportedRPMIDs` must stay disjoint from `rpmMatchableIDs`, so a future edit can never flag a supported distro as coverage=unsupported and relabel its real OS findings as a gap.

PR #1031, opened 2026-09-12
feat(reachability): tier-2 raise-only rust affected-symbol reachability

nghiadaulau:
## What

Tier-2 **raise-only** Rust affected-symbol reachability (EPIC #860 D4.6). Rust findings had Tier-1 import reachability, and the RustSec affected functions now flow through the advisory pipeline; this asks the sharper question: does first-party Rust source reference the specific vulnerable function (`crate::path::func`), not merely import the crate?

## Soundness (the crux)

**Raise-only by construction.** It mints a *reachable* judgment (raising a finding's urgency) only for a **proven** reference:
- a fully-qualified path call `crate::…::func(`, or
- a `use crate::…::func` (incl. aliased/grouped) followed by a call `func(`.

It **never** mints not-reachable, because a source scan cannot prove absence without type resolution (a method call `x.foo()` or a macro cannot be tied to a crate). So its worst case is over-prioritizing a finding, never hiding one. The new proof actors are deliberately **excluded from `IsDeterministicReachabilityProof`**, so a Rust symbol verdict can never become an OpenVEX `not_affected`.

**Precision.** Matching requires the last **two** path segments to agree (function + its module/type qualifier), so a same-named local `func()` with no qualifier/use is not matched. Hyphenated advisory crate names are normalized to the underscore source form.

## Architecture

Scanner in `srcimports` (infrastructure, source-only, case-preserving); analyzer in a new `rustsymreach` usecase package behind a `ports.RustSymbolIndex`/`RustSymbolScanner` boundary (usecase imports no infrastructure — architecture test passes); coordinator runs it raise-only over each finding's affected symbols. Enabled with the existing `SYNAPSE_REACH_RUST` flag.

## Tests

`rustsymreach` unit tests: qualified reference, type-qualified method path, use+call, aliased use+call, unreferenced (not reachable), same-named local (NOT matched), hyphenated-crate normalization. `reachproof`, `srcimports`, `sca`, `judgment`, and the clean-architecture test pass; `CGO_ENABLED=0 go build ./cmd/... ./internal/...` passes.

Refs: #860


nghiadaulau:
Addressed the Codex review. (1) The analyzer is now physically incapable of a negative: it appends only reachable (proven) results, so it cannot suppress even without the raise-only wrapper. (2) A qualified reference now requires CALL position (`crate::…::func(`), so an imported-but-never-called function no longer raises on the qualified pass. (3) Block comments are stripped. (4) `rustCallNames` excludes method calls (`x.foo()`) and `fn foo(` declarations, fixing the `use …::insert_many` + `x.insert_many()` false match. Added regression tests for all four (import-only, method-call, block-comment, declaration → not reachable). On the accepted precision bounds (#4): last-two-segment matching can over-raise a same-tail path from another crate/module and miss a renamed root crate or a re-exported shorter path — tolerable for a raise-only signal (over-raise only over-prioritizes; a miss is unknown, never a false negative). Subject builder (#5) keys cargo components and only carries RustSec symbols to cargo findings.

PR #1032, opened 2026-09-12
test(sca): owned-only scan gate proving no Syft or Grype is needed

nghiadaulau:
## What

The end-to-end proof for EPIC #860's "run without Syft or Grype" goal, plus a readiness-guard fix it surfaced.

- **`TestOwnedOnlyScanNeedsNoSyftOrGrype`**: wires the owned SBOM producer (`ownsbom`) and owned detection source (`ownadvisory` over a populated store) as the ONLY sources — no Syft generator, no Grype scanner — and asserts a full scan of an npm lockfile catalogs the component and flags its advisory (lodash CVE-2019-10744). If either owned half regresses, the finding disappears here. This complements the manual `python:3.9.18-slim` benchmark (owned engine measured at Trivy parity) with a committed regression gate.
- **Readiness-guard fix**: building the gate surfaced a false positive in the in-memory deployment. The detection-readiness guard keys on the owned source's provenance marker, which comes from `AdvisoryCorpusFreshness`; only the Postgres store implemented it, so a **populated in-memory** owned store reported an empty marker and the guard wrongly flagged the scan incomplete. `memory.AdvisoryStore` now implements `AdvisoryFreshness` (a populated store reports its count and the current time; an empty one reports zero), so the guard reads it correctly in in-memory dev mode as it already did with Postgres.

## Soundness

The in-memory store carries no per-advisory publish date (dev/test-only, re-loaded each process), so a populated store honestly reports "as fresh as this process's load"; an empty store reports a zero time + zero count, so the readiness guard still correctly treats an unpopulated store as no coverage (fail-closed under strict).

## Tests

`sca` and `memory` suites pass; `CGO_ENABLED=0 go build ./cmd/... ./internal/...` passes.

Refs: #860


nghiadaulau:
Addressed the Codex nit: the in-memory store now records the last-Upsert time and reports THAT as freshness (rather than time.Now() at each check), so the 'as fresh as its load' comment is literally true. Readiness-guard behavior is unchanged (populated ⇒ non-zero, empty ⇒ zero). Codex found no concrete defects.

PR #1067, opened 2026-09-12
docs(vulnintel): exclude internal implementation plan from nav

nghiadaulau:
## What

`TestEveryGuidePageIsPublished` fails on `main`: `docs/guide/vulnerability-intelligence-implementation-plan.md` is committed but absent from the MkDocs nav, so the drift guard flags it as an unpublished page.

That file is an internal implementation/tracking plan, not end-user guide content. It sits under `docs/guide` for co-location. Add it to `navExclusions` with a reason so "not in nav" stays a deliberate decision rather than an oversight, and the guard goes green.

## Why not add it to nav

The published guide is end-user documentation; an implementation plan does not belong in that navigation. Excluding it is the honest classification.

## Test

`go test ./docs/` passes.

### DIFF
--- a/docs/docs_publication_test.go
+++ b/docs/docs_publication_test.go
@@ -29,7 +29,11 @@ type mkdocsConfig struct {
 // navExclusions lists Markdown files under the guide directory that are intentionally absent from the
 // navigation. It is empty on purpose: every current guide is reachable. An entry added here must carry a
 // reason, so "not in nav" stays a decision rather than an oversight.
-var navExclusions = map[string]string{}
+var navExclusions = map[string]string{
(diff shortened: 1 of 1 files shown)


nghiadaulau:
Reviewed (CI disabled) — trivial, correct docs-nav fix: adds `vulnerability-intelligence-implementation-plan.md` to `navExclusions` **with a reason** (the map's contract requires every exclusion carry one), so `TestEveryGuidePageIsPublished` no longer fails on that internal plan co-located under docs/guide but intentionally out of the published nav. `go test ./docs/` passes. Merging. (Comment rather than a formal approve since this PR is authored by the account I act under.)

Issue #1049, opened 2026-09-12
reach(bench): reachability precision/recall corpus + gated ratchet vs OSS baselines


nghiadaulau:
Part of EPIC #1042 (section 5.1). Reachability accuracy benchmark, greenfield corpus.

**Scope / files owned:** a new benchmark harness mirroring `internal/infrastructure/tools/ast/owaspbench_integration_test.go`; a new labeled corpus dir. NOTE: `internal/usecase/accuracyeval` measures DETECTION accuracy, not reachability, so this corpus is new.
**Depends on:** none (harness); each language ratchet lands as its engine ships. **Blocks:** none.

Deliver: a labeled reachable/unreachable corpus per language scoring the four-state output, with a CI ratchet (recall is the gating metric), plus reproducible head-to-head vs OSV-Scanner call analysis and Semgrep CE on the same corpus, and a sampled manual Snyk comparison.

**Read/corpora:** NIST SARD / Juliet (samate.nist.gov); Securibench Micro; OWASP Benchmark; Reif et al. ISSTA 2019 (call-graph soundness eval).
**Acceptance:** gated CI test; recall floor only rises; parity-or-better vs OSV-Scanner + Semgrep CE recorded per language.

nghiadaulau:
Delivered in #1109 (squash-merged as 1052df2f). Acceptance met:

- **Gated CI test, recall floor only rises**: the reachability-benchmark workflow now has a Go job and a Python job; each runs the pinned OSS tools on the corpus fixtures and gates the owned engine on recall (monotonic floors.json ratchet).
- **Parity-or-better vs OSV-Scanner + Semgrep CE, recorded per language**: Go is scored against OSV-Scanner call analysis (govulncheck) and a Semgrep CE direct-call rule; Python against Semgrep CE (OSV-Scanner has no Python call analysis). The owned engine must meet or beat each baseline's positive-reachability recall AND precision on the same language subset. The Python head-to-head records the owned engine's precision win: Semgrep CE proves the os.system sink call site exists but not its entrypoint reachability, so it over-reports the 4 unreached fixtures (precision 0.556) while the owned call-graph engine is exact (1.0).
[...]

Issue #1042, opened 2026-09-12
EPIC: Reachability to best-in-market — dataflow-to-exploit-path, symbol-precise, sound-by-construction


nghiadaulau:
## Why this EPIC

Reachability is the one dimension where Synapse can be measurably strongest in market and charge for it: the industry does it partially, few do it soundly, and combining static reachability with runtime observed execution is a differentiator absent from the OSS SCA baselines this EPIC benchmarks. This EPIC takes Synapse reachability from "real but shallow" to a **static-plus-runtime, dataflow-to-exploit-path, symbol-precise engine across the major ecosystems**, backed by a curated vulnerable-methods database, with a coverage model that makes every suppression sound, proven by a gated benchmark.

Written after reading the real source of OSV-Scanner (and `golang.org/x/vuln`), Semgrep, and Snyk's public repos plus published methodology, then reviewed twice by an internal algorithm+architecture panel and by Codex against the live tree, and audited a second time for completeness. Findings are folded in. Every item carries the algorithm it needs and the official specs/papers to read.

The governing bar is absolute: **never hide a real vulnerability.**

## The suppression discipline (load-bearing, read first)

A verdict is either RAISE-ONLY (mint only `reachable`; raises urgency, never removes a finding; sound on imperfect analysis) or SUPPRESSING (mint `not_reachable` becoming OpenVEX `not_affected`; only sound with exhaustive, fail-open coverage). Rules for every item:

1. **Fail-open.** An analyzer that cannot fully observe the target mints nothing; the prior tier stands (`reachproof/coordinator.go:261-262`).
2. **No suppressing verdict on a graph blind to the construct in play.** Tightening a graph produces more `not_reachable` on the same blind spots (reflection, DI/`ServiceLoader`, `invokedynamic`, JNI, `//go:linkname`, `unsafe`, generic-entry, Python dynamic dispatch, JS `eval`/computed/monkey-patch). When an affected symbol's reachable surface uses such a construct, downgrade that symbol to raise-only or `no_analysis`.
3. **A negative requires positive coverage** (entry points non-empty AND coverage complete for the affected surface). Enforced by the new coverage model (0.6) and guard (0.5).
4. **State-aware supersession** (0.4): a weaker state must never shadow a proven `reachable` at equal tier rank.
5. **Runtime and taint signals are raise-only** and never enter `IsDeterministicReachabilityProof`.

## Call-graph precision decision (state it once)

[...]

PR #1069, opened 2026-09-12
feat(reach): add gated OSS reachability benchmark

VietSory:
## Summary

Implements the reproducible Go slice of EPIC #1042 / #1049.

* Adds a versioned labelled reachability corpus and recall ratchet.
* Adds real vulnerable-function controls for `GO-2026-4514` / `github.com/buger/jsonparser.Delete`: one called and one present-but-uncalled.
* Normalizes OSV-Scanner call analysis and Semgrep CE output into one digest-bound scorecard.
* Gates the owned SSA call graph on recall and precision parity-or-better against both OSS baselines.
* Adds pinned CI that uploads raw baseline output and normalized scorecards.
* Adds the maintainer-facing evidence runbook, including the credential-free Snyk review procedure.

## Evidence / fail-closed behaviour

* Candidate and baseline must have the same `corpus_digest`.
* Missing external analysis remains `no_analysis`; it is never treated as uncalled.
* Contradictory OSV data for one advisory/source fails closed.
* Missing baseline language, recall regression, or precision regression fails the gate.
* Owned Go engine: **4/4 exact, precision 1.000, recall 1.000**.
* OSV-Scanner v2.4.0 was executed locally against the corpus and reported `called=true` for the called control and `called=false` for the uncalled control.

## Validation

* `go test -count=1 ./internal/usecase/reachbench ./cmd/synapse-bench ./internal/infrastructure/tools/ssacallgraph`
* `go vet ./internal/usecase/reachbench ./cmd/synapse-bench ./internal/infrastructure/tools/ssacallgraph`
* `go build ./...`
* `go vet ./...`

## Snyk evidence

The strict `reachability-snyk-sample` adapter and review procedure are included. The actual sample is intentionally not fabricated or committed: a reviewer must export it from the approved Snyk organization and attach it as evidence without credentials or customer data.

Refs #1049

### DIFF
--- a/.github/workflows/reachability-benchmark.yml
+++ b/.github/workflows/reachability-benchmark.yml
@@ -1,0 +1,80 @@
+# Reachability scorecard (EPIC #1042 / #1049). The corpus and external runners are pinned so a PR cannot
+# improve a score by changing the tool, omitting an analysis record, or comparing a different denominator.
+name: Reachability Benchmark
+
+on:
+  push:
+    branches: [main]
+  pull_request:
+  workflow_dispatch:
+
+permissions:
+  contents: read
+
+jobs:
+  go-oss-baselines:
+    name: Go corpus vs OSV-Scanner and Semgrep CE
+    runs-on: ubuntu-latest
+    env:
+      # Reviewed external-tool pins. Updating either pin requires reviewing the resulting artifact diff.
+      OSV_SCANNER_VERSION: v2.4.0
+      SEMGREP_VERSION: 1.176.0
+      REACHBENCH_DIR: internal/infrastructure/tools/ssacallgraph/testdata/reachbench
+      REACHBENCH_REPORT_DIR: ${{ runner.temp }}/reachbench
+    steps:
+      - uses: actions/checkout@v7
+
+      - name: Set up Go
+        uses: actions/setup-go@v7
+        with:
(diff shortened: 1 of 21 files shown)


nghiadaulau (approved):
Reviewed locally (CI disabled) — strict pass. A gated OSS **reachability accuracy benchmark** — measurement infrastructure, off the production scan path (a separate `synapse-bench` CLI + `internal/usecase/reachbench`), comparing the owned reachability engine against OSV-scanner + semgrep-ce on a checked-in Go fixture corpus.

- **Anti-gaming scoring:** measures public, derived labels (engine-agnostic, not a specific tier); positive-reachability **precision AND recall** (can't buy recall by labelling everything positive); the corpus is fixed so a candidate can't omit a hard case to improve measured recall; conditional reachability counts positive (absent = a prioritisation false negative).
- **Monotonic ratchet:** `floors.json` — floors only rise; a language absent from floors is reported but not yet gated, letting a new engine land its corpus before a reviewed floor is calibrated.

`go build` + reachbench + synapse-bench tests pass. Note: it adds `.github/workflows/reachability-benchmark.yml`, but that workflow is inert while Actions is disabled — merging this does not turn CI on; it'll simply be there when/if you re-enable Actions. Nice, regression-proofing work. LGTM.

PR #1070, opened 2026-09-12
feat(reach): add Maven-scoped JVM Tier-2 call graph

### LINKED ISSUE #1056: reach(jvm): raise-only Tier-2 bytecode call graph (CHA + opt-in points-to)
Part of EPIC #1042 (section 3.2). JVM Tier-2 bytecode call graph, RAISE-ONLY.

**Scope / files owned:** `internal/infrastructure/tools/jvmreach/` (bytecode graph) + a new Tier-2 JVM usecase coordinator. Keep `ProofActorJVM*` OUT of the promotion/ suppression gate until suppression is proven sound.
**Depends on:** A-CANON (symbol match). **Blocks:** none.

Deliver: an owned bytecode call graph (CHA recall-first default; Andersen points-to, worst-case O(n^3) in variables, opt-in only). Mints Tier-2 `reachable` to raise urgency; does NOT drive `not_affected` because CHA/points-to are blind to reflection, DI (Spring/CDI), `ServiceLoader`, JNDI, `invokedynamic`, proxies, JNI, classloaders (the paths Log4Shell/Spring4Shell travel). Caveats: needs a built project (no source fallback); attribution Maven-coordinate-only (a Gradle jar without `pom.properties` is not attributable). Bytecode-level, so Kotlin/Scala/Groovy classes are covered like Java.

**Read:** JVM Spec SE 4/5/6 + invokedynamic/BootstrapMethods + LambdaMetafactory; ASM docs; Andersen 1994; Steensgaard POPL 1996; Lhotak-Hendren CC 2003 (Spark); Bravenboer-Smaragdakis OOPSLA 2009 (Doop); Livshits et al. APLAS 2005 (reflection).
**Acceptance:** a referenced affected method raises the finding; no `not_affected` minted; DI/reflection surfaces stay raise-only.

VietSory:
Closes #1056.

## Summary

This PR adds an owned, method-level JVM Tier-2 reachability pipeline for Maven-backed Java/Kotlin targets.

- Builds a JVM bytecode call graph directly from compiled `.class` files and shipped JARs.
- Uses Class Hierarchy Analysis (CHA) as the recall-first default dispatch strategy.
- Adds opt-in bounded receiver points-to refinement through `SYNAPSE_JVM_REACH_TIER2_POINTS_TO_ENABLED` (default: `false`).
- Falls back to CHA whenever receiver facts are incomplete or unknown.
- Requires exact Maven `group:artifact` attribution from `pom.properties` before raising a finding.
- Prevents same-named methods from unrelated dependencies from raising the wrong finding.
- Emits only positive `reachable` judgments; never emits or authorizes `not_affected`.
- Keeps reflection, dependency injection, ServiceLoader, JNDI, invokedynamic, proxies, JNI, and custom classloader paths raise-only.
- Composes Go and JVM reachability recorders independently so one engine's no-coverage result cannot block another.
- Wires JVM reachability into both API and worker scan paths.
- Preserves the judgment lifecycle and explicit startup-gate behavior.
- Adds focused bytecode, Maven-attribution, configuration, coordinator, recorder-composition, and API-gate tests.
- Documents the new JVM Tier-2 points-to operator setting.

## Validation

- Targeted `jvmreach`, `reachproof`, `sca`, `scacompose`, and `config` tests pass.
- Frontend, Agent Package Matrix, AI Triage Evaluation, and OWASP SAST Benchmark checks pass.
- Linux and Windows build/vet pass.
- JVM packages and all #1056-specific paths pass in both full backend runs.
- Remaining full-test failures are unrelated baseline or runner issues:
  - existing docs publication-nav omission;
  - pre-existing Postgres/response tests;
  - Windows Go 1.27 `crypto/ecdsa` runner access violation;
  - Windows filesystem/source-upload `Access is denied`.
- Lint reports 12 pre-existing findings in files outside this PR.

Final commit: `6e8b2c6e2fcab155123a502d67a8ab0fc941a25a`
Base: `3a8faca0862107bac5f1039e2af31434c242e98b`
Branch status: one commit ahead, no review threads or blockers.


nghiadaulau:
**Reviewed strictly (CI disabled). Nearly there — a lot of this is carefully done — but holding merge on a blocking DoS in the bytecode parser. The fix is small and local.**

The design is sound where it counts: the explicit-bounds cursor is solid (every `u1/u2/u4/skip/str` guards `off+n > len(b)` and latches `err`; `attrLen`/`codeLen`/switch counts re-validated → no panic path); zip-bomb is mitigated (every decompressed read is `io.ReadAll(io.LimitReader(rc, limit))`, not just entry-count); recursion is `seen`-guarded + nested-jar depth-capped to 1; zip-slip is impossible (entries read into memory, names used only for `BOOT-INF/`/`WEB-INF/` classification, no disk extraction); **raise-only is correct** (only ever appends `Reachable:true`, kept out of `IsDeterministicReachabilityProof`, source-only/not-built ⇒ no result — a failed/malformed analysis can never mint a `not_affected` suppression); and keeping a read-only bytecode *parser* in-process is defensible under ADR-0007 (it's not the compiler-grade `go/ssa` builder).

**Blocking — DoS from attacker-controlled `max_locals`:**
1. **Critical · `internal/infrastructure/tools/jvmreach/classparse.go:233,295-300`.** `parseCode` reads `maxLocals := int(c.u2())` with no ceiling (≤65535), handed to `localPointsTo` which does `make([]map[string]bool, maxLocals)` + `make([]bool, maxLocals)` + a loop initialising every map — **per method**. `methodCount` is also a `u2` and a Code attribute is ~26 bytes, so a ~1.7 MB class (under the 4 MB `maxClassBytes` cap) can declare 65535 methods × `max_locals=65535` ⇒ ~4.3e9 empty-map allocations ⇒ OOM-kill the worker. And there is **no `ctx` check inside the class parse** — `Analyze`/`scanTier2Workspace` only check `ctx.Err()` *between* entries, so the timeout can't interrupt one hostile class.
2. **High · `classparse.go:325,360-370`.** The points-to fixpoint is bounded by `maxLocals*maxLocals+1` (~4.3e9 rounds), and `instructionAt` is a linear scan used as array indexing → `storedReferenceSource`/`precedingReceiverLocal` become O(n²) in instruction count (millions for a 4 MB straight-line body). Same attacker field.
3. **Medium aggravator · `internal/platform/config/config.go:997`.** `SYNAPSE_JVM_REACHABILITY_ENABLED` defaults **true** and the Tier-2 coordinator is wired whenever set (`scacompose/jvm_tier2.go:28`) — so the in-process untrusted-bytecode parser runs by default for every Maven SCA scan, turning #1/#2 from lab-only into reachable-in-default-config.

**Fix (small, local):** clamp `max_locals` to a sane ceiling **before any allocation** (real methods are < a few hundred locals; reject/clamp well below 65535), and bound `methodCount × maxLocals`; bound the fixpoint by `maxLocals` (it converges in ≤ maxLocals rounds), and replace `instructionAt` with direct `ins[idx]` indexing; thread `ctx` into `parseTier2Class`/`localPointsTo` so the timeout actually fires; and keep the feature **default-off** until those bounds land (given it parses attacker-supplied binaries in the worker). Add a malformed-input case with a huge `max_locals`/method count to `tier2_test.go` to lock it.

Everything else is clean — build + jvmreach/reachproof/config tests pass here. Fix the `max_locals`/fixpoint bounds + default-off and I'll re-review and merge quickly.

nghiadaulau:
## Re-review — changes requested (one HIGH left)

Big improvement over the previous revision. The **parse phase is now genuinely hardened**: every attacker-controlled `u2`/`u4` count is clamped before it drives an allocation (`cpCount≤16384`, `methodCount≤1024` + a per-class `maxTier2MethodsTotal=250000` budget, `maxLocals≤512` + a `maxTier2LocalSlotsPerClass=16384` budget, `codeLen`/instruction/switch counts all bounds-checked), the fixpoint is correctly linear (`rounds < maxLocals`, not squared), `ctx.Err()` is threaded through every parse loop, and the config default is now `false`. Golden-rule posture is clean: propose-only + raise-only, a *distinct* verifier, JVM actors excluded from `IsDeterministicReachabilityProof` (so a claim can never mint OpenVEX `not_affected`), no exec/network/`x/tools`, zip-bomb-bounded reads. Nice.

One blocking issue remains — the DoS wasn't eliminated, it moved one phase downstream.

### 🔴 HIGH — the graph-build phase is uncancellable and superlinear over the same attacker-controlled input

`internal/infrastructure/tools/jvmreach/tier2.go` — `buildEdges` (`:310`), `resolveCall` (`:328`), and `reachableMethods` (`:417`) take **no `context.Context`** and check nothing. `Analyze` honors `ctx` only during `scanTier2Workspace` (the parse); once parsing finishes, `:77-78` runs `g.buildEdges(...)` → `reachableMethods(...)` with zero cancellation.

For a virtual call (`0xb6`/`0xb9`) with points-to disabled — **the default** (`NewTier2(false)`) — `resolveCall` at `:353` does `for name := range g.classes { g.isSubtype(...) }`, i.e. iterates *every* class (up to `maxTier2Classes=20000`) per call site, and `buildEdges` does that for every method of every class (budget `maxTier2MethodsTotal=250000`). Worst case ≈ 250k × 20k subtype walks ≈ 5×10⁹, each walking a super/interface chain — minutes-to-hours of pinned CPU, and `g.edges` can balloon toward `methods × classes` entries → OOM. It's invoked best-effort under the plain scan `ctx` with no dedicated deadline, and the hot loops don't observe `ctx` anyway.

**Failure scenario:** a crafted fat-jar (a normal SCA target) with ~5–20k classes sharing one method signature down a deep interface hierarchy, each emitting virtual calls to it. Parsing passes all caps; `buildEdges` then spins uninterruptibly / OOMs the API or worker. This is the exact "in-process DoS on attacker-controlled JVM bytecode" class from the last round, just relocated from the parser to the graph builder.

**Fix:** thread `ctx` into `buildEdges`/`resolveCall`/`reachableMethods` and check `ctx.Err()` periodically (every N methods in the `buildEdges` outer loop; every N nodes in the BFS), and wrap the whole JVM `Analyze` in a bounded `context.WithTimeout` at the coordinator/composition seam so the bound is a wall-clock guarantee rather than dependent on an upstream deadline the loops never read. Optionally cap total edges to fail closed.

### 🟡 MEDIUM — `localPointsTo` fixpoint: per-round cost unbounded by the slot cap, edges never deduped (points-to path, itself default-off)

`bytecode_tier2.go` — the round count is correctly bounded (`rounds < maxLocals ≤ 512`), but each round iterates `edges` (bounded only by instruction count ≤65535) and `applyStoreConstraint` (`:423`) appends inclusion edges with **no dedup**; inner work is `len(edges) × |pts[e.from]|` where a local's type-set can hold thousands of classes, and `ctx.Err()` is checked only once per round (`:400`), never inside the edges loop. A method within the 65535-byte code cap can push this to ~10⁹ map ops with no interruption. Bounded, but a real CPU amplifier that compounds the HIGH.

**Fix:** dedupe edges (a `seen` set keyed on `from<<16|to`) and add a periodic `ctx.Err()` check inside the edges loop; optionally cap distinct types tracked per method.

### 🟢 LOW — `.env.example` ships the feature ON

`.env.example:307` — `SYNAPSE_JVM_REACHABILITY_ENABLED=true` is uncommented, unlike the `# SYNAPSE_PYREACH_*=false` lines right below it. The code default is correctly `false`, but an operator copying the sample verbatim silently enables the DoS-capable path above. Comment it out (`# SYNAPSE_JVM_REACHABILITY_ENABLED=false`) to match the code default and the neighboring entries.

---

Verified locally against current `main`: affected packages build, `go test ./internal/infrastructure/tools/jvmreach/... ./internal/usecase/reachproof/... ./internal/usecase/sca/...` green, go-arch dependency rule clean (heavy parser stays in infrastructure behind the port; usecase never touches `x/tools`). CI stays off per repo policy. Fix the HIGH (and ideally MEDIUM + LOW in the same push) and I'll re-review and merge.

VietSory:
Addressed the re-review findings in `4d98bce`.

* Graph construction is now bounded and cancellable: `context` flows through edge building, virtual dispatch, subtype checks, and BFS; the JVM analyzer also applies a 15s wall-clock deadline over parse + graph phases.
* Added a global call-graph edge cap (`500,000`), which fails closed rather than retaining a partial graph.
* Hardened opt-in points-to further: inclusion edges are deduplicated and capped, tracked type sets are capped, and all over-limit cases conservatively fall back to CHA.
* Updated `.env.example` so JVM bytecode reachability is explicitly opt-in.

Validation completed:

* `go test ./internal/infrastructure/tools/jvmreach`
* targeted tests for `reachproof`, `sca`, `config`, `scacompose`, CLI, and API
* `go vet` on affected packages
* builds for `synapse-cli` and `synapse-api`

nghiadaulau:
## Approved — merging

The DoS is fully closed this round. Re-verified locally against current `main` (CI stays off per repo policy):

- **HIGH (graph phase) — fixed.** `ctx` is now threaded through `buildEdges`/`resolveCall`/`virtualDispatch`/`isSubtype`/`reachableMethods`, `Analyze` wraps both phases in `context.WithTimeout(…, 15s)`, and `isSubtype`/`virtualDispatch` check `ctx.Err()` on entry — so the CHA scan observes cancellation on effectively every class, not once per method. Added `maxTier2Edges=500_000` as a fail-closed cap.
- **MEDIUM (points-to) — fixed.** Inclusion edges are deduped (`edgeSeen`, key `from<<16|to`), bounded by `maxTier2PointsToEdges`, and the type-set is capped by `maxTier2PointsToTypes`; both caps degrade to a conservative CHA over-approximation (`limitHit` → all-unknown), and there's a periodic `ctx.Err()` inside the edges loop.
- **LOW — fixed.** `.env.example` now ships the flag commented out (`# SYNAPSE_JVM_REACHABILITY_ENABLED=false`).

Adversarial re-audit confirms: every attacker-scaling loop has either a periodic `ctx` check or a hard numeric cap; the 15s deadline can't be starved by a long inner stretch; new error returns propagate and a timeout degrades to raise-nothing (`coordinator.go` treats `DeadlineExceeded` as "no coverage, prior tier stands" — never a suppressing verdict); and all caps fail closed. The raise-only + `skipUnresolvedSubjects` contract holds, so silent `maxTier2Classes`/`methodsBudget` truncation can only *under*-report reachability, never mint a false `not_affected`.

Build + `go test ./internal/infrastructure/tools/jvmreach/... ./internal/usecase/reachproof/... ./internal/usecase/sca/...` green; dependency rule clean (parser stays in infrastructure behind the port, default-off).

One optional follow-up (non-blocking): the raise-only safety of that silent truncation is contingent on this analyzer's coordinator staying `raiseOnly=true` — worth a comment or a test pinning it so a future reuse under a non-raise-only coordinator can't regress into a false `not_affected`. Nice work on the two rounds of hardening.

Issue #962, opened 2026-09-10
Epic: enterprise identity and organization-scoped OIDC


pho-veteran:
## Planning baseline

Prepared on **2026-09-11** from repository commit `9d10499a`. The design was checked against the current domain, use-case, HTTP, OIDC, persistence, migration, deployment, audit, and web boundaries, then independently re-verified by two architecture/security reviews. This issue is the implementation plan; it does not itself change code, schema, or configuration.

## Problem

Synapse currently represents a human as one `users` row containing one tenant, role, API-key hash, and disabled flag (`internal/domain/user/user.go:85-97`). Human HTTP authentication resolves either a bearer token or an opaque browser session and then derives the tenant from that principal, not from request data (`internal/adapter/httpapi/auth.go:80-117`). Authorization remains the static `role -> permission` decision in `Role.Can` and the router's single `authz` wrapper (`internal/domain/user/user.go:60-81`, `internal/adapter/httpapi/router.go:346-355`).

Two verified defects make this more than a feature request:

- An unknown OIDC `(issuer, subject)` is provisioned automatically after group-to-role mapping (`internal/usecase/identitybff/service.go:99-164`), although ADR 0006 permits only an exact approved identity (`docs/adr/0006-oidc-bff-trust-model.md:15-23`). This EPIC deliberately reverses that behavior.
- Disabling a user changes a flag but does not revoke the user's key or browser sessions; re-enabling makes the same key usable again (`internal/usecase/users/service.go:257-282,411-423`).

Useful foundations remain: authorization state is consumed with tenant-bound `DELETE ... RETURNING` (`internal/infrastructure/persistence/postgres/identity_store.go:86-106`); session rotation inserts the replacement and revokes its predecessor atomically (`internal/infrastructure/persistence/postgres/identity_store.go:134-155`); and tenant transactions bind `app.current_tenant` locally and reject a nested tenant mismatch (`internal/infrastructure/persistence/postgres/tenant.go:55-90,124-133`). FORCE RLS maps an unset or empty tenant to no rows (`migrations/0057_row_level_security.sql:8-22,30-47`), while startup rejects a runtime role that can bypass it (`internal/infrastructure/persistence/postgres/tenant.go:152-174`).

[...]

PR #1119, opened 2026-09-14
feat(identity): establish D1-D5 human principal rollout foundation

VietSory:
Refs #962

Implements D1–D5 of the human-principal rollout:

- explicit authenticated human principal; no implicit operator fallback
- stable identity error contract, OIDC session handling, and privacy-safe access logging
- enterprise identity foundation with FORCE RLS, exact-hash credential locator, and immutable identity audit primitives
- bounded legacy-user backfill, OIDC shadow import, reconciliation, and rollout evidence ledger
- D5 legacy credential classification/projection, atomic locator updates, rollback-safe rotation, and cutover gate

Validation:
- Linux build and vet pass
- frontend tests, typecheck, and build pass
- D1–D5 PostgreSQL tests pass on PostgreSQL 17
- OWASP scorecard test passes

Known unrelated CI failures remain:
- repository-wide lint debt outside this diff
- two pre-existing full-suite failures in vulnerability/SLA and incident-response tests
- Windows vet / artifact-upload infrastructure failures


nghiadaulau:
## Review — changes requested (one MAJOR; otherwise excellent)

Impressive, disciplined foundation. Reviewed locally against current `main` with a full security audit + architecture audit (CI stays off per repo policy): `go build ./internal/...` clean, `go vet` clean on httpapi + identity, identity domain/usecase tests pass, web build + the 10 auth-client tests pass, and `TestMigrationInventoryNoDuplicateGooseVersions` passes (0177/0178/0179 sit cleanly above main's 0176 — no goose-version collision).

### What's clean
- **Security audit: CLEAN** (no critical/high/medium). All tenant tables in 0177/0178/0179 are `synapse_enable_tenant_rls` (ENABLE+FORCE + policy), incl. the `digest`-bearing `legacy_human_credentials`, with composite tenant FKs. `digest` is a SHA-256 **hash**, never a raw secret; placeholder-safety is enforced in *both* code and SQL — non-`issued` rows force `digest=""` and only `issued` credentials enter `credential_index`, so a placeholder/random hash can never become an auth credential. No secrets in logs/errors/DTOs; backfill is snapshot-frozen, drift-checked, idempotent, and fails closed; the four global tables are deliberately global with in-file justification. No new HTTP route consumes the identity/rollout/credential services yet (foundation-only), so live attack surface is minimal.
- **Architecture: PASS.** Domain pure, usecase depends only on domain + ports, adapter infra imports confined to tests, `cmd/*` DI-only, every new adapter carries a `var _ ports.X` assertion.
- The frontend auth-client change fails **closed** toward preserving credentials (only the explicit `authentication_invalid` code, or a bare 401 with no code, signs out — unknown future codes don't), which is the right posture for a rolling upgrade.

### 🔴 MAJOR — `users.Update` drops a role-change audit while committing the change (audit-atomicity, golden rule #6)

`internal/usecase/users/service.go:236` — `Update` records `user.updated` via `_ = s.audit.Record(...)`, swallowing the error, whereas the sibling mutations `createUser` (:186), `setDisabled` (:281) and `rotateAPIKey` (:327) all route through `recordUserAudit`, which **returns** the error (aborting the transaction) when `legacyCredentials != nil` — the D5 transactional path. `user.updated` carries `role`/`previous_role`, so this is exactly the security-relevant audit the struct's own comment (lines 37-44, 62-63) promises to commit atomically with the mutation ("source, derived credential, exact-hash index and mandatory audit share one commit"). As written, a failed `user.updated` audit is silently dropped yet the role change still commits.

**Fix:** call `s.recordUserAudit(ctx, ...)` here and return its error, identical to the other three mutations. (Please add a test that a failing audit backend aborts `Update` in the `legacyCredentials != nil` path.)

### Minor (fold in if easy; non-blocking)
- `internal/infrastructure/persistence/postgres/org_identity_store.go:232` — `identityWriteError` maps Postgres `23514` (CHECK violation) to `shared.ErrForbidden`; a check-constraint failure is a validation failure → map to `shared.ErrValidation` (avoids a misleading 403). The `23503`/FK→Forbidden cases are at least internally consistent; the `23514`→Forbidden one is the odd one out.
- `internal/adapter/httpapi/identity_error.go:71,90` — `writeIdentityError`/`writeIdentityFailure` take `ctx` as the second param (after `w`); reorder to ctx-first if you want strict idiom conformance (transport helpers, so minor).

### Forward-looking — must be honored by the auth-cutover PR (capturing here, not blocking this foundation)
1. `credential_index` carries no status and retains disabled-but-`issued` digests, so `ResolveExactHash` will resolve a **disabled** credential. Safe only if the auth consumer re-loads `legacy_human_credentials` and rejects `status <> 'active'` after resolving — the locator alone must never authenticate. Please make that an explicit requirement/test in the cutover PR.
2. `person_organization_index` is global/no-RLS but carries `tenant_id`+`person_id`; today it's FK-only with no read surface. Codify the "platform-only, never tenant-queryable" invariant with a guard test so a later PR can't quietly add a tenant-scoped read (cross-tenant membership disclosure).
3. `classifyLegacyCredentialEvidence` checks the issuance audit before the OIDC-link check, so if OIDC provisioning ever emits `user.created`, those users classify as `issued` (their random hash lands in `credential_index`). Not a bypass (no preimage), but it defeats the intended placeholder classification — prefer the OIDC-link signal, or add a test pinning OIDC-provisioned users → `placeholder`.

Fix the MAJOR (and ideally the two minors) and I'll re-review and merge — the data model, RLS, credential safety and rollout machinery are in great shape.

VietSory:
i'll am fixing it,please wait for me until i filnish it

nghiadaulau (changes requested):
Requesting changes — this cannot merge in its current state, and given the surface it touches it needs a green pipeline before a line-by-line review is worthwhile.

Blockers:

1. **Merge conflict with `main`.** `mergeStateStatus` is `DIRTY` (CONFLICTING). Please rebase onto the current `main` and resolve the conflicts. At +5732/-315 across 82 files this will need care; a conflicted diff also can't be reviewed reliably because the final shape isn't settled.

2. **CI is not green.** Only the GitGuardian check has run; the Backend (Go), Windows, Frontend, Lint, and PostgreSQL jobs are not passing on this head. The PR description itself lists known failures (repo-wide lint debt, two full-suite failures in vulnerability/SLA and incident-response, Windows vet/artifact-upload). For a change of this scope the branch has to go green — please either fix them or, for anything genuinely pre-existing on `main`, link the same failure on a clean `main` run so we can separate it from this diff.

3. **Tier-0 surface.** This is human-principal identity, OIDC sessions, RLS, credential locators, and immutable audit. That warrants full CI plus a focused security/architecture review, which I'll do once (1) and (2) are resolved and the diff is stable.

Once it's rebased and the pipeline is green, re-request review and I'll go through D1–D5 in detail. The direction reads well from the summary; the gate is mergeability and a clean run, not the concept.

Issue #1129, opened 2026-09-14
feat(cq): compute new_coverage + new_duplication server-side (+ fix stale gate-metric note)


nghiadaulau:
Part of #1120. Small wiring job — most of the coverage path already works server-side.

**Reuse (already exists):** `coverage` metric IS measured (`projectanalysis/analysis.go:363-365`); parser (3 formats) + `LineCoverage.NewCodePercent(changed)` (`coverage.go:195`); server-side changed lines `Analysis.FileChanges[].Added []LineRange`; `Analysis.Coverage` storage + measures read model + both ingestion paths.

**To do:**
- [ ] Convert `Analysis.FileChanges[].Added` → `map[string]map[int]bool` and set `MetricNewCoverage` in `buildMeasures` (`analysis.go:351`).
- [ ] Un-hardcode the snapshot `NewCodeCoverage` (`measure/snapshot.go:520-526`); add a changed-lines field to `BuildSnapshotInput`.
- [ ] Compute `new_duplication` over the same changed-line set.
- [ ] Correct the stale note at `qualitygate/metrics.go:26-31`; surface "no coverage data" explicitly (still fail-closed).
- [ ] Tests.

**Acceptance:** a `new_coverage`/`new_duplication` gate condition enforces real measured values. **Priority/Effort:** P0 / S.

thx2an:
I'd like to take this one, claiming here before starting. The "no silent gate" bar is the part I care about most — the current note's *"a `<=` new_duplication condition passes at 0"* is the same shape as the vacuous-pass fix in #579, and I'd rather it fail closed with an explicit reason than read 0.

Plan: feed `Analysis.FileChanges[].Added` into `buildMeasures` for `new_coverage`/`new_duplication`, un-hardcode `NewCodeCoverage` in the snapshot using the existing `DecimalMetric` availability model, correct the stale note, and tests for measured / no-data / boundary cases.

One thing I noticed while reading and will state up front: `NewCodePercent` lives on the infra `coverage.LineCoverage` type, and `buildMeasures` is domain, so the method needs a domain home (`measure.LineCoverage`) with the infra one delegating — no behaviour change, just the arch boundary.

I'll also take #1130 (same parser, same subsystem) as a separate PR unless someone else wants it.

Issue #1120, opened 2026-09-14
EPIC: Code Quality & ALM to best-in-market — PR decoration, coverage-complete gates, architectural metrics


nghiadaulau:
## Why this EPIC

Synapse's Code Quality subsystem is already a real, end-to-end, deterministic (LLM-free), owned SonarQube-class engine: A–E ratings on Security/Reliability/Maintainability with the SQALE debt-ratio model, quality gates with the Clean-as-You-Code new-code period, per-language quality profiles, the Bug/Vulnerability/CodeSmell/SecurityHotspot issue taxonomy with triage lifecycle, cyclomatic **and** cognitive complexity across ~15 languages, an owned token-based duplication detector, ~2,345 owned langpack rules over 14 language families, and coverage import for lcov/Cobertura/JaCoCo. It is wired through CLI, server, HTTP API, and the web dashboard, with multi-branch history/gate/new-code per project.

This EPIC consolidates every gap we assessed as **missing, half-wired, or weaker than the market**, so Code Quality reaches best-in-market on the two axes where competitors still lead: **ALM / PR integration** (SonarQube) and **architectural / behavioral analysis** (CodeScene), plus finishing the **coverage-on-the-server** path that is currently accepted-but-not-measured. It also folds in the small **reachability residuals** left open when EPIC #1042 closed.

The bar is the same discipline that governed #1042: owned + deterministic, no silent gate pass, prove it with a gated head-to-head benchmark.

## Sub-issues — pick one to work on

Each is single-PR-sized and links back here. Suggested order: **C → A+B → E / D2 / F → D1 / D3 / G**. Priority + effort + a reuse map are in each issue.

**A. PR decoration (P0 — biggest gap)**
- [ ] #1121 — PR decoration foundation (PRDecorator port + PR-identity fields + completion trigger)
- [ ] #1122 — GitHub adapter (commit status + check-run + PR comment)
- [ ] #1123 — GitLab adapter (pipeline status + MR note + code-quality report)
- [ ] #1124 — Bitbucket adapter (build status + Code Insights + comment)
- [ ] #1125 — wire into CLI `--decorate` + server auto-decorate

**B. Branch & PR model (P1)**
- [ ] #1126 — long-lived vs short-lived branch classification + retention
- [ ] #1127 — first-class pull-request analysis object (build with A1)
- [ ] #1128 — branch/PR selector + comparison in the dashboard

**C. Coverage (P0 — small)**
[...]

PR #1142, opened 2026-09-14
feat(cq): parse a Go -coverprofile natively

### LINKED ISSUE #1130: feat(cq): Go-native coverage (-coverprofile) parser
Part of #1120.

**Reuse:** `coverage.ParseBytes` format auto-detection (`coverage.go:43-56`); the `LineCoverage`/`CoverageReport` types feed everything downstream.

**To do:**
- [ ] Add a 4th `case` detecting the Go `mode: set|count|atomic` header; expand each `file:startLine.col,endLine.col numStmt count` block to per-line coverage (covered = count>0); emit the same `LineCoverage`.
- [ ] Tests (mode set/count/atomic; multi-file).

**Acceptance:** `go test -coverprofile` feeds the coverage gate with no external conversion. **Priority/Effort:** P1 / S.

thx2an:
Closes #1130. Part of EPIC #1120, workstream C. Independent of #1141 (#1129); the two share a package
and nothing else.

## The 4th `case`

`ParseBytes` detects the `mode: set|count|atomic` header and hands the data to `parseGoCoverProfile`,
which expands each `file:startLine.startCol,endLine.endCol numStmts count` block to every line it spans,
marking the line covered when `count > 0`. It emits the same `LineCoverage` every other format does and
goes through the same `NewCoverageReport` path, so nothing downstream changes — the issue's requirement.

A line spanned by several blocks — an `if` and its `else` body on one line, say — is covered if any of
them is. `mark` already implements that union (it is the rule merged reports rely on), so the parser
reuses it rather than reasoning about overlap itself.

## Two places this parser is stricter than the lcov one, on purpose

**A malformed line fails the parse.** `parseLCOV` skips a line it cannot read; this parser returns an
error. An lcov file is often hand-assembled from several tools and a stray line is noise; a coverprofile
is written by one tool in one pass, so a line that does not parse means the file is not what it claims to
be, and reporting coverage from the lines that did parse would present a partial profile as a whole one.
Unknown mode, wrong field count, missing position, span not `start,end`, position not `line.col`, end
before start, line zero, and a non-numeric or negative hit count each fail.

**Detection runs before the lcov fallback.** The lcov parser accepts any bytes and finds no `SF:`/`DA:`
records in a coverprofile, so without the new case a Go profile would parse "successfully" as an empty
report. `TestParseGoCoverProfileIsDetectedBeforeLCOV` first proves that precondition and then that the
profile is caught.

## Import paths — the thing the issue does not mention, flagged in my claim

Go names files by import path (`github.com/acme/app/internal/x/y.go`). `NormalizeLines` keeps only paths
that match the inventory exactly, so an unstripped profile would be dropped whole and `new_coverage`
would silently read as "no data" — the opposite of what #1129 is for.

`coverage.Options{GoModulePath}` strips `<module>/` from each file. `Parse` and `ParseBytes` keep their
signatures as thin wrappers over `ParseWithOptions` / `ParseBytesWithOptions`, so no existing caller
changes. `synapse-cli gate --coverage` reads the `module` directive from `<dir>/go.mod` via
`golang.org/x/mod/modfile` (already in the tree, used by `reachcache`) — any read or parse failure is
simply "no module", which keeps the import paths, never an error on a non-Go tree.

**Stated limit, not hidden:** the server-side upload path (`project_handler.go`) has no module path to
strip with, so a Go profile uploaded there keeps import paths and will not match the tree. The CLI guide
says so. Closing that gap needs either the module path recorded on the analysis or a suffix match in
`NormalizeLines`; both touch code the issue asked to leave unchanged, so I would rather you choose than
have me pick one inside this PR.

## Tests

- `TestParseGoCoverProfileStripsModulePathAndUnionsBlocks` — multi-file; module prefix stripped; a line
  under both a `count 1` and a `count 0` block is covered; lines no block spans are not recorded; totals.
- `TestParseGoCoverProfileModes` — `set`, `count`, `atomic`.
- `TestParseGoCoverProfileKeepsImportPathsWithoutAModule` — no module path, or a non-matching one, leaves
  paths verbatim.
- `TestParseGoCoverProfileFailsClosed` — the nine malformed shapes above.
- `TestParseGoCoverProfileIsDetectedBeforeLCOV`.

**Mutation-tested rather than merely green:**

| Mutation | Result |
| --- | --- |
| drop the Go `case` (fall through to lcov) | all five Go tests fail |
| treat `count >= 0` as covered | the union/`count 0` assertions fail |
| skip module-path stripping | the stripping tests fail |
| skip malformed lines instead of failing (lcov's leniency) | 5 of 9 `FailsClosed` subtests fail |

## Verification

- `gofmt -l` clean on every file touched; `go build ./...`; `go vet` over `coverage` and `cmd/synapse-cli`.
- `go test -count=1` green for `internal/infrastructure/tools/coverage`, `cmd/synapse-cli`,
  `internal/adapter/httpapi`, `internal/platform` (arch tripwire).
- `golangci-lint v2.13.2` (the CI pin): **0 issues**. `go.mod` is unchanged.


thx2an:
All four are real and are fixed in `da72443a`.

**Unbounded span expansion (critical).** Correct — and the one I should have caught, since it is the same
shape as the range-expansion bug fixed in #1141 the same day. `for ln := start; ln <= end; ln++` never
terminates when `end` is the maximum int, and nothing bounded the map. A block may now span at most
65,536 lines and a profile expand to at most 2^22; either excess is a parse error, consistent with the
parser refusing malformed lines, and the loop stops on `end` before incrementing. Tests cover
over-the-block-bound, over-the-profile-bound (many in-bound blocks), a block ending at `MaxInt`, and
exactly-at-the-bound parsing in full. With the old loop, the `MaxInt` case hangs.

**Negative statement count.** Correct; `numStmts` is now validated like the hit count.

**`go.mod` file type.** Correct; `goModulePath` now checks for a regular file before reading, the guard
`reachcache/fingerprint.go` already applies to manifests. One honest note on the test: I first used a
directory named `go.mod` as a stand-in and it did not exercise the guard — `ReadFile` fails on a
directory without it. The test now creates a real FIFO and asserts under a deadline; removing the guard
makes it time out.

**CLI-level wiring test.** Correct that nothing would have caught a regression to `Options{}`.
`TestRunGateStripsGoModulePathForNewCodeCoverage` builds a two-commit git repo with a `go.mod`, feeds a
profile keyed by import path, and asserts a `coverage >= 50` floor passes under `--new-code-only` —
which it can only do if the module prefix was stripped so the keys match the diff. It then rewrites
`go.mod` to a module the tree does not declare and asserts the gate fails. Reverting the CLI to
`coverage.Parse` fails the test.

Re-verified: `gofmt` clean on touched files, `go build ./...`, `go vet`, tests green over
`internal/infrastructure/tools/coverage`, `cmd/synapse-cli`, `internal/adapter/httpapi`,
`internal/platform`; `golangci-lint v2.13.2` 0 issues.

nghiadaulau (approved):
Reviewed: build/vet/lint/tests verified locally on the PR branch, green. Safe to merge.

PR #1245, opened 2026-09-15
feat(cq): add coverage-aware complexity rollups and analysis deltas

### LINKED ISSUE #1133: feat(cq): complexity rollup to file/module + trend
Part of #1120.

**Reuse:** per-function cyclomatic + cognitive already computed (`astwalk`); measures tree + `Delta` trend already exist.

**To do:**
- [ ] Aggregate per-function complexity to file/module nodes; surface trend vs previous analysis.
- [ ] Tests.

**Acceptance:** complexity shows at file/module level with a trend delta. **Priority/Effort:** P2 / S–M.

XUanhoa04:
## Summary

Closes #1133
Refs #1120

Project Code Quality now has truthful complexity rollups and immutable trend evidence. Function-level AST metrics aggregate to files, directories, and the project root only when per-file parse coverage proves the value. Compatible analyses persist signed per-path deltas, baseline provenance, and coverage so reductions remain visible and never become false zeroes.

## Changes

- Added versioned, bounded complexity evidence indexing with canonical-path, overflow, duplicate-coverage, unsupported-language, and parse-error handling.
- Added measured/eligible complexity coverage summaries and fail-closed rollups to the immutable measure snapshot.
- Added signed cyclomatic/cognitive deltas with baseline analysis ID, timestamp, source ref, branch/source compatibility checks, and explicit unavailable reasons.
- Cloned complexity delta maps in the memory store; PostgreSQL keeps the new fields in the existing JSON payload.
- Extended the Measures API/OpenAPI contract and dashboard with delta columns, coverage, baseline provenance, and signed formatting.
- Added domain, persistence, usecase, mapper, and dashboard regression tests plus guide/CHANGELOG documentation.

## Validation

- Targeted Go tests: pass (measure, projectanalysis, projectuc, memory persistence, AST adapter/walker, OpenAPI parser).
- go vet on changed Go packages: pass.
- go build ./cmd/...: pass.
- cd web && pnpm run typecheck: pass.
- cd web && pnpm test: pass (124 files, 735 tests).
- cd web && pnpm run lint: pass with 9 pre-existing warnings.
- cd web && pnpm run build: pass.
- go test ./... reaches the changed packages successfully; the Windows run still has unrelated pre-existing failures in Mkfifo/symlink privilege, runtime sensors, gobinreach, and IaC path fixtures.

## Checklist

- [x] OpenAPI and guides updated
- [x] Safety invariants preserved: deterministic, bounded, append-only, fail-closed evidence
- [x] docs/plans/ intentionally excluded from the commit


nghiadaulau (changes requested):
Reviewed at head `b60749da` in an isolated worktree (build + `CGO_ENABLED=0` + tests + lint + drift guards + web typecheck/vitest). Build, tests, and drift guards are all green, and much of this is done well: the baseline is a committed prior analysis (not working-tree state), signed deltas survive JSON round-trip, genuine measurement gaps fail closed instead of becoming false zeroes, the store deep-copies delta nodes, and the DTO/OpenAPI/UI are wired and drift-clean. One blocker keeps it from merge.

## Blocker: the project/directory rollup goes unavailable on any polyglot repo (a regression)

The coverage rollup marks a directory or the project root `available` only when **every** eligible descendant file is measured (`internal/domain/measure/snapshot.go:571-579`: parent available iff `MeasuredFiles == EligibleFiles`). Every inventory file node is stamped `EligibleFiles: 1` unconditionally (`snapshot.go:318`), but the complexity metrics come from the tree-sitter sidecar, which has grammars for 17 languages only (`internal/infrastructure/tools/astwalk/parse_cgo.go:67-201`: Python, JS/TS/TSX, Java, Go, C, C++, C#, Rust, Ruby, PHP, Scala, Swift, Kotlin, CSS, HTML).

The code inventory keeps every Programming and Markup file (`internal/infrastructure/codeinventory/analyzer.go:108`), so any file outside those 17 grammars (Shell, Makefile, Dockerfile, YAML, SQL, Perl, Lua, Terraform, ...) becomes eligible-but-unmeasured (`complexity_coverage_missing`). One such file anywhere in the tree drags its directory and the project root to `MeasuredFiles < EligibleFiles` → complexity and its trend delta render `—`.

This regresses base behavior. On `main`, when the sidecar ran (`in.Complexity != nil`), file nodes kept the default `ComplexityAvailable: true` (`snapshot.go:228`, `git show origin/main:internal/domain/measure/snapshot.go:296-312`), so the project root stayed available and simply summed complexity over the measurable files. The PR flips the default to unavailable and gates on full coverage, so the project-level number this PR is built to add is unreachable on real polyglot repos, synapse-ce itself included (it has a `Makefile` and shell scripts). The CHANGELOG line "rolls up to files, directories, and the project root" and the Behavioral/Measures UI would show `—` at the root here. It is fail-closed and documented, so it does not create a false zero, but a language the tool does not analyze is `not_applicable`, not a measurement gap, and blanking the whole project over it makes the headline capability a scaffold.

**Fix:** distinguish "language out of scope" from "measurement gap". Exclude out-of-scope-language files from the eligible denominator (or classify them `not_applicable` and skip them in the rollup at `snapshot.go:318` / `:571`), so aggregates stay available over the measurable subset. Keep the fail-closed path only for a genuine parse failure on a supported-grammar language (a `.go`/`.py` the sidecar tried and failed) — that should still make the parent unavailable. Ship a test with a mixed Go + shell inventory asserting the project root reports complexity available over the Go subset.

## Secondary (fix in the same change)

Non-deterministic aggregate reason string (`snapshot.go:555-566`): `children` is built by ranging the `nodesByPath` map, so when a directory has multiple unmeasured children with different reasons the surfaced `unavailable_reason` varies run to run. Numeric fields stay deterministic; only the persisted diagnostic string drifts. Sort children by path before selecting the reason, or always prefer the `N_of_M_files_unmeasured` summary.

## Verified sound (no change needed)

- Baseline selection: `service.go:571-594` uses the previous stored analysis on the same recording branch; empty branch or mismatched `SourceRevision.Kind` fails closed to `unknown_branch` / `incompatible_source`; legacy coverage-version-0 baselines fail closed to `incompatible_complexity_version`. Not working-tree state.
- Signed deltas preserved (reductions survive; openapi `value: [integer,"null"]` has no `minimum`), overflow-guarded (`MaxInt32`), no divide-by-zero (`EligibleFiles==0 → "no_complexity_files"`).
- Producer stamps `ComplexitySchemaVersion`; store deep-copies `Delta.Complexity.Nodes`; DTO + `api/openapi.yaml` + UI (signed `Δ`, AST coverage badge, `vs <ref>` provenance, `—` for unavailable) wired and drift-clean.
- Build, CGO0 build, vet, lint, affected tests, docs/api drift guards, web typecheck, and the two changed web test files (23 passing) all green.

Fix the rollup denominator so the project-level metric is reachable on polyglot repos, fold in the determinism fix, and this is an approve. The rest of the vertical slice is solid.

XUanhoa04:
Implemented the requested changes and pushed them in `d945a904`:

- Unsupported AST languages are explicitly not applicable and excluded from the eligible denominator, so mixed Go + Shell repositories retain project/directory complexity over the supported subset.
- Supported-language parse failures still fail closed.
- Child nodes are sorted by path before choosing an aggregate unavailability reason.
- Added mixed-language and deterministic-reason regression tests.

I also pushed `0a599b64` to restore the currently failing CI baseline:

- Excluded Unix-only FIFO/runtime/binary tests from Windows where appropriate.
- Fixed Windows directory-sync and canonical slash-path behavior.
- Fixed the full golangci-lint v2.13.2 finding set.
- Serialized packages in the shared-Postgres CI test step to prevent fixture resets from racing across packages.

Local verification:
- `go test -count=1 ./...` (Windows): pass
- `go vet ./...` (Windows): pass
- `golangci-lint v2.13.2 run`: 0 issues
- `GOOS=linux GOARCH=amd64 CGO_ENABLED=0 go build ./...`: pass

The branch is also merged with current `upstream/main`.

nghiadaulau:
## Review — changes requested (two correctness/perf issues in the domain accessor)

Nice slice overall — reviewed locally against current `main` (CI stays off per repo policy): dependency rule is clean (the rollup + coverage + delta math all live in `domain/measure` + `domain/projectanalysis`, pure), base cases are handled well (coverage rollup guards `EligibleFiles == 0` before the ratio; "no previous analysis" degrades to `no_previous_analysis`; delta maps are nil-safe and consumed by keyed lookup so iteration order never reaches output), memory/postgres stay consistent (postgres persists the whole `Analysis` as JSON so `Delta.Complexity` round-trips; the memory clone deep-copies `Delta.Complexity.Nodes`), and the web mapping reads snake_case tags correctly and keeps "unavailable" distinct from a real `0`. `go test` + `go vet` green on the touched packages. Two issues worth fixing before merge, both in the rewritten complexity accessors.

### 🟠 MAJOR 1 — O(files²) regression: the whole index is rebuilt on every per-file call

`internal/domain/measure/complexity.go:112` (`FileCyclomatic`) and `:125` (`FileCognitive`) were rewritten to call `r.ComplexityIndex()`, which rebuilds the entire per-file index (O(files + functions)) on **every** invocation. `internal/domain/measure/behavioral_hotspots.go:193` calls `FileCyclomatic(canon)` once per inventory file inside the walk loop, so the whole pass becomes O(files² + files·functions) where it was O(files·functions). On a large repo the added `files²` term is a real slowdown on a hot analysis path.

**Fix:** build the index once before the loop (`idx, err := in.Complexity.ComplexityIndex()`) and read `idx[canon]`, or memoize the index on the report so repeated accessor calls are O(1).

### 🟠 MAJOR 2 — one malformed entry silently poisons the whole tree

`ComplexityIndex()` returns an *error* if **any** coverage/function path is non-canonical or duplicated, and both `FileCyclomatic`/`FileCognitive` collapse that error to `(0, false)`. So a single bad entry anywhere makes every file report `!measured`, which in `behavioral_hotspots.go` becomes a blanket `complexity_not_measured` gap for the entire tree. The previous linear scan localized the damage to the one offending file. `BuildSnapshot` validates evidence up front so the snapshot path is safe, but `behavioral_hotspots` can be handed a report that never went through `ValidateComplexityEvidence`.

**Fix:** have the accessors distinguish "this file is bad/absent" from "the report is structurally invalid" (e.g. return the index+error to the caller once, or validate at report ingress and treat the accessors as total), so one malformed path can't blank out every file's complexity.

### Minor (fold in if easy)
- Exported symbols missing doc comments (revive/golangci consistency): `usecase/projectuc/measures_dto.go:76` `ComplexityCoverageMetric`, `:84` `ComplexityBaseline`; `domain/measure/complexity.go:9` `ComplexitySchemaVersion`; `domain/projectanalysis/analysis.go:42` `ComplexityDeltaSchemaVersion` — the neighboring new types are documented.
- `ComplexityIndex` skips the `Cyclomatic < 0` guard for functions whose file is absent from `Files`, so `ValidateComplexityEvidence` doesn't fully reject a malformed wire model (harmless today — those values are never summed).

Fix the two MAJORs and I'll re-review and merge — the layering, math base-cases, and persistence round-trip are all in good shape.

XUanhoa04:
Addressed all feedback from the review:

1. **MAJOR 1 (O(files²) regression)**: `BuildBehavioralHotspots` now indexes complexity once up front via `in.Complexity.ComplexityIndex()` and performs O(1) map lookups inside the inventory loop (`e39a0bfb`).
2. **MAJOR 2 (Silent tree poisoning & damage localization)**:
   - Behavioral hotspots validates complexity evidence at report ingress (`in.Complexity.ValidateComplexityEvidence()`), failing fast with an explicit error on malformed evidence rather than silently degrading all files to unmeasured.
   - `FileCyclomatic` and `FileCognitive` use a localized linear scan so a malformed or duplicate entry elsewhere in the report does not invalidate unaffected files. Added `TestFileCyclomaticAndCognitive_LocalizedDamage`.
3. **Minors**:
   - Added doc comments on exported symbols (`ComplexityCoverageMetric`, `ComplexityBaseline`, `ComplexitySchemaVersion`, `ComplexityDeltaSchemaVersion`).
   - Moved negative metric checks in `ComplexityIndex` before the file existence check and added `TestComplexityIndex_RejectsNegativeMetricForAbsentFile`.
4. **CI & Integration**:
   - Merged latest `upstream/main` (`70c6dc96`).
   - Aligned test build tags with runtime OS guards to resolve staticcheck SA4032 (`1e697c1a`).

All CI checks (Lint, Frontend, Backend Linux, Backend Windows, and all benchmarks) are now 100% green.

nghiadaulau (approved):
Approve. Core complexity-rollup/trend logic is coverage-honest and well-tested (ComplexityIndex + snapshot ComplexityCoverage: canonical-path + duplicate + overflow guards, no false zeros, legacy/unsupported/parse-error fail closed with explicit reasons). Independently verified: build clean, measure + projectuc suites pass; all CI green.

Two non-blocking notes for a follow-up: (1) internal/domain/measure/behavioral_hotspots.go is not gofmt-clean (the new BehavioralHotspotsSchemaVersion const misaligns the block; CI's lint doesn't run gofmt) — a gofmt -w fixes it; (2) three unrelated fixes are bundled here (ci.yml go test -p 1, sla/assessment.go microsecond truncation, a taintrules //nolint) — worth splitting per one-change-per-PR, and -p 1 serializes all package tests, so scoping it to the Postgres package would keep CI faster.

Issue #1043, opened 2026-09-12
ci: add reusable benchmark workflows and lifecycle controls


pho-veteran:
Parent EPIC: #1034

Coordinates GitHub Actions delivery for #1035, #1036, #1037, #1038, #1039, and #1040. Domain issues continue to own benchmark logic, corpus semantics, soundness, and accuracy thresholds; this issue owns workflow orchestration, trust boundaries, cost controls, final evidence, and shutdown lifecycle.

## Goal

Provide capability-named, reusable GitHub Actions workflows that run while #1034 is being implemented, make a green final run on one exact `main` commit an EPIC acceptance gate, then stop all automatic/manual execution by disabling the workflows while retaining their YAML and supporting code in the repository for future re-enablement.

## Naming and hygiene

Workflow identity is capability-based and must not include an EPIC or issue number in filenames, display names, job IDs, artifact names, schemas, or reusable interfaces.

Initial workflow inventory:

- [ ] `engine-accuracy.yml` — independent SCA oracle and pinned competitor comparison for #1035.
- [ ] `reachability-benchmark.yml` — per-language reachability corpus and nightly/full evaluation for #1038.
- [ ] `sast-benchmark.yml` — generalized OWASP/Juliet/Securibench/Semgrep evaluation for #1039; evolve the existing OWASP workflow rather than retaining duplicate responsibility.
- [ ] `security-accuracy.yml` — secrets, IaC/misconfiguration, and host-CVE evaluation for #1040.
- [ ] `dynamic-security-benchmark.yml` — trusted DAST/CSPM evaluation for #1040.
- [ ] `performance-benchmark.yml` — repeatable throughput/latency/memory measurement for #1040.
- [ ] `owned-default-readiness.yml` — final hard gate supporting the owned-only switch in #1037.

Any material consolidation or split of this topology must preserve capability-based identity and be recorded through the ADR process below.

## Active lifecycle while the EPIC is open

- [ ] PR-safe workflows use `pull_request` without credentials and run only code/data safe for untrusted forks.
- [ ] Relevant workflows run on push to `main`, scheduled full evaluation, and `workflow_dispatch` as appropriate to their cost and trust boundary.
- [ ] DAST/CSPM jobs run only from trusted refs through schedule/manual dispatch and protected environments; they never execute untrusted PR code with credentials.
- [ ] Performance comparisons run on an equivalent runner/environment identity and do not claim cross-hardware comparability.
[...]

PR #1261, opened 2026-09-18
feat(reachability): expand sound Tier-2 coverage

### LINKED ISSUE #1038: feat(reachability): expand sound Tier-2 coverage and gate its accuracy
Parent EPIC: #1034

Maps EPIC work: **D4.7 + D4.8 + A4 + C2 + feasible raise-only D4.6**.

## Goal

Expand function/export-level reachability without introducing false suppression, and measure reachable/unreachable/unknown behavior per language with committed regression floors.

## Scope

- [ ] Add Go-binary symbol evidence from `.gopclntab` and available symbol metadata, normalized to owned affected-symbol identities.
- [ ] Add bounded JavaScript interprocedural affected-export paths using parameter binding, return flow, higher-order calls, closures, and re-exports.
- [ ] Build an embedded or content-pinned reachability corpus per supported language with reachable, unreachable, opaque, and no-coverage cases.
- [ ] Add precision/recall/unknown reducers and CI ratchets; compare with external tools only where licensing, automation, and equivalent semantics permit.
- [ ] Extend Tier-2 coverage to the highest-value gaps identified by the corpus.
- [ ] Provide PHP/Ruby raise-only analyzer input through the existing provenance-backed `SymbolOverlay`; ship production signals only when curated affected-symbol data exists.
- [ ] Keep every analyzer raise-only unless complete enumeration and all language-specific opaque constructs can be proven.

## Guardrails and non-goals

- Symbol absence in a Go binary is no-coverage, never `not_reachable`.
- Lexical or bounded PHP/Ruby/JS symbol analysis cannot mint OpenVEX `not_affected` unless a separately reviewed completeness proof exists.
- CommonJS mutation, dynamic import/require, computed access, re-export/package-exports ambiguity, closure escape, prototype paths, side-effecting getters, aliases, and skipped/generated/minified files must force Unknown for suppressing analysis.
- Do not fabricate affected symbols when feeds do not publish them.

## Acceptance criteria

- [ ] Presence of a matched affected symbol in a Go binary raises finding urgency; absence never creates a negative judgment.
- [ ] Go tests cover stripped/no-pclntab binaries, inline/generic naming, vendored/module paths, plugins/shared objects, malformed binaries, and panic containment.
- [ ] JavaScript proves raise-only paths through two-hop, higher-order, returned-closure, and re-export fixtures.
- [ ] Every opaque JavaScript construct in the guardrail list prevents a suppressing verdict.
- [ ] Per-language evaluation reports reachable precision/recall, suppressing precision/recall where applicable, Unknown rate, and coverage rate separately.
- [ ] Raise-only ProofActors are excluded from deterministic negative reachability and physically cannot mint OpenVEX `not_affected`.
- [ ] Any PHP/Ruby production overlay entry records advisory ID, symbol identity, source URL/revision, curator, and review provenance.

## Dependencies

None for Go/JavaScript implementation and corpus work. PHP/Ruby production coverage depends on provenance-backed affected-symbol curation.

## Delivery contract

- This issue may be delivered through multiple reviewable PRs. Every PR must reference this issue and parent EPIC #1034.
- Preserve #1034's no-false-suppression and no-fabricated-data guardrails. Unknown or incomplete coverage must remain explicit rather than being converted into a clean result.
- If implementation changes an architectural boundary, invariant, scope, benchmark methodology, dependency, guardrail, or acceptance criterion, post an ADR-style comment on #1034 before merging the affected change. The comment must include **Context**, **Decision**, **Alternatives considered**, **Consequences**, and links to this issue and the relevant PR; link that EPIC comment back here.
- Ship behavior changes with tests at the narrowest useful layer. Update configuration documentation, OpenAPI, and user-visible release notes when those surfaces change.
- Before closure, request an independent **Fable coverage review** against this issue and #1034. Post the review result as a comment here, covering implemented criteria, missing coverage, adversarial cases, and remaining risks.
- Incorporate actionable Fable insights into this issue as checklist/acceptance updates or create linked follow-up issues with an explicit disposition before closing.

## Issue-definition coverage review amendments

- [ ] JavaScript suppression remains opt-in and never becomes the default, even if a reviewed completeness gate passes; missing-symbol results cannot become negatives without the full opaque-construct gate.
- [ ] PHP/Ruby **symbol-level** suppression and Rust/Ruby source suppression remain blocked in this scope. Complete package-reference enumeration is not proof that a library-internal affected function is unreachable.
- [ ] Build and wire both PHP and Ruby observed-symbol indexes and Tier-2 raise-only analyzers now, including coordinator integration and actors that are self-confirm/raise-only and excluded from deterministic negative proof. Curated external symbol data blocks production coverage, not engine construction.
- [ ] Positive-reference and missing-data/no-coverage tests cover both PHP and Ruby analyzers; neither analyzer can emit `not_reachable` or OpenVEX `not_affected`.
- [ ] Explicitly disposition the separate PHP AST/import-level suppression path described by #1034 as delivered here or as a linked follow-up; do not conflate it with symbol-level suppression.
- [ ] Add A4's nightly reachability accuracy job over the pinned corpus.
- [ ] Closure requires measured C2 improvement over a committed baseline in a predeclared reachability breadth/accuracy metric, with Unknown and coverage rates reported separately and no regression of suppressing precision.

## GitHub Actions delivery

- Workflow orchestration and lifecycle are owned by #1043 through capability workflow `reachability-benchmark.yml`.
- While #1034 is open, PR-safe soundness/adversarial gates and the scheduled/manual full language corpus run according to #1043's trust and cost controls.
- Closure requires the full reachability scorecard, Unknown/coverage accounting, nightly/full evidence, and no-false-suppression invariants to be green on the exact final EPIC `main` SHA.
- After final acceptance, `reachability-benchmark.yml` remains committed but becomes `disabled_manually`; future work must explicitly re-enable it before claiming new measured reachability results.

pho-veteran:
## Problem

Synapse's production reachability surfaces had useful positive evidence, but several bounded paths were missing or too coarse: Go binary entry calls, JavaScript aliases/callbacks/returned callables, and PHP/Ruby symbol provenance. The project also lacked one trusted, repeatable acceptance lifecycle that could measure those changes without converting uncertainty into a false suppression.

## Approach

- extend the production analyzers with bounded positive-evidence paths while preserving fail-closed behavior for unknown, dynamic, incomplete, or unsupported inputs
- route SCA and reachability benchmarks through shared bounded two-pass lifecycle mechanics, while keeping their domain contracts separate
- freeze an 83-case, 19-cohort corpus and execute 95 production cells twice with exact source, fixture, analyzer, configuration, and authority identities
- keep normal production coordinators raise-only; separately exercise four candidate-only suppression projection controls through fresh coordinators with distinct proposer and verifier identities
- derive candidate authority from the exact checkout commit and tree, publish only sanitized digest-bound projections, and restore PR-safe OSV-Scanner and Semgrep CE comparisons

## Alternatives rejected

- Treat partial or opaque evidence as complete: rejected because it could turn an analyzer blind spot into `present_unreached` and a suppressing projection.
- Rebuild reachability as a second standalone benchmark framework: rejected in favor of reusing the shared bounded lifecycle while retaining reachability-specific semantics.
- Require operators to supply commit SHAs or synchronize helper stages manually: rejected; identity, authority inputs, and lifecycle state are derived automatically from the checkout and trusted controller inputs.

## Verification

Exact head: `cc7268f98dbfba214382d144c209c3616f727699`  
Exact tree: `810cc8f07e0cdb25f1ecdba376d97442e0d2cb77`  
Integrated `origin/main`: `a854487a4f02ddfaca0743afad517eaf34f9133c`

- `make lint build vet test typecheck` — passed
- `CGO_ENABLED=0 go build ./cmd/...` — passed
- changed reachability and Python AST packages with `go test -count=1` — passed
- current-main integration preserved bounded-dispatch observation before display-field early return; focused AST and reachbench suites passed
- `cd web && pnpm test` — 125 files, 739 tests passed
- `cd web && pnpm build` — passed
- `actionlint` with the repository's two custom trusted-runner labels declared — passed
- exact-head publication verifier — 18/18 artifacts matched; lifecycle commit/tree matched; candidate, safety, repeat, and all eight suppression projection checks passed
- final bounded confirmation — passed with no reproducible defect in the 18 accepted fix areas; full build/test and uncached focused suites passed

### Reachability scorecard

| Metric | Protected baseline | Candidate |
|---|---:|---:|
| C2 production breadth | 3 | 13 |
| C2 correct positive cases | 5 | 22 |
| C2 macro reachable recall | 5/38 | 32/57 |
| Reachable precision | 5/6 | 27/29 |
| Reachable recall | 1/9 | 3/5 |
| Complete / partial / unavailable coverage | 6 / 7 / 82 | 23 / 19 / 53 |
| No-analysis observations | 83 | 54 |
| False / invalid suppressions | 0 / 0 | 0 / 0 |

Candidate acceptance passed with 95/95 observations in each repetition, deterministic semantic equivalence, and safety pass. Sanitized report ID: `sha256:97152ba54212df6cb5839a3371aee92c092db3fc01407f9c408b2d650ff9f3bc`.

The production scorecard intentionally reports suppression as not applicable because normal capture remains raise-only. Suppression safety is exercised independently by the four exact negative controls in each repetition; all eight projected `present_unreached` evidence records passed identity, tier, coordinator, and sanitization checks.

## Risk and rollback

The main risk is an analyzer overstating completeness or allowing benchmark-only suppression behavior to leak into production capture. The implementation keeps uncertainty subject-local where bounded, analysis-wide where unbounded, and preserves `.WithRaiseOnly()` on normal production coordinators. Candidate acceptance fails closed on missing executions, coverage overclaims, precision/recall regression, suppression errors, source-identity drift, or semantic-repeat mismatch.

Rollback is a normal revert of this branch. There are no migrations, new HTTP routes, production infrastructure mutations, or required configuration changes.

## Checklist

- [x] Tests added or updated for behavior changes
- [x] Exact-head trusted candidate lifecycle accepted
- [x] Safety, semantic repeat, and suppression projection conformance passed
- [x] Generated benchmark runs and private review material kept out of Git
- [x] No changelog entry requested

Closes #1038  
Refs #1034  
Refs #1043


pho-veteran:
## Exact-head reachability benchmark evidence

Authoritative candidate lifecycle completed on:

- commit: `ddd6acecf9ab45fdadd7d13905211aca8c783dd1`
- tree: `151f7e837b5bccea5a824eedd77bbd9b3bf3bf40`
- required executions: `95`
- observed executions: `95`
- repetitions: `2`
- semantic repeat: `true`
- safety: `pass`
- candidate accepted: `true`

| Metric | Protected baseline | Candidate |
|---|---:|---:|
| C2 production breadth | 3 | 7 |
| C2 correct positive cases | 4 | 14 |
| C2 macro reachable recall | 2/19 | 9/38 |
| Reachable precision | 1/1 | 1/1 |
| Reachable recall | 4/45 | 14/45 |
| Complete coverage | 6 | 11 |
| Partial coverage | 7 | 14 |
| Unavailable coverage | 82 | 70 |
| No-analysis observations | 83 | 73 |
| False suppressions | 0 | 0 |
| Invalid suppressions | 0 | 0 |

Candidate report identity:

`sha256:e1c70540c3a85ebae72a0d2d74bf68bfc486e94b96ff300f6b334ab4238698d9`

The published evidence was exported from the Linux-backed atomic publication directory. Raw/private benchmark state and its Docker volume were removed; generated run artifacts are not committed.

pho-veteran:
Exact-head benchmark and local gate evidence for `1cfa750b3e61d7d9702d078273d07a3b9fb3b6cb` (`c7a185e4ae30de3d2482a65de8b24c5de5ecec24`):

- latest `origin/main`: `a404c08f6eb6abbdbd27935667dd298f6aec522f` (direct ancestor)
- candidate lifecycle: accepted, 95/95 executions in each repetition
- C2: `13,22,32/57` versus protected baseline `3,5,5/38`
- reachability: precision `27/29`, recall `3/5`
- coverage: complete `23`, partial `19`, unavailable `53`
- safety: passed with no findings
- semantic repeat: passed; both report IDs are `sha256:707e1c5ccefa122ebf05c58a6bacabb9e4c7421c8b33f7537d74bb5a303a3dfc`
- suppression projection conformance: all four controls passed in both repetitions (`8/8`), each producing sanitized `present_unreached` evidence with the expected tier and distinct proposer/verifier identities
- publication integrity: all 18 sanitized artifacts matched `artifact-manifest.json`; lifecycle commit/tree matched the checkout
- local baseline: `make lint build vet test typecheck`, CGO-disabled command build, 125 frontend test files / 739 tests, frontend production build, uncached changed-package tests, and workflow validation all passed

Raw evidence, private review material, controller state, and keys remain outside Git. The PR will stay draft until required GitHub checks for this exact SHA are green.

pho-veteran:
Exact-head benchmark and local gate evidence for `c38926df5a080ba029ad2b597c3d477c4db23187` (`cd0debae24d6e89d64e8c2ffdc54380ca872730f`):

- latest `origin/main`: `a404c08f6eb6abbdbd27935667dd298f6aec522f` (direct ancestor)
- candidate lifecycle: accepted, 95/95 executions in each repetition
- C2: `13,22,32/57` versus protected baseline `3,5,5/38`
- reachability: precision `27/29`, recall `3/5`
- coverage: complete `23`, partial `19`, unavailable `53`; no-analysis `54`
- safety: passed with no findings
- semantic repeat: passed; sanitized report ID `sha256:32702af43d183b85a14cc15aa3a3001f23b20140a3e8a4d1b3ff335df328b145`
- suppression projection conformance: all four controls passed in both repetitions (`8/8`), each producing sanitized `present_unreached` evidence with the expected tier and distinct proposer/verifier identities
- publication integrity: all 18 sanitized artifacts matched `artifact-manifest.json`; lifecycle commit/tree matched the checkout; manifest file digest `sha256:409ebf03a5765393f7482c1b7be5ba315c93bb7b9cfd6dac3c3edd8b3b9fe568`
- exact-head local baseline: `make lint build vet test typecheck`, CGO-disabled command build, uncached reachbench policy tests, workflow validation, and Taurus gate all passed
- CI correction: PR lifecycle now selects isolated .NET SDK `8.0.100` and JDK `21.0.5`; Go/Python comparator outputs use explicit non-hidden bounded directories while the aggregate gate remains strict

Raw evidence, private review material, controller state, and keys remain outside Git. The PR remains draft until required GitHub checks for this exact SHA are green.

pho-veteran:
Exact-head benchmark and local gate evidence for `cc7268f98dbfba214382d144c209c3616f727699` (`810cc8f07e0cdb25f1ecdba376d97442e0d2cb77`):

- latest `origin/main`: `a854487a4f02ddfaca0743afad517eaf34f9133c` (direct ancestor)
- current-main integration: preserved bounded-dispatch observation before the display-field early return; focused AST and reachbench suites passed
- candidate lifecycle: accepted, 95/95 executions in each repetition
- C2: `13,22,32/57` versus protected baseline `3,5,5/38`
- reachability: precision `27/29`, recall `3/5`
- coverage: complete `23`, partial `19`, unavailable `53`; no-analysis `54`
- safety: passed with no findings
- semantic repeat: passed; sanitized report ID `sha256:97152ba54212df6cb5839a3371aee92c092db3fc01407f9c408b2d650ff9f3bc`
- suppression projection conformance: all four controls passed in both repetitions (`8/8`), each producing sanitized `present_unreached` evidence with the expected tier and distinct proposer/verifier identities
- publication integrity: all 18 sanitized artifacts matched `artifact-manifest.json`; lifecycle commit/tree matched the checkout; manifest file digest `sha256:625808708d46c58760ab174e74f8d6a88278c3d497bf589b79e17856a156b87d`
- exact-head local baseline: `make lint build vet test typecheck`, CGO-disabled command build, uncached AST/reachbench tests, frontend tests/build, workflow validation, and Taurus gate all passed
- CI correction: PR lifecycle selects isolated .NET SDK `8.0.100` and JDK `21.0.5`, probes `jar` with its supported `--version` flag in both lifecycle routes, and uploads Go/Python comparator outputs from explicit non-hidden bounded directories while the aggregate gate remains strict

Raw evidence, private review material, controller state, and keys remain outside Git. The PR remains draft until required GitHub checks for this exact SHA are green.

nghiadaulau (changes requested):
Thorough work on the lifecycle and the breadth of the new positive-evidence paths, and the raise-only wirings and the digest-bound suppression-projection controls are careful. CI is green, the branch merges clean onto current `main`, and the combined state builds and passes tests. Two blockers on the paramount never-false-suppress invariant have to be fixed before this can merge. Both were confirmed by executing the code, and both are safe→unsafe regressions (on `main` these degrade to blind; this branch seals a false negative).

## Blocker 1 (high): the Python dynamic-import scanner reopens a false-suppression hole

`internal/infrastructure/tools/pyimports/scanner.go` replaces the broad fail-closed `dynamicRE` with narrow call-shaped regexes. The block comment claims "Every other dynamic loader remains a global coverage gap so an unknown target can never produce a negative conclusion", but indirection through a local name or a two-level attribute defeats every new regex. Verified by running the actual regexes:

- `im = importlib.import_module` then `im("requests")` sets neither `DynamicModules` nor `DynamicUnknown`.
- `loader = __import__` then `loader("requests")` sets neither.
- `from importlib import import_module as im` then `im("requests")` sets neither.
- `importlib.util.find_spec('requests')` / `importlib.machinery.*` set neither (`pythonImportlibCallRE` matches only `importlib.NAME(`, not `importlib.util.NAME(`).
- `import runpy as rp` then `rp.run_module('requests')`, and `from runpy import run_module` then `run_module('requests')` set neither.

On `origin/main`, `dynamicRE` matched a bare `\bimportlib\b` / `\brunpy\b` / `\bpkgutil\b` / `__import__`, forcing `DynamicImports=true` (whole-analysis no-coverage). pyreach Tier-1 is a suppressing path (`cmd/synapse-api/main.go`: `NewCoordinatorForTier(..., judgment.Tier1)` then `SetPyReachability(pyCoord)`, no `.WithRaiseOnly()`, logs "→ OpenVEX not_affected"), so with none of these forms flagged, a project that loads a declared vulnerable direct dependency only through one of them scans clean, Tier-1 concludes not-imported, mints `not_reachable`, and drives an OpenVEX `not_affected`, suppressing a real finding. Both an independent security review and an independent soundness review reproduced this.

Fix: restore a fail-closed backstop. Detect the dynamic-import surface first (any `importlib`, `runpy`, `pkgutil`, `imp.load`, `__import__`, `exec`, `eval`, including aliased / `from`-imported / assigned handles). For each dynamic call either recover a literal into `DynamicModules` (positive-only) or set `DynamicUnknown=true`. A name bound to `importlib.import_module` / `__import__` / `importlib` (or aliased on import) must make subsequent calls through that name unknown-dynamic unless the argument is a tracked constant. The .NET path is the model: it emits `unresolvedDotNetReflection` for every unrecognized reflective form. Add regression tests for the evasion classes above.

## Blocker 2 (high): the JS alias resolver seals a false not-reachable on closure reassignment

`internal/domain/jsprogram/resolver.go` (`aliasCallableTargets` / `indexCallableAliases`) counts assignments per lexical scope and bails only on more than one assignment within the declaring scope. It never considers an inner function rebinding a closed-over (free) variable, so a reassignment filed under a nested scope is invisible to the outer resolution. Verified counterexample:

```js
let strategy = defaultStrategy;
function useAggressive() { strategy = aggressiveStrategy; }
useAggressive();
strategy();            // at runtime calls aggressiveStrategy
```

On this branch the outer `strategy()` resolves to `defaultStrategy` only, `Complete==true`, zero gaps, so `aggressiveStrategy` has no incoming edge and no gap and is sealed not-reachable although it is reachable. On `main` the outer call was `unresolved` with a gap (correctly blind). This flows to a suppression path through `jsreach.InterprocRecorder.WithSuppression` (the opt-in `SYNAPSE_JSREACH_INTERPROC_TIER2_ENABLED` interproc-Tier-2 suppression), so it becomes a false `not_affected` there.

Fix: widen to blind (record an `unresolved_call` gap, or drop the alias resolution) whenever an alias name is written from any scope other than, or in addition to, its single declaring-scope assignment. Treat a variable written from a descendant scope as escaping. Add a closure-reassignment regression test.

## Secondary (not blockers, latent)

- Tier-2 bounded-dispatch relaxation (`pythonprogram/resolver.go` `boundedPythonClosure`, `pyreach/tier2.go`) narrows the fail-closed rule from `resolution.Complete` to "complete except a finite literal-map dispatch". No production emitter of `bounded_callees` exists yet, so it is a no-op today, but it becomes a false-suppression vector the moment the `synapse-ast` extractor emits it unless the extractor's immutability / full-enumeration contract is independently verified. Please pin that contract with a comment and a test before anything emits it.
- Closed-world first-party entrypoint narrowing (`pyreach/tier2.go` `closedWorldEntrypoints`) is sound for an application root but would under-approximate entrypoints for a library whose exports are invoked externally. Please confirm every `FirstPartySymbolSubject` `python:*` call site targets a closed-world/application root and document the contract there.

## Verified sound

- Coordinator raise-only enforcement holds at both mint sites (`reachproof/coordinator.go`), and all four raise-only production wirings (JS-symbol, Rust-symbol, source-symbol, Go-binary) call `.WithRaiseOnly()`.
- Python bounded literal-map dispatch keeps candidates out of the graph, roots all bounded candidates, marks the downstream closure `BoundedUncertainNodes`, and sets `Complete=false` whenever that set is non-empty (monotonic, fail-closed).
- JS returned-callable `f()()` and synchronous-callback modeling fail closed (unique factory + single direct return; single positional param invoked exactly once and never stored/returned/passed on).
- srcreach/nuget `ConditionalPackages ⊆ ImportedPackages`, so a conditional package stays reachable with only a blind annotation. .NET reflection recovers known names as conditional positives and fails closed to `unresolvedDotNetReflection` otherwise.
- `NormalizeSourceProvenance` rejects `..`, backslash, NUL, absolute, drive/scheme, leading `/`, and is used only as descriptive metadata on a positive result.
- The new judgment-domain files have no non-test consumers yet (inert scaffolding).

Happy to re-review once the two fail-closed backstops are restored with the evasion-class and closure-reassignment regression tests.

pho-veteran:
Addressed all findings from review `5255863330` in `c98766e3` and integrated current `main` in `b52a4ab8`.

- **Python dynamic imports:** added per-file binding tracking for imported, assigned, and propagated aliases; literal/constant targets remain positive-only evidence, while unknown arguments, nested `importlib.util` / `importlib.machinery`, `runpy`, `pkgutil`, `imp`, `exec`, `eval`, wildcard imports, indirect `getattr`, and poisoned reassignments now fail closed. Added scanner and Tier-1 coordinator regressions proving these forms cannot mint a false `not_reachable`.
- **JavaScript closure reassignment:** descendant writes now poison matching ancestor aliases and emit an unresolved-value gap; direct-call and non-call escape regressions verify the graph cannot remain complete with a stale target. Added a real extractor-to-analyzer regression as well.
- **Bounded dispatch:** documented that `BoundedCallees` requires complete enumeration from immutable producer state, and pinned multi-candidate extraction plus subject-local uncertainty without fabricated static edges.
- **Closed-world Python subjects:** documented the `FirstPartySymbolSubject` application-owned contract and made the production benchmark callsite fail closed unless it is the explicitly identified application fixture/source entry.

Focused suites, the full Go baseline (`make lint build vet test typecheck`), CGO-disabled command builds, all 739 frontend tests, and the production frontend build passed locally. Exact-head CI is running for `b52a4ab89af3d57ba316fbdd84ab7710a1ed232c`; I will request re-review once that head is green.

pho-veteran:
Exact-head authoritative candidate lifecycle now passes for `b52a4ab89af3d57ba316fbdd84ab7710a1ed232c` (tree `f9b230b0054d8e8cf8759da83c1262c9541f316c`).

- candidate route, authoritative acceptance mode
- 2 repetitions, each observing 95/95 required executions
- semantic repeat: pass across all 95 cells
- safety: pass in both repetitions, no findings
- suppression-projection conformance: 4/4 controls pass in each repetition
- C2: production breadth 13; 22 correct positive cases; macro reachable recall 32/57
- identical report ID in both repetitions: `sha256:b3ab2c32ce9450f603476dc2b118263cbd557674a2ad3caf66841ea75ee60a07`
- all 18 sanitized publication artifacts are canonical and match the artifact manifest
- artifact manifest digest: `sha256:bb6391df5a100ce4f7ca4f5974934c95bb00c4f3b94dfc4c1c5c411c98040e0b`

The raw/private controller material and signing keys are not published. GitHub exact-head CI is still completing its remaining backend jobs; re-review will be requested after those finish.

nghiadaulau (approved):
Re-reviewed at `b52a4ab8`. Both blockers are fixed, and I verified each independently rather than taking the summary on trust.

## Blocker 1 (Python false-suppression): fixed

`pyimports/scanner.go` now tracks per-file bindings and classifies every dynamic-loader form: a literal/constant target is recovered into `DynamicModules` (positive, so the dependency reads as reachable and is raised), and every aliased, assigned, propagated, two-level (`importlib.util` / `importlib.machinery`), `runpy`/`pkgutil`/`imp`, `exec`/`eval`, wildcard-import, `getattr`-indirect, or poisoned-reassignment form sets `DynamicImports=true` (whole-analysis no-coverage), which `pyreach` Tier-1 refuses coverage on before minting any verdict. I traced all six evasion classes from the original review through the new code and confirmed none is silently invisible; `TestScanDynamicAliasesAndNestedFormsFailClosed` and `TestPyReachDynamicImportAliasCannotMintNotReachable` pin the behavior.

## Blocker 2 (JS closure-reassignment false-not-reachable): fixed

`jsprogram/resolver.go` now runs a poison pass: a write to a bare name in any scope nils the same-name alias in every ancestor scope and emits a `closure_alias_write` / unresolved-value gap, so the outer resolution goes blind instead of sealing the single declaring-scope target. I traced the exact `let strategy = default; function f(){ strategy = aggressive } f(); strategy()` counterexample through the new code: it now yields `Complete==false` with an unresolved-call gap and does not seal `defaultStrategy`. Coverage holds for directly-nested, deeper-descendant, and callback-passed writes, and destructuring / augmented-assignment targets surface as poison-eligible facts. `TestResolveClosureReassignmentDefeatsCompleteAlias` and the non-call-escape variant pin it.

## Latent notes: addressed

Bounded-dispatch `BoundedCallees` immutability/full-enumeration contract is documented and the closed-world `FirstPartySymbolSubject` benchmark callsite now fails closed unless it is the identified application source. Good.

## Verification

All the reachability, jsprogram, pythonprogram, pyreach, jsreach, and reachproof packages pass on this head, and I merged current `main` (which has advanced since the first review) into it locally: the combined tree builds and the reachability plus my own changed packages pass, so there is no semantic conflict.

## Optional follow-ups (not blockers)

- Python: `getattr(builtins, '__import__')('pkg')` with the loader name as a string literal still evades both the binding tracking and the surface backstop (the `'__import__'` string is blanked and `builtins` is not in the surface set). Very narrow (the direct `builtins.__import__(...)` form is caught, and any other `importlib`/`exec`/`eval` mention already flips the whole-analysis flag), and outside the original six classes, so it does not block. Adding `builtins`/`__builtins__` to the surface set, or flagging `getattr(<dynamic-surface>, ...)`, closes it.
- JS: the poison pass cannot distinguish a descendant's independent local shadow from a closed-over write, so an unrelated same-name ancestor alias is also dropped. That is a recall loss, never a soundness loss, which is the right side to err on.

Approving. Thorough fix.

